PR #65, opened 2025-12-01
feat(sourceables): more freedom over the way the wrapped package is launched

BirdeeHub:
in `nix` makeWrapper implementation

`argv0type = command: ''eval "$(${command})"'';`

Why? starship needs STARSHIP_CONFIG in the SHELL, not its own process

now you can `source ${wrappedstarship}/bin/starship` in a .zshrc

Example useage

https://github.com/BirdeeHub/birdeeSystems/blob/a3c089ee035e5b12bcc65c040fec957a2e4b1ac6/common/wrappers/starship/default.nix

https://github.com/BirdeeHub/birdeeSystems/blob/a3c089ee035e5b12bcc65c040fec957a2e4b1ac6/common/wrappers/wezterm/zdot/default.nix#L46

https://github.com/BirdeeHub/birdeeSystems/blob/a3c089ee035e5b12bcc65c040fec957a2e4b1ac6/common/modules/shell/fish.nix#L17

### DIFF
--- a/modules/makeWrapper/makeWrapperNix.nix
+++ b/modules/makeWrapper/makeWrapperNix.nix
@@ -6,6 +6,7 @@
   ...
 }:
 let
+  inherit (builtins) elemAt;
   generateArgsFromFlags = (import ./genArgsFromFlags.nix { inherit lib wlib; }).genArgs flaggenfunc;
   flaggenfunc =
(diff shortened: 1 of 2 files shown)


BirdeeHub:
These tests all pass locally IDK what is going on

PR #90, opened 2025-12-14

BirdeeHub:


BirdeeHub:
Also something about this cache action makes tests fail the first time when I do nix flake update. Rerunning it then works, but its weird. Locally it doesnt happen, its just in the action when Im using the magic cache action

PR #112, opened 2025-12-26
feat(mako): patch systemd service units

patwid:
Patching the systemd user units as well as the dbus service, so that starting mako either through systemd or dbus will use the wrapped module.

### DIFF
--- a/wrapperModules/m/mako/check.nix
+++ b/wrapperModules/m/mako/check.nix
@@ -4,7 +4,7 @@
 }:
 
 let
-  mpvWrapped =
+  makoWrapped =
     (self.wrapperModules.mako.apply {
       inherit pkgs;
       "--config".content = ''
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Yeah that seems important! :laughing: Thank you!

BirdeeHub:
Oh, also, it's not super important, which is why I went ahead and merged it regardless, but we have a convention for naming commits, which you may find useful in figuring out what to title the commits! https://github.com/BirdeeHub/nix-wrapper-modules/blob/main/CONTRIBUTING.md#commit-messages You did good enough though, for now there is nothing automated attached to those names so it is not important.

BirdeeHub:
Also, looking at that test I had, it is fantastically bad. I should fix that. The whole thing was copied from mpv and I suppose I forgot about it.

Edit: Improved the test further slightly

PR #163, opened 2026-01-10
type(wrapperModules.waybar): init

patwid:
Add waybar wrapper module

### DIFF
--- a/wrapperModules/w/waybar/check.nix
+++ b/wrapperModules/w/waybar/check.nix
@@ -1,0 +1,28 @@
+{
+  pkgs,
+  self,
+}:
+let
+  waybarWrapped = self.wrapperModules.waybar.wrap {
+    inherit pkgs;
+
+    settings = {
+      position = "top";
+      modules-left = [ ];
+      modules-right = [ ];
+      modules-center = [ ];
+    };
+
+    "style.css".content = "";
+  };
+
+in
+if builtins.elem pkgs.stdenv.hostPlatform.system self.wrapperModules.mako.meta.platforms then
+  pkgs.runCommand "waybar-test" { } ''
+    "${waybarWrapped}/bin/waybar" --version | grep -q "${waybarWrapped.version}"
+    touch $out
+  ''
+else
+  pkgs.runCommand "waybar-test-disabled" { } ''
+    touch $out
+  ''
--- a/wrapperModules/w/waybar/module.nix
+++ b/wrapperModules/w/waybar/module.nix
@@ -1,0 +1,74 @@
+{
+  config,
+  wlib,
+  lib,
+  ...
+}:
+let
+  jsonFmt = config.pkgs.formats.json { };
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    settings = lib.mkOption {
+      type = jsonFmt.type;
+      default = { };
+      description = ''
+        Waybar configuration settings.
+        See <https://github.com/Alexays/Waybar/wiki/Configuration>
+      '';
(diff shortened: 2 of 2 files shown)


BirdeeHub:
Hey!

So, the tests are failing because the module requires `imports = [ wlib.modules.default ];` at the top so that it has access to the default options for making wrapper scripts. The core, always imported options are about the derivation. Excited to have it once it is all in order though!

wrapperModules/w/waybar/module.nix
```diff
@@ -0,0 +1,73 @@
...
+  config,
+  wlib,
+  lib,
+  ...
+}:
+let
+  jsonFmt = config.pkgs.formats.json { };
+in
+{
+  _class = "wrapper";
```
BirdeeHub: needs `imports = [ wlib.modules.default ];` here or it won't work, but otherwise, looks pretty good!

Also, you can drop the `_class = "wrapper";`, it would just prevent you from importing modules that are not of also of class `wrapper` or `null`

BirdeeHub: Specifically, these modules can be used as submodules, so you want to be able to import them anywhere.

wrapperModules/w/waybar/module.nix
```diff
@@ -0,0 +1,73 @@
...
+
+  config.package = lib.mkDefault config.pkgs.waybar;
+  config.flags = {
+    "--config" = toString config.configFile.path;
+    "--style" = toString config."style.css".path;
+  };
+  config.filesToPatch = [
+    "share/systemd/user/waybar.service"
+  ];
+  config.meta.maintainers = [
```
BirdeeHub: Also, you should put your entry in [The module maintainers list](https://github.com/BirdeeHub/nix-wrapper-modules/blob/main/maintainers/default.nix) and reference it here like `wlib.maintainers.patwid`

BirdeeHub:
But, yeah other than those 3 things, looks good!

wrapperModules/w/waybar/check.nix
```diff
@@ -0,0 +1,24 @@
...
+        modules-left = [ ];
+        modules-right = [ ];
+        modules-center = [ ];
+      };
+
+      "style.css".content = "";
+    };
+
+in
+pkgs.runCommand "waybar-test" { } ''
```
BirdeeHub: Also, it might run on mac, but if it doesn't we will have to add a check like

https://github.com/BirdeeHub/nix-wrapper-modules/blob/1682470f1044bb530627b8385b4bcf8169f45fbd/wrapperModules/m/mako/check.nix#L14

BirdeeHub: I should let the check.nix files in the repo return null to disable themselves, when I do that I will do the update treewide on all the things so you won't need to worry about that though

patwid:
Is there something you want me to do about the failing test on macos? As far as I understand it, the `waybar` package is only available on linux, which I already specified in `meta.platforms`.

BirdeeHub:
Awesome, thanks! LGTM now!

BirdeeHub:
Oh I missed something. In the tests

https://github.com/BirdeeHub/nix-wrapper-modules/pull/167

Its wrappedModules

wrapperModules is going to be for the unevaluated ones eventually.

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/blob/f5ed0f75af12e3911aac70c6402e8b4934c16b03/wrapperModules/w/waybar/check.nix

The updated check.nix file

PR #165, opened 2026-01-11
feat(lib.types.seriesOf): for `.override` and `.overrideAttrs` style behavior

BirdeeHub:
Fully type-merge aware types

This is a big update. You can override the fields of DAG and DAL types, or construct your own ad-hoc ones! `wlib.dag.mkDagEntry` and `wlib.dag.dagNameModule` have been added for this purpose.

`wlib.types.seriesOf` was added, to fix type merging for `config.overrides`, and `config.overmods`, a prior workaround for the issue, has been deprecated with a descriptive warning about how to make it yourself.

`wlib.types.subWrapperModuleWith` is simply a submodule alias with more arguments

### DIFF
--- a/lib/core.nix
+++ b/lib/core.nix
@@ -219,93 +219,66 @@ in
         values from `config.overrides` applied to it.
       '';
     };
-    overmods = lib.mkOption {
-      type = lib.types.deferredModule;
-      default = { };
-      description = ''
-        Provide a module to modify the `wlib.types.spec` type in `config.overrides`.
-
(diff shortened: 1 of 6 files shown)


BirdeeHub:
Some nested type docs are still not showing up on the website as I would expect, so, I will hold off on this while I look into it and work on it. To be fair, none of them worked before, and I had to mark the data name before and after fields as internal for noise reasons because it made the docs just too much.

The options that aren't showing up work fine, I just need to make sure its just docgen and why it is happening.

PR #167, opened 2026-01-11
fix(wrapperModules.waybar): remove deprecated value from tests

BirdeeHub:


BirdeeHub:
I really need to figure out what is making these mpv and jujutsu tests flaky when using detnix....

Means I need to install detnix and find out though....

PR #171, opened 2026-01-12
ci(cachix): try cachix action instead

BirdeeHub:


BirdeeHub:
Aww it doesn't work on mac....

PR #174, opened 2026-01-13
feat(wrapperModules.openssh): init

patwid:
Add openssh wrapper module (allowing for project specific ssh client configuration).

Currently, completely unstructured. Later it we could switch to a structured configuration (maybe analogous to home-manager). Also I don't know how to test the module, yet.

### DIFF
--- a/wrapperModules/o/openssh/check.nix
+++ b/wrapperModules/o/openssh/check.nix
@@ -1,0 +1,27 @@
+{
+  pkgs,
+  self,
+}:
+let
+  opensshWrapped = self.wrappedModules.openssh.wrap {
+    inherit pkgs;
+
+    settings = ''
+      Host foo
+        User bar
+        HostName 192.168.0.2
+        ProxyJump baz
+
+      Host baz
+        HostName 192.168.0.1
+    '';
+  };
+
+in
+if builtins.elem pkgs.stdenv.hostPlatform.system self.wrappedModules.openssh.meta.platforms then
+  pkgs.runCommand "ssh-test" { } ''
+    "${opensshWrapped}/bin/ssh" -G foo | grep -q 'HostName 192.168.0.2'
+    touch $out
+  ''
+else
+  null
--- a/wrapperModules/o/openssh/module.nix
+++ b/wrapperModules/o/openssh/module.nix
@@ -1,0 +1,40 @@
+{
+  config,
+  wlib,
+  lib,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    settings = lib.mkOption {
+      type = lib.types.string;
+      default = "";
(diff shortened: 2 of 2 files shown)


BirdeeHub:
As far as testing it goes, you should grep the output of `ssh -G host` or `ssh -G client` as that outputs the existing settings for the host. Or you could set the settings for a particular url and `ssh -G <thaturl>`.

That way you can set the options, and then grep the output for the things you added to make sure they are picked up by the wrapped ssh command created.

BirdeeHub:
Also, it appears the home manager module actually only installs the program and sets the config file content

https://github.com/nix-community/home-manager/blob/82fb7dedaad83e5e279127a38ef410bcfac6d77c/modules/programs/ssh.nix#L616-L639

If you want, you can basically just copy paste this thing, import the wlib.modules.default module, and swap the thing it sets from `config.home.packages` and `config.home.files` to `config.package` and `config.flags."-F"`, then get rid of the enable option (or just make it default to true, that is also fine, might be useful when used as a submodule, just make sure it sets `config.package` regardless), and change the names of the options slightly so they aren't nested under `programs.ssh`. Oh, and you would also want to get rid of [this](https://github.com/nix-community/home-manager/blob/82fb7dedaad83e5e279127a38ef410bcfac6d77c/modules/programs/ssh.nix#L461-L488), it just handles their deprecations, which we never had to begin with.

Because right now, people can pretty easily already do what this module does

```nix
  wlib.wrapProgram [
    { inherit pkgs; }
    ({ pkgs, ... }: {
      package = pkgs.openssh;
      flags."-F" = pkgs.writeText "ssh-config" ''
        <config-content>
      '';
    })
  ]
```

It might be nice to just go straight for converting the home manager module to work here, rather than dealing with deprecating stuff later.

Then if people just want to override the file they can do it in a few lines like I just showed (or `config.flags."-F" = lib.mkForce somethingelse`), but if they want structured options they can use the module.

BirdeeHub:
Also I figured out that if I squash merge, it should still show you on the contributors pane on github despite you using a different email for the commit compared to your account's noreply email.

So I will do that for your future commits, sorry I didn't realize that before.

I have a git config like this that fixes that for me for my own commits.

```nix
{config, pkgs, wlib, lib, ... }: {
  imports = [ wlib.wrapperModules.git ];
  settings = {
    user.name = "Birdee";
    user.email = "<85372418+BirdeeHub@users.noreply.github.com>";
  };
}
```

Please replace that with your own github noreply email and username if you copy that.

---

As for the current PR, the suggestions I gave are still relevant.

I would highly recommend running the tests locally by running `nix flake check -Lv` from the root of the repository on your own machine, as it will give you faster feedback than submitting the PR and waiting for me to get around to checking github.

PR #175, opened 2026-01-14
refactor(modules.makeWrapper): contain wrapperFunction logic to wrapperFunction

BirdeeHub:


BirdeeHub:
Also added a print to the jujutsu test to see if I can figure out what is going on with those (of course now I can't get it to fail...)

PR #177, opened 2026-01-14
 feat(wrapperModules.ov): init

rencire:
Added wrapper module for [https://github.com/noborus/ov](https://github.com/noborus/ov).

### DIFF
--- a/wrapperModules/o/ov/module.nix
+++ b/wrapperModules/o/ov/module.nix
@@ -21,7 +21,7 @@ in
     };
   };
   config.flags = {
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Awesome!

Its simple, but it supports structured settings, and the program looks like it does have a significant amount of config.

I'd obviously prefer it have a test, however, I'm not sure what you really even test here that isn't just testing that the `makeWrapper` module works. So, I am fine with it not having one.

So, looks good to me! Thank you!

PR #181, opened 2026-01-18
feat(wrapperModules.claude-code): init

vinnymeller:
Adds a claude code wrapper module with the ability to add your MCPs & main claude settings via the wrapper

### DIFF
--- a/wrapperModules/c/claude-code/module.nix
+++ b/wrapperModules/c/claude-code/module.nix
@@ -1,0 +1,90 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+let
+  jsonFmt = pkgs.formats.json { };
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+
+    settings = lib.mkOption {
+      type = jsonFmt.type;
+      default = { };
+      description = ''
+        Claude Code settings
+
+        These settings will override local, project, and user scoped settings.
+
+        See <https://code.claude.com/docs/en/settings>
+      '';
+      example = {
+        includeCoAuthoredBy = false;
+        permissions = {
+          deny = [
+            "Bash(sudo:*)"
+            "Bash(rm -rf:*)"
+          ];
+        };
+      };
+    };
+
+    mcpConfig = lib.mkOption {
+      type = lib.jsonFmt.type;
+      default = { };
+      description = ''
+        MCP Server configuration
+
+        Exclude the top-level `mcpServers` key from the configuration as it is automatically handled.
+
+        See <https://code.claude.com/docs/en/mcp>
+      '';
+      example = {
+        nixos = {
+          command = "${pkgs.mcp-nixos}/bin/mcp-nixos";
+          type = "stdio";
+        };
+      };
+    };
+
+    strictMcpConfig = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Whether to enable the `--strict-mcp-config` flag for Claude Code.
+
+        When enabled, Claude will only use the MCP servers provided by the `mcpConfig` option.
+
+        If disabled, Claude may use MCP servers defined elsewhere (e.g., user or project scoped configurations).
+      '';
+    };
+
+  };
+
+  config = {
+    package = lib.mkDefault pkgs.claude-code;
+    unsetVar = lib.mkDefault [
+      "DEV"
+      # the vast majority of users will want to authenticate with their claude account and not an API key
+      "ANTHROPIC_API_KEY"
+    ];
+    envDefault = lib.mkDefault {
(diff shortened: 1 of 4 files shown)


wrapperModules/c/claude-code/module.nix
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+  };
+
+  config = {
+    package = lib.mkDefault pkgs.claude-code;
+    unsetVar = lib.mkDefault [
+      "DEV"
+      # the vast majority of users will want to authenticate with their claude account and not an API key
+      "ANTHROPIC_API_KEY"
```
BirdeeHub: Maybe we could either add an option `allow_impure_auth = true` to avoid unsetting this?

Or you could most likely also `env.ANTHROPIC_API_KEY = "";` and then users can still override it in a pinch, but test that that works first if you go with that, I don't have a subscription.

Otherwise, its looking pretty good to me

BirdeeHub:
Oh.

Um, so that its part of this commit, can you add `config.allowUnfree = true;`

in the attrset on this line here

https://github.com/BirdeeHub/nix-wrapper-modules/blob/346f57363014a25585946b4ac1d126c1469f710c/flake.nix#L14

(sorry I know that part of the flake is a bit of a mess, its for non-flake compat)

The test environment doesn't want to build claude code. That should fix the issue.

Edit: Thanks! Lets see! The flaky tests might strike again too, if jujutsu or mpv fails thats not you.

vinnymeller:
hmm running it locally it doesn't seem to care if i changed the line you said, but the section above makes it happy:

```nix
        else if inputs.nixpkgs.legacyPackages.${system} or null != null then
          import inputs.nixpkgs {
            inherit system;
            config.allowUnfree = true;
          } 
```

BirdeeHub:
Oh. That makes sense actually. I guess there isnt a way to do that without reinstantiating pkgs.

Yeah do that one too please.

Thanks, sorry about that.

People provide their own pkgs for their modules when they build them, so this only affects the tests. And it doesn't affect how it fetches the lib either. It actually doesn't affect the docs either.

vinnymeller:
yep lol i tested my branch of the module in my config first and it worked, pushed branch, then realized i should run nix flake check in the repo at ~the same time you first commented xD

BirdeeHub:
that would be mpv that time... It seems to sometimes make it fail the first run but I cannot figure out why. jujutsu also. And only in actions.

BirdeeHub:
I can go over that section at the top of the flake on my own after this commit, but since this works, that part is fine for now.

BirdeeHub:
Thoughts on the option about unsetting the api variable? I use opencode, so I am not sure what the effect of that would be for claude code.

vinnymeller:
basically, i think it technically will work if you just have `ANTHROPIC_API_KEY` in your env, but they heavily encourage you to actually sign in with the `/login` command instead, and it will be fussy if you both have an api key in your env and are logged in at the same time

they removed mentions of just having the `ANTHROPIC_API_KEY` env var both in their quickstart: https://code.claude.com/docs/en/quickstart

and in the identity and access management docs: https://code.claude.com/docs/en/iam

instead, they push you to turn on "extra usage" in your account settings, which will automagically switch to just charging you per token once your limit on whatever plan you're on runs out, along with warning you that you'll now be paying-as-you-go until your limit refreshes: https://support.claude.com/en/articles/12429409-extra-usage-for-paid-claude-plans

there are probably some edge cases where its reasonable to use the api key directly, but for essentially everbody its not the "right" thing to do

also happy to just remove that line if you dont like it and unset it in my own config ;p

BirdeeHub:
nono, its good. It says best practices.

And actually, they can still set it via the wrapper, by default unset is first, followed by env.whatever, although you can target stuff with the DAG too. And you can always mkForce over the whole list.

But also, here. I have something direct that helps a bit and doesnt need a new option.

replace

`"ANTHROPIC_API_KEY"`

with

```nix
{
  name = "ANTHROPIC_API_KEY";
  data = "ANTHROPIC_API_KEY";
}
```

directly within the list.

Then if people want to run something before via the dag to save the value they can, or they can just mkForce over the whole unsetVars list and fix it themselves, its only 2 items.

Otherwise, it looks good to me.

I probably should have made that a set type option, but you could technically unset something before and after a runShell, so its a list.

BirdeeHub:
Honestly, now that I am remembering `lib.mkForce` exists, this is fine. If people complain one of us can add the option later.

BirdeeHub:
Thank you for your cooperation on this with the flake thing!

BirdeeHub:
I missed a typo, that was my bad.

Also I simplified the main flake.nix

Those mpv and jujutsu tests are driving me crazy, that is definitely next on my list, although what is causing it I have no idea.

Im going to disable those 2 tests for now. Im done dealing with them on PRs until I can figure out what is wrong with them in actions.

PR #189, opened 2026-01-19
fix(wrapperModules.claude-code): broken mcpConfig

vinnymeller:
1. The `--mcp-config` flag was incorrectly using `config.settings` instead of `config.mcpConfig`
2. It turns out the previous test wasn't actually doing was it appeared to. The MCP config was invalid (since it was using settings), and it just ? does nothing ? in that case instead of erroring out. Additionally, it turns out the `claude mcp get <server>` command doesn't respect the CLI flags at all, so the test was never actually going to work.

(i was way too tired last night apologies again)

### DIFF
--- a/wrapperModules/c/claude-code/check.nix
+++ b/wrapperModules/c/claude-code/check.nix
@@ -16,6 +16,11 @@ let
   };
 in
 pkgs.runCommand "claude-code-test" { } ''
-  "${claudeCodeWrapped}/bin/claude" mcp get nixos
+  claude="${claudeCodeWrapped}/bin/claude"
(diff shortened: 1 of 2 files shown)


BirdeeHub:
> Additionally, it turns out the claude mcp get <server> command doesn't respect the CLI flags at all

Oh fun. Well, it was made with claude code...

> (i was way too tired last night apologies again)

Apology accepted I'm glad you are back to maintain it!

Thank you for the fixes!

vinnymeller:
LOL ive been having lots of problems with claude lately... seemingly ever since they said ~100% of updates are done via claude himself. might have to try opencode :eyes:

PR #201, opened 2026-01-19
feat(niri): update structure

patwid:
Support properties and not just arguments/values to nodes. 

For example the following `output` already relied on arguments/values (this was implemented using `_attrs`):

```kdl
output "DP-1" {
  // ...
}
```

In order to define a `position` for this `output`, we had to use `_keys`:

```nix
output = {
  DP-1 = {
    _keys = true;
    x = 0;
    y = 0;
  };
};
```

To produce the following output:

```kdl
output "DP-1" {
  position x=0 y=0
}
```

Now this would be specified using `_attrs` instead:

```nix
output = {
  DP-1 = {
    _attrs = {
      x = 0;
      y = 0;
    };
  };
};
```

In fact, `_attrs` and `_keys` are kind of related to each other, that's why I merged them. Arguments/values are now specified by passing `null` as a value to the property. This is more consistent with the generic kdl mapping anyway. I think we can even further simplify the implementation, by getting rid of `mkTagged` later. Also I am not really happy, with the name of `mkBlock`, because it now is used for missing values. Anyway, I think we can address this later.

### DIFF
--- a/wrapperModules/n/niri/module.nix
+++ b/wrapperModules/n/niri/module.nix
@@ -60,7 +60,7 @@ let
     ) a;
   mkTagged =
     t: k: v:
-    "${t} ${k}=${toVal v}";
+    "${t} ${mkAttrs { ${k} = v; }}";
   mkRule =
     block: r:
     let


BirdeeHub:
What does it look like to use this new feature? How would I add the modifier? We could add it to the example field for the option maybe so people could know about it and how to use it?

Edit: I get it now I think actually? So at least I somewhat know the goal now.

(also I fixed the flaky tests by the way)

BirdeeHub:
awesome. TY for adding the examples. Its failing the formatting check at the moment, but the other checks seem to have passed. Running `nix fmt` in the repository would fix that!

BirdeeHub:
Draft looks good to me now, let me know when its ready!

patwid:
> What does it look like to use this new feature? How would I add the modifier? We could add it to the example field for the option maybe so people could know about it and how to use it?
> 
> Edit: I get it now I think actually? So at least I somewhat know the goal now.
> 
> (also I fixed the flaky tests by the way)



> What does it look like to use this new feature? How would I add the modifier? We could add it to the example field for the option maybe so people could know about it and how to use it?
> 
> Edit: I get it now I think actually? So at least I somewhat know the goal now.

I just updated the description to explain the change in more detail. Let me know what you think!

> (also I fixed the flaky tests by the way)

Nice!

BirdeeHub:
_keys = true changing the behavior of the node was a bit weird, having them grouped into their own attribute is OK with me.

The main issue, is that it is breaking. I would prefer if it had a deprecation message, but I understand why it does not, as its kinda hard to insert one in the middle of that parsing.

PR #204, opened 2026-01-20
feat(wrapperModules.btop): init

ameertaweel:
Create a wrapper module for the resource monitoring program [Btop](https://github.com/aristocratos/btop) and add myself as maintainer.

### DIFF
--- a/wrapperModules/b/btop/module.nix
+++ b/wrapperModules/b/btop/module.nix
@@ -1,0 +1,61 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib.types)
+    attrsOf
+    oneOf
+    bool
+    float
+    int
+    str
+    ;
+
+  toBtopConf = lib.generators.toKeyValue {
+    mkKeyValue = lib.generators.mkKeyValueDefault {
+      mkValueString =
+        v:
+        if builtins.isBool v then
+          (if v then "True" else "False")
+        else if builtins.isString v then
+          ''"${v}"''
+        else
+          builtins.toString v;
+    } " = ";
+  };
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Nice!

If you want to come by soon and add [the themes thing home manager has for it](https://github.com/nix-community/home-manager/blob/63a87808f5f9b6e4195a1d33f6ea25d23f4aa0df/modules/programs/btop.nix#L131-L141) as well via the `--themes-dir` option that would be awesome but this is totally fine for now.

If you wanted to do that, you could build the directory via `config.drv.buildPhase = ...;` just like in a derivation.

If you do that, include the `runHook preBuild` and `runHook postBuild` thing before and after it, and build them to `$out/${lib.escapeShellArg config.binName}-themes/*.theme` and then you can `config.flags."--themes-dir" = "${placeholder "out"}/${config.binName}-themes";`

Edit: whoops, config.flags is already escaped, but you could set the following to ensure it always uses the one you want

```nix
config.drv.buildPhase = ''
  runHook preBuild
  # build to $out/${lib.escapeShellArg config.binName}-themes/${lib.escapeShellArg *}.theme
  runHook postBuild
'';
config.flags."--themes-dir" = { # <- these take specs! you can also set `sep`, `before` and `after`!
  data = "${placeholder "out"}/${config.binName}-themes";
  esc-fn = lib.escapeShellArg 
};
```

BirdeeHub:
Nice! Built it and it works.

If you want to add the theme thing at some point, please do, how to get started with that is in the above comment, otherwise, LGTM for now!

ameertaweel:
Thanks for your quick review and advice on how to improve this further @BirdeeHub.

I will add support for themes and submit a new PR.

BirdeeHub:
I slightly messed up my suggestion in the comment above, edited it, it was just an extra escapeShellArg I didn't need, but I also had more to add to it

PR #227, opened 2026-01-24
feat(integrations): integration with other module systems

BirdeeHub:
changed `outputs.wrappedModules` to `outputs.wrappers` (Sorry! But also this is much better and the repo is only 2 months old still. You have 8 months to change it before the old one is removed. This name will not change again)

Added flake-parts module for setting up your outputs, (`wrapperModules`, `wrappers`, and `packages.*.*`)

Added flake-parts template.

Added function for creating nixos and home manager modules from any wrapper module.

Added flakeModules nixosModules and homeModules outputs to the main flake.

Improved documentation.

Did the treewide update for checks from outputs.wrappedModules to outputs.wrappers

Added specific dates of removal to some deprecations.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,7 +133,7 @@ Example:
   self,
 }:
 let
-  gitWrapped = self.wrappedModules.git.wrap {
+  gitWrapped = self.wrappers.git.wrap {
     inherit pkgs;
     settings = {
       user = {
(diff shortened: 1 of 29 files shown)


MacaylaMarvelous81:
Do you write news for the project to inform about changes like these to people who don't watch the repo?

BirdeeHub:
No but I probably should.

I don't anticipate any other crazy ones like "I changed the name of the flake output"

That is reserved for, well, right now basically because the repo is still new.

But its starting to get to the point where I can't just go and do stuff like that anymore, hence why I just pushed it out fast, because I wanted to get it in before it gets set in stone in everyone's configs.

So, I won't be doing any more big changes to what is there like majorly changing core value names anymore, this is the last one. I might mess with wlib.checks but I think this repo is the only thing which will actually consume that value anyway, no one else is going to want to run our tests for us lol

Im taking a break for a week or so in order to do some IRL stuff (I will still be here to review, its fun, just not working on docgen stuff until next week) and then coming back and tackling docgen properly next. But that shouldn't change anyone's usage, Ive already made sure I can access all the necessary information for that.

I can pin a thread in discussions when I start working on service file generation options, but I anticipate that not having a major effect on usage luckily, other than maybe a new stdenv phase existing by default, or some new options.

Outside of that, that is all the changes to the core I have planned, after that, its all just adding helper modules for various things (theme options, docker, bubblewrap, etc) and wrapper modules and the core should be pretty solid and shouldn't need to change.

PR #229, opened 2026-01-24
feat(wrapperModules.vim): init

ameertaweel:
Create a wrapper module for Vim and add myself as maintainer.

### DIFF
--- a/wrapperModules/v/vim/check.nix
+++ b/wrapperModules/v/vim/check.nix
@@ -1,0 +1,67 @@
+{
+  pkgs,
+  self,
+}:
+
+let
+  vimWrappedDefault = self.wrappers.vim.wrap { inherit pkgs; };
+
+  vimWrapped = vimWrappedDefault.wrap (
+    { pkgs, ... }:
+    {
+      plugins = [
+        pkgs.vimPlugins.vim-sleuth
+      ];
+
+      optionalPlugins = [
+        pkgs.vimPlugins.vim-surround
+      ];
+
+      vimrc = ''
+        set tabstop=17
+      '';
+    }
+  );
+
+in
+pkgs.runCommand "vim-test" { } ''
+  res=$('${vimWrappedDefault}/bin/vim' -es +':redir >> /dev/stdout' +'echo g:loaded_sensible' +'q')
+
+  if ! echo "''${res}" | grep -q 'yes'; then
+    echo 'Vim does not load default plugins.'
+    touch $out
+    exit 1
+  fi
+
+  res=$('${vimWrapped}/bin/vim' -es +':redir >> /dev/stdout' +'scriptnames' +'q')
+
+  if ! echo "''${res}" | grep -q 'sleuth.vim'; then
+    echo 'Vim does not load plugins.'
+    touch $out
+    exit 1
+  fi
+
+  if ! echo "''${res}" | grep -qv 'surround.vim'; then
+    echo 'Vim loads optional plugins on startup.'
+    touch $out
+    exit 1
+  fi
+
+  res=$('${vimWrapped}/bin/vim' -es +':redir >> /dev/stdout' +'packadd vim-surround' +'scriptnames' +'q')
+
+  if ! echo "''${res}" | grep -q 'surround.vim'; then
+    echo 'Vim does not load optional plugins.'
+    touch $out
+    exit 1
+  fi
+
+  res=$('${vimWrapped}/bin/vim' -es +':redir >> /dev/stdout' +'set tabstop?' +'scriptnames' +'q')
+
+  if ! echo "''${res}" | grep -q 'tabstop=17'; then
+    echo 'Vim does not apply configuration.'
+    touch $out
+    exit 1
+  fi
+
+  touch $out
+''
--- a/wrapperModules/v/vim/module.nix
+++ b/wrapperModules/v/vim/module.nix
@@ -1,0 +1,93 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) types;
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    plugins = lib.mkOption {
+      type = types.listOf types.package;
+      default = [ pkgs.vimPlugins.vim-sensible ];
+      example = lib.literalExpression "[ pkgs.vimPlugins.YankRing-vim ]";
+      description = ''
+        List of vim plugins to install.
+
+        Loaded on launch.
+
+        To get a list of supported plugins run:
+        {command}`nix-env -f '<nixpkgs>' -qaP -A vimPlugins`.
+      '';
+    };
+
+    optionalPlugins = lib.mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = lib.literalExpression "[ pkgs.vimPlugins.elm-vim ]";
+      description = ''
+        List of vim plugins to install.
+
+        Manually loadable by calling `:packadd $plugin-name`.
+
+        If a plugin has a dependency that is not explicitly listed in
+        {option}`optionalPlugins`, that dependency will always be added to
+        {option}`plugins` to avoid confusion.
+
+        To get a list of supported plugins run:
+        {command}`nix-env -f '<nixpkgs>' -qaP -A vimPlugins`.
+      '';
+    };
+
+    vimrc = lib.mkOption {
+      type = types.lines;
+      default = "";
+      example = ''
+        set nocompatible
+        set nobackup
+      '';
+      description = ".vimrc config";
+    };
+
+    overrides = lib.mkOption {
+      type = wlib.types.seriesOf (
+        wlib.types.spec {
+          config.after = lib.mkDefault [ "makeCustomizable" ];
+        }
+      );
+    };
+  };
+
+  config.overrides = [
+    {
+      name = "makeCustomizable";
+      data =
+        package:
+        wlib.makeCustomizable "customize" { }
+          (
+            if package ? customize then
+              package.customize
+            else
+              (pkgs.vimUtils.makeCustomizable package).customize
+          )
+          {
+            vimrcConfig = {
+              customRC = config.vimrc;
+              packages.nix-wrapper-modules.start = config.plugins;
+              packages.nix-wrapper-modules.opt = config.optionalPlugins;
+            };
+          };
+    }
+  ];
+
+  config.package = lib.mkDefault pkgs.vim-full;
+  config.binName = "vim";
+  config.exePath = "bin/vim";
+
+  meta.maintainers = [ wlib.maintainers.ameer ];
+}


MacaylaMarvelous81:
I tested this module like this:
```sh
$ nix-build --expr '(import ./. {}).wrappers.vim.wrap { pkgs = import <nixpkgs> {}; }'
$ nix-shell -p ./result
[nix-shell]$ vim -c ':redir >> /dev/stdout | echo g:loaded_sensible | qall!'
yes
```
I think a check could be written which tests the plugins and vimrc options automatically using vim commands like this.

wrapperModules/v/vim/module.nix
```diff
@@ -0,0 +1,59 @@
...
+      example = "pkgs.vim";
+    };
+  };
+
+  config.package =
+    let
+      vimCustomized = config.packageConfigurable.customize {
+        vimrcConfig = {
+          customRC = config.vimrc;
+          packages.nix-wrapper-modules.start = config.plugins;
```
BirdeeHub: Should we make an option for the opt directory too? Like `config.optionalPlugins` ?

---

Or (if you want) you could use a wlib.types.spec type to have the ability to include `{ opt = true; plugin = idk; }`

```nix
lib.types.listOf (wlib.types.spec {
  options = {
    plugin = lib.mkOption {
      type = lib.types.package;
      description = "...";
      # no default makes this the main field which bare items in the list get added to.
     };
    lazy = lib.mkOption { # or opt is also a fine name
      type = lib.types.bool;
      default = false;
      description = "...";
    };
  };
})
```

And then config.plugins can accept

```nix
config.plugins = [
  pkgs.vimPlugins.vim-sleuth
  {
    plugin = pkgs.vimPlugins.idk;
    opt = true;
  }
];
```

and it will auto normalize the list for you in `config.plugins`, and you can just `lib.partition` them based on their value of `lazy` or `opt`

And people could also then type-merge in extra fields to them too if they wanted.

---

Otherwise, outside of having a way to add optional plugins, looks reasonable, it just uses the nixpkgs vim wrapper.

Im not personally a fan of that wrapper for neovim, which is why I wrote my own, but the reasons why I don't like it for neovim is because of the neovim specific wrapping they put on top, not the vim pack function that makes the packpath. Plus, it was originally written for vim so I would assume it is good enough for vim.

wrapperModules/v/vim/module.nix
```diff
@@ -0,0 +1,59 @@
...
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) types;
+
+  defaultPlugins = [ pkgs.vimPlugins.vim-sensible ];
```
BirdeeHub: Any reason why this is in an extra variable? Its fine I suppose but you only use it once and Im not sure where else it would be used.

BirdeeHub:
And as Macayla says, you are definitely able to run this package in a test and verify that stuff has been added to the config, so I would appreciate it having a check.nix file so that it doesn't break without us knowing about it.

But once it gets a test, and maybe an option for putting stuff in the opt directory (because its added the same way as you are adding the existing plugins option just .opt instead of .start), then it looks good to me.

MacaylaMarvelous81:
@BirdeeHub, have you made review comments? No-one except for you will be able to see them unless you press the button to finish your review, since it's like a draft. I noticed that in my atool PR your review comments suddenly appeared after you approved the PR, and I hadn't known you had left comments until you had done so. I think the same thing may be happening with this PR.

BirdeeHub:
TYSM for correcting my github skill issue I did not know it worked like that.

I did have a review comment or 2, hopefully fairly simple to implement. Those being a test, and a feature to be added.

wrapperModules/v/vim/module.nix
ameertaweel: No reason. I have inlined it.

wrapperModules/v/vim/module.nix
ameertaweel: I went with the simpler approach of `config.optionalPlugins`.

ameertaweel:
I believe I addressed everything except for writing tests. I'm hoping to get it done today as well...

wrapperModules/v/vim/module.nix
```diff
@@ -0,0 +1,80 @@
...
+  config.package =
+    let
+      vimCustomized = config.packageConfigurable.customize {
+        vimrcConfig = {
+          customRC = config.vimrc;
+          packages.nix-wrapper-modules.start = config.plugins;
+          packages.nix-wrapper-modules.opt = config.optionalPlugins;
+        };
+      };
+      vimCustomizedWithMeta = vimCustomized.overrideAttrs {
```
BirdeeHub:
```nix
  config.package =
    let
      vimCustomized = config.packageConfigurable.customize {
        vimrcConfig = {
          customRC = config.vimrc;
          packages.nix-wrapper-modules.start = config.plugins;
          packages.nix-wrapper-modules.opt = config.optionalPlugins;
        };
      };
    in
    vimCustomized;
  config.binName = "vim";
  config.exePath = "bin/vim";
```

You don't have to actually give it back its `meta.mainProgram` by the way. you just have to specify `exePath` and `binName` so it doesnt try to check the package for it.

I didn't realize why you were doing this earlier, sorry for not pointing it out to you then.

wrapperModules/v/vim/check.nix
```diff
@@ -0,0 +1,58 @@
...
+}:
+
+let
+  vimWrappedDefault =
+    (self.wrappers.vim.apply {
+      inherit pkgs;
+    }).wrapper;
+
+  vimWrapped =
+    (self.wrappers.vim.apply {
```
BirdeeHub: you can .wrap here instead if you want, would look cleaner but be the same. Otherwise, aside from the last one failing these tests are fine.

wrapperModules/v/vim/module.nix
BirdeeHub: Would this `wlib.makeCustomizable` be a useful function to add? 

It would turn `.customize` into a `.override`-like function.

```nix
  # type merging lets us add after = [ "makeCustomizable" ]; to all of them by default.
  options.overrides = lib.mkOption {
    type = wlib.types.seriesOf (
      wlib.types.spec {
        config.after = lib.mkDefault [ "makeCustomizable" ];
      }
    );
  };
  config.overrides = [
    {
      name = "makeCustomizable";
      data =
        package:
        wlib.makeCustomizable "customize" { }
          (
            if package ? customize then
              package.customize
            else
              (pkgs.vimUtils.makeCustomizable package).customize
          )
          {
            vimrcConfig = {
              customRC = config.vimrc;
              packages.nix-wrapper-modules.start = config.plugins;
              packages.nix-wrapper-modules.opt = config.optionalPlugins;
            };
          };
    }
  ];
  config.package = pkgs.vim;
  config.binName = "vim";
  config.exePath = "bin/vim";
```

Then people could also

```nix
config.overrides = [ { type = "customize"; data = { /* extra customize args */ }; } ];
```

or

```nix
config.overrides = [ { type = "customize"; data = prev: { /* extra customize args */ }; } ];
```

And they could do that as many times as they wanted

I have a working version of such a function actually but Im trying to see if that makes any sense.

```nix
  makeCustomizable =
    name:
    {
      patches ? [
        "override"
        "overrideAttrs"
        "overrideDerivation"
      ],
      mergeArgs ? origArgs: newArgs: origArgs // (if lib.isFunction newArgs then newArgs origArgs else newArgs),
    }@opts:
    f:
```

Edit:

```nix
nix-repl> res = outputs.lib.makeCustomizable "withPackages" { mergeArgs = og: new: lp: (og lp) ++ (new lp); } pkgs.luajit.withPackages (lp: with lp; [ inspect ])

nix-repl> res.withPackages (lp: [ lp.cjson ])
```

Ok so I might have to add this function

BirdeeHub: https://github.com/BirdeeHub/nix-wrapper-modules/pull/234 <- said function

BirdeeHub: This unfortunately wont fix the issue with the lazy plugin test though

wrapperModules/v/vim/check.nix
BirdeeHub:
```nix
  vimWrappedDefault = self.wrappers.vim.wrap { inherit pkgs; };

  vimWrapped = vimWrappedDefault.wrap (
    { pkgs, ... }:
    {
      plugins = [
        pkgs.vimPlugins.vim-sleuth
      ];
      optionalPlugins = [
        pkgs.vimPlugins.vim-surround
      ];
    }
  );
```

wrapperModules/v/vim/module.nix
BirdeeHub: I found out why the test is failing.

You have to packadd `vim-surround` and not `surround`

I would like [the above change](https://github.com/BirdeeHub/nix-wrapper-modules/pull/229#discussion_r2725840940) added though, since I added a function which is perfect for it. It lets customize be called multiple times, lets us just use the existing package option, and allows advanced users access to all the args to add their own options with it.

But it would need you to rebase this on top of main to get the function. If you don't know how, it could be done in a second PR I suppose but I would prefer it in this one if you can.

wrapperModules/v/vim/check.nix
```diff
@@ -0,0 +1,58 @@
...
+    exit 1
+  fi
+
+  if ! echo "''${res}" | grep -qv 'surround.vim'; then
+    echo 'Vim loads optional plugins on startup.'
+    touch $out
+    exit 1
+  fi
+
+  res=$('${vimWrapped}/bin/vim' -es +':redir >> /dev/stdout' +'packadd surround' +'scriptnames' +'q')
```
BirdeeHub:
```suggestion
  res=$('${vimWrapped}/bin/vim' -es +':redir >> /dev/stdout' +'packadd vim-surround' +'scriptnames' +'q')
```

Fix for failing test

wrapperModules/v/vim/check.nix
ameertaweel: Thanks for the advice. I used `wrap` instead of `apply`.

BirdeeHub:
and also thats probably a good catch to use vim-full there too

wrapperModules/v/vim/check.nix
ameertaweel: Yesterday I didn't get to finish the tests, but I pushed my incomplete work with a WIP in the commit message. I probably should've wrote a comment on the PR or something so that you don't waste time reviewing it yet.

BirdeeHub: Its fine. The plugin thing I just was going to leave up to you and then I was doing something else and realized what it was.

But Im happy because the makeCustomizable function gave me an excuse to put knowledge I thought was going to be useless to surprisingly good use.

Now we can finally add a python wrapper module and stuff like that without rewriting the world! A lot of stuff probably have a python input people will want to configure. Could be cool to just stick a `type = wlib.types.subWrapperModule wlib.wrapperModules.python3;` there for those programs, and same for lua. And theres also a lot of other stuff like vim that have extra functions like that. It will be a good addition.

ameertaweel:
The Git part is not a problem for me, I'm quite good with Git.

But the `makeCustomizable` thing. I took more time than I'd like to admit to wrap my head around it. But it's so elegant I'm glad that I took the time to understand it. I think I just leveled-up my Nix understanding.

BirdeeHub:
If you understood the implementation itself, then you figured it out way faster than I did. But even just how lib.makeOverridable works is crazy to figure out the first time. Also, the `finalAttrs:` thing for derivations with overrideAttrs and how that works as well.

Took me several months to understand that one. And the first time I messed with it it was just a single alias that I wanted to add so that callPackage wouldnt shadow it.

But its how the `.override` thing works. I just made it generalized.

Actually the .extendModules thing also works similarly but with modules.

It internally collects a list, and then each one is just that list applied in a chain. But nixpkgs does that. I just expose it as an option that submodules can see so that you can use it when you use it as one. They helpfully gave it to us in our module arguments.

Also, the value of `config.package` you receive when you check is the configured one, the overrides are applied by the `apply` field of `config.package`'s option

My big skill level-up from this repository is that I finally understand the module system, at least in its majority.

Flakes took me like, a day from initializing the generated config on a new machine the first time. Im still learning stuff about the module system.

ameertaweel:
I took so much time understanding the override mechanisms in Nixpkgs last year. I wanted to create a Lua overlay that applies to all Lua interpreter versions.

I wrote what I learned in this Wiki section:
https://wiki.nixos.org/wiki/Lua#Override_a_Lua_package_for_all_available_Lua_interpreters

So I'm already quite familiar with the topic.

BirdeeHub:
Thats useful actually. I knew some ways to do that but I was wondering THE way to do that.

ameertaweel:
I remember I asked more experienced Nixers on Reddit to give me feedback on my Lua thing, but I didn't get much feedback. So I'm not even sure if I have stupid mistakes in it or not.

But I guess I can trust it more since you didn't find stupid mistakes with it.

BirdeeHub:
I knew there was some way to do it that involved an overlay, but I was doing it by going to the package sets and calling .extend and it was a little awkward? That also works though it seems? But you only get back the package set, so then you have to go to each interpreter and update over it.

Basically the way I have been doing it has been kinda bad so this is good, not that I have done it often.

ameertaweel:
I see. We probably went through a similar path dealing with the same issue. I arrived at my solution precisely to avoid updating each interpreter.

PR #230, opened 2026-01-24
feat(wrapperModules.claude-code): add agents and  pluginDirs options

vinnymeller:
- adds `agents` option to allow defining custom agents
- adds `pluginDirs` option to allow defining plugin folder locations 
  - just accepts strings, so can use this to pass both raw, non-nix paths to the cli flag, or e.g. flake inputs 

tested manually with config using my branch here: https://github.com/vinnymeller/nixos-config/commit/9167e877cd6e09bdece763e92195d8e1882fb650

edit: draft again because i want to add a couple more things

edit 2: nevermind in the end i didnt want to add any more lol

### DIFF
--- a/wrapperModules/c/claude-code/module.nix
+++ b/wrapperModules/c/claude-code/module.nix
@@ -13,23 +13,25 @@ in
 
   options = {
 
-    settings = lib.mkOption {
+    agents = lib.mkOption {
       type = jsonFmt.type;
       default = { };
       description = ''
-        Claude Code settings
+        Custom agents to add to Claude Code.
 
-        These settings will override local, project, and user scoped settings.
-
-        See <https://code.claude.com/docs/en/settings>
+        See <https://code.claude.com/docs/en/sub-agents>
       '';
       example = {
-        includeCoAuthoredBy = false;
-        permissions = {
-          deny = [
-            "Bash(sudo:*)"
-            "Bash(rm -rf:*)"
+        code-reviewer = {
+          description = "Expert code reviewer. Use proactively after code changes.";
+          prompt = "You are a senior code reviewer. Focus on code quality, security, and best practices.";
+          tools = [
+            "Read"
+            "Grep"
+            "Glob"
+            "Bash"
           ];
+          model = "sonnet";
         };
       };
     };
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Ooo nice you used the list form for flags to add `--plugin-dir somedir --plugin-dir someotherdir`

Im not sure any of the modules have used that feature yet actually.

BirdeeHub:
This git diff is really weird it makes it look like you either removed the settings option or just added it but you did not.

Not an issue, just mentioning it.

wrapperModules/c/claude-code/module.nix
```diff
@@ -52,6 +54,43 @@ in
       };
     };
 
+    pluginDirs = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
```
BirdeeHub: use `wlib.types.stringable` here so that it also accepts derivations maybe?

Is the required path ever the top level of a derivation?

If it sometimes is, then you should use `wlib.types.stringable` instead of `lib.types.str` so that you don't need to always stringify stuff first when using it, otherwise this is fine.

BirdeeHub:
Awesome, gonna let the tests run, and then it looks good! Thank you!

wrapperModules/c/claude-code/module.nix
vinnymeller: nice, i missed that type - yes that's actually the most common pattern for user-contributed plugins

the claude plugins repo has them in `plugins/<pluginName>`, but the recommendation for publishing your own plugins is to have everything at the root

tested with latest commit just putting `inputs.<plugin-input>` and it works :+1: 

also since github doesnt want to let me reply to the diff thing, i moved it just to make the options alphabetical but the diff  does end up looking odd sorry about that

BirdeeHub: Yeah, of all the types that I added, that one is one of the most useful to be honest. I find myself randomly using it in my config too. `wlib.types.spec` is cool and all, but I use this more than I use `lib.types.str` now haha

And that does make the diff make sense actually. It wasn't an issue just looked funny. I just hit view file instead, wasn't a problem.

BirdeeHub:
Also, unrelated

I only just now realized I have `wlib.types.nonEmptyline` not `wlib.types.nonEmptyLine` fml... It might stay like that... IDK. Do I deprecate that? If I do I could leave the old one around with a warning basically indefinitely, its not in anything's way.


```nix
  /**
    A single-line, non-empty string
  */
  nonEmptyLine = lib.mkOptionType {
    name = "nonEmptyLine";
    description = "non-empty line";
    descriptionClass = "noun";
    check =
      x:
      lib.types.str.check x && builtins.match "[ \t\n]*" x == null && builtins.match "[^\n\r]*" != null;
    inherit (lib.types.str) merge;
  };
  nonEmptyline = lib.warn "`lib.types.nonEmptyline` is deprecated due to having a mistake in its name, use `lib.types.nonEmptyLine`" (wlib.types.nonEmptyLine);
```

I could leave it like this literally forever with no downside

vinnymeller:
i might have never noticed if you didnt point it out, but now ill never unsee it, so i vote for your comment :laughing:

BirdeeHub:
The font in the docs makes it really hard to tell right?! And for the code that used it I was copy pasting, didnt notice

Edit: done!

PR #234, opened 2026-01-25
feat(lib.makeCustomizable): generalized lib.makeOverridable

BirdeeHub:
useful for packages that have function attributes for configuration already but which can only be called once. Pairs well with `config.overrides` option.

### DIFF
--- a/lib/lib.nix
+++ b/lib/lib.nix
@@ -364,4 +364,153 @@ in
   */
   escapeShellArgWithEnv = arg: ''"${lib.escape [ ''\'' ''"'' ] (toString arg)}"'';
 
+  /**
+    Wrap a function (or callable attribute set) to make it customizable via a
+    named override entry.
+
+    A slightly generalized version of `nixpkgs.lib.makeOverridable`, with explicit
+    support for:
+    - custom override names
+    - configurable argument-merging semantics
+    - preserving override entry points across common derivation-like patch
+      functions (e.g. `override`, `overrideAttrs`, `overrideDerivation`)
+
+    This helper turns `f` into a functor that:
+    - Preserves the original argument signature of `f`
+    - Exposes an override function under the attribute `${name}`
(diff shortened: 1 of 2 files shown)


lib/lib.nix
```diff
@@ -364,4 +364,153 @@ in
...
+      luaEnv = wlib.makeCustomizable
+        "withPackages"
+        { mergeArgs = og: new: lp: og lp ++ new lp; }
+        pkgs.luajit.withPackages
+        (lp: [ lp.inspect ]);
+
+      # inspect + cjson
+      luaEnv2 = luaEnv.withPackages (lp: [ lp.cjson ]);
+      # inspect + cjson + luassert
+      luaEnv3 = luaEnv.withPackages (lp: [ lp.luassert ]);
```
ameertaweel, line 448: I think this should be:

```nix
# luaEnv3 = luaEnv2.withPackages (lp: [ lp.luassert ]);
```

Am I right?

BirdeeHub: Yes i fixed it in the next pr.

Missed it.

Your understanding is correct. It works like .override and makes sure .override and friends don't remove it

Except you can choose the name of the output field and its merge behavior

and you can patch extra override functions if needed for example if more were added, such as customize, if you wanted to add another one you could add customize to the list of patches in the second arg

BirdeeHub: Im considering submitting it to nixpkgs and using it to implement lib.makeOverridable there, but I am not sure yet, but it is definitely useful here.

BirdeeHub: Also, if you were wondering how I had that ready so fast, [I have been staring at this in nixCats for like 2 years](https://github.com/BirdeeHub/nixCats-nvim/blob/43fbf4d12b0a613f1a792503da4bb2bf270173c7/utils/lib.nix#L73-L127) and suddenly knew what to do with that knowledge when I looked at your PR last night.

PR #238, opened 2026-01-27
feat(wrapperModules.btop): support themes

ameertaweel:
Added support for Btop themes similar to [how home-manager does it](https://github.com/nix-community/home-manager/blob/63a87808f5f9b6e4195a1d33f6ea25d23f4aa0df/modules/programs/btop.nix#L131-L141). Based on our discussion in https://github.com/BirdeeHub/nix-wrapper-modules/pull/204.

### DIFF
--- a/wrapperModules/b/btop/module.nix
+++ b/wrapperModules/b/btop/module.nix
@@ -25,7 +25,6 @@ let
     name: theme:
     if builtins.isPath theme || lib.isStorePath theme then theme else pkgs.writeText "btop.theme" theme;
 
-  binNameEsc = lib.escapeShellArg config.binName;
   themesDir = "${builtins.placeholder "out"}/themes";
 in
 {


wrapperModules/b/btop/module.nix
```diff
@@ -50,11 +52,80 @@ in
...
+        Custom Btop themes.
+      '';
+    };
   };
 
+  config.drv.buildPhase =
+    let
+      themes = builtins.mapAttrs mkBtopTheme config.themes;
+      cpCommands = lib.mapAttrsToList (
+        name: theme: "cp ${theme} ${themesDir}/${lib.escapeShellArg name}.theme"
```
BirdeeHub, line 114: Should this be ln -s ?

I actually don't know if one is better than the other. I guess cp is fine.

wrapperModules/b/btop/module.nix
```diff
@@ -50,11 +52,80 @@ in
         for options.
       '';
     };
+
+    themes = lib.mkOption {
+      type = types.lazyAttrsOf (types.either types.path types.lines);
```
BirdeeHub, line 56: I was going to as why lazyAttrsOf here but I went and checked and they have it too so I would assume they were running into a specific issue?

wrapperModules/b/btop/module.nix
```diff
@@ -27,18 +20,27 @@ let
           builtins.toString v;
     } " = ";
   };
+
+  mkBtopTheme =
+    name: theme:
+    if builtins.isPath theme || lib.isStorePath theme then theme else pkgs.writeText "btop.theme" theme;
+
+  binNameEsc = lib.escapeShellArg config.binName;
+  themesDir = "${builtins.placeholder "out"}/${binNameEsc}-themes";
```
BirdeeHub: `themesDir = "${builtins.placeholder "out"}/${config.binName}-themes";`

This shouldnt be escaped yet here because flags also will

It works because you didnt have an illegal character in your test binName string probably

wrapperModules/b/btop/module.nix
```diff
@@ -50,11 +52,80 @@ in
...
+      runHook preBuild
+      mkdir -p ${themesDir}
+      ${lib.concatStringsSep "\n" cpCommands}
+      runHook postBuild
+    '';
+
   config.package = lib.mkDefault pkgs.btop;
   config.flags = {
     "--config" = pkgs.writeText "btop.conf" (toBtopConf config.settings);
+    "--themes-dir" = themesDir;
```
BirdeeHub, line 127:
```nix
"--themes-dir" = {
  data = themesDir;
  esc-fn = lib.escapeShellArg;
  # we should probably force it to use this one always regardless of default escaping function
  # but this one is up to you, it already uses lib.escapeShellArg by default
};
```

wrapperModules/b/btop/module.nix
BirdeeHub, line 114:
```suggestion
        name: theme: "cp ${theme} ${lib.escapeShellArg themesDir}/${lib.escapeShellArg name}.theme"
```

wrapperModules/b/btop/module.nix
ameertaweel: Can't I solve this by setting `esc-fn` to `x: x`? I just didn't want to define the themes directory in two places separately.

ameertaweel: I had an idea. Why are we using `binName` in the themes directory name in the first place? We can just remove it and not worry about escaping.

BirdeeHub: > I had an idea. Why are we using `binName` in the themes directory name in the first place? We can just remove it and not worry about escaping.

To be honest, it probably isnt needed

But basically, symlinkScript module doesn't have this option, but if someone wanted to install multiple versions of the package like one can with the neovim module, then this would prevent them from colliding.

But I suppose it isn't needed here because this wrapper module doesn't have that feature anyway, and it isnt something I plan to have be a feature built into modules.symlinkScript because it is not always possible.

So, yes. That is fine.

wrapperModules/b/btop/module.nix
```diff
@@ -27,18 +20,27 @@ let
           builtins.toString v;
     } " = ";
   };
+
+  mkBtopTheme =
+    name: theme:
+    if builtins.isPath theme || lib.isStorePath theme then theme else pkgs.writeText "btop.theme" theme;
+
+  binNameEsc = lib.escapeShellArg config.binName;
```
BirdeeHub: This can be removed now too I suppose.

Yeah its fine to just do `$out/themes` like you have now.

The other one was overengineering tbh

So, other than this unused variable, I'm ok with this.

```suggestion
```

BirdeeHub: I hit the button on accident but its not like you were going to delete the line differently so, that is ok I guess? Apologies.

ameertaweel: Indeed I would've done the same thing. No worries.

ameertaweel: I should setup a Nix LSP or something. I'm sure I wouldn't have missed the unused variable that way.

PR #240, opened 2026-01-27
feat(wrapperModules.yazi): init

apetrovic6:
Add a wrapper for yazi terminal file manager.

Haven't done much with wrappers in Nix, so I've took some inspiration from the existing wrapper modules in the repo.

### DIFF
--- a/wrapperModules/y/yazi/module.nix
+++ b/wrapperModules/y/yazi/module.nix
@@ -8,355 +8,428 @@
 let
   tomlFmt = pkgs.formats.toml { };
 
-  yazi = tomlFmt.generate "yazi" config.yazi;
-  keymap = tomlFmt.generate "keymap" config.keymap;
-  theme = tomlFmt.generate "theme" config.theme;
-  vfs = tomlFmt.generate "vfs" config.vfs;
+  yazi = tomlFmt.generate "yazi" config.settings.yazi;
+  keymap = tomlFmt.generate "keymap" config.settings.keymap;
+  theme = tomlFmt.generate "theme" config.settings.theme;
+  vfs = tomlFmt.generate "vfs" config.settings.vfs;
+  package = tomlFmt.generate "package" config.settings.package;
 in
 {
   imports = [ wlib.modules.default ];
 
-  options = {
-    yazi = {
-      mgr = lib.mkOption {
-        type = tomlFmt.type;
-        description = ''
-          Manager settings
-          See <https://yazi-rs.github.io/docs/configuration/yazi#mgr>
-        '';
-        default = { };
-      };
+  options.settings = {
+    yazi = lib.mkOption {
+      default = { };
 
-      preview = lib.mkOption {
-        type = tomlFmt.type;
-        description = ''
-          Preview settings
-          See <https://yazi-rs.github.io/docs/configuration/yazi#preview>
-        '';
-        default = { };
-      };
+      description = ''
+        Content of yazi.toml file.
+        See the configuration reference at <https://yazi-rs.github.io/docs/configuration/yazi>
+      '';
 
-      opener = lib.mkOption {
-        type = tomlFmt.type;
-        description = ''
-          Opener settings
-          See <https://yazi-rs.github.io/docs/configuration/yazi#opener>
-        '';
-        default = { };
-      };
+      type = lib.types.submodule {
+        freeformType = tomlFmt.type;
+        options = {
+          mgr = lib.mkOption {
+            type = tomlFmt.type;
+            description = ''
+              Manager settings
+              See <https://yazi-rs.github.io/docs/configuration/yazi#mgr>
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Awesome! let me take a look!

BirdeeHub:
Are these all the valid settings for those files?

We should make `options.yazi` `options.keymap` etc. into freeform submodules like so:

```nix
yazi = lib.mkOption {
  default = { };
  description = ''
    content of the yazi.toml file.
    
    See configuration reference at ....
  '';
  type = lib.types.submodule {
    freeformType = tomlFmt.type;
    options = {
      # the existing options from `options.yazi` here so that they still get documented
    };
  };
};
```

That way if they add another setting people can still use it even before you get to add it to here.

I also have some improvements for the test if possible, give me a moment to give you good instruction on that, but `yazi --debug` allows you to view the paths you added theoretically.

Also from the output of that, it appears there is also another configuration file, which would be nice to add.

`Package          : /home/birdee/.config/yazi/package.toml`

If you don't want to list out all the options, it is also acceptable to use a freeform type submodule like I showed and only list 2 or so major ones, but since you already listed the others you should leave at least the majority the ones you have listed already.

wrapperModules/y/yazi/check.nix
```diff
@@ -0,0 +1,11 @@
+{
```
BirdeeHub, line 1:
```nix
{
  pkgs,
  self,
}:
let
  yaziWrapper = self.wrappers.yazi.apply { inherit pkgs; };
  cfgdir = yaziWrapper.env.YAZI_CONFIG_HOME.data;
in
pkgs.runCommand "yazi-test" { } ''
  "${yaziWrapper.wrapper}/bin/yazi" --debug | grep -q "${cfgdir}"
  touch $out
''
```

It shows us the files we added in the --debug output so we can just check for that.

wrapperModules/y/yazi/module.nix
```diff
@@ -0,0 +1,388 @@
...
+  yazi = tomlFmt.generate "yazi" config.yazi;
+  keymap = tomlFmt.generate "keymap" config.keymap;
+  theme = tomlFmt.generate "theme" config.theme;
+  vfs = tomlFmt.generate "vfs" config.vfs;
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    yazi = {
```
BirdeeHub: First change is that these should be freeform type submodule options.

```nix
options.yazi = lib.mkOption {
  default = { };
  description = ''
    content of the yazi.toml file.
    
    See configuration reference at ....
  '';
  type = lib.types.submodule {
    freeformType = tomlFmt.type;
    options = {
      # the existing options from `options.yazi` here so that they still get documented
    };
  };
};
```

The second change is more difficult.

The problem is, there is 1 more file yazi looks for called package.toml

But we cant do `options.package = lib.mkOption ...` because that name is already taken.

I would somewhat prefer if we didnt have 1 `packageToml` option and then the rest are just normal?

What if we did

```nix
options."yazi.toml" = lib.mkOption {
  default = { };
  description = ''
    content of the yazi.toml file.
    
    See configuration reference at ....
  '';
  type = lib.types.submodule {
    freeformType = tomlFmt.type;
    options = {
      # the existing options from `options.yazi` here so that they still get documented
    };
  };
};
```

We could also do camel case too? That is also fine? But there is precedence for doing it this way in a few of the other modules so far. It is a bit more typing with the quotes though. Less convenient. Thoughts? Or maybe we should do

```nix
options.settings = {
  yazi = lib.mkOption {
    default = { };
    description = ''
      content of the yazi.toml file.
    
      See configuration reference at ....
    '';
    type = lib.types.submodule {
      freeformType = tomlFmt.type;
      options = {
        # the existing options from `options.yazi` here so that they still get documented
      };
    };
  };
};
```

Either would be fine with me, feel free to pick the way that suits you best for that, but we should have options for that file since we are providing the directory as a unit.

And because it will affect the naming of the other options due to this, we should do it now rather than after the module has been added.

apetrovic6: Hi, thanks for checking out my PR.

I've did the changes that you've mentioned. I think I've understood correctly what you've meant.

Regarding the package.toml, as you've said packageToml would be quite different from the rest of the module, having to type "package.toml", "yazi.toml' etc. would just be annoying.

So I think that your proposition of wrapping the options in a settings block is a good choice.
Also, I've added the option for the package.toml.

Thanks!

BirdeeHub:
You did understand what I meant, thank you!

Making them freeformType submodules means they allow extra values and all extra values that you pass are typechecked according to the freeformType value.

This means if there are any settings you are missing either now or one day, it will still be usable even if you haven't noticed and came to update it yet.

Anyway, it looks good to me now, thank you for your time and contribution!

PR #262, opened 2026-02-01

BirdeeHub:

### DIFF
--- a/docs/flake.lock
+++ b/docs/flake.lock
@@ -2,16 +2,16 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1769268028,
-        "narHash": "sha256-mAdJpV0e5IGZjnE4f/8uf0E4hQR7ptRP00gnZKUOdMo=",
+        "lastModified": 1769988320,
+        "narHash": "sha256-2G6aLzXxnbH/qLG/H9+c2EmkU0Re4D6aDLkrE1bVIqo=",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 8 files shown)


BirdeeHub:
something is trying and failing to build something that depends on both swift and a broken python package. But only on mac.

Unsure what, fairly sure whatever it is would be a bug in the package itself from nixpkgs because the modules themselves definitely do not require python or swift.

Maybe the toml format type depends on a bit of python (it depends on remarshal) but I would be surprised if nixpkgs allowed something like that which is used everywhere to be the problem. 

And it doesn't seem to be a configuration issue for whatever package this is, it is failing the build which is just a drv from nixpkgs.

Edit: ok, well, claude-code and mpv are definitely broken right now on mac.

`claude-code` is the one pulling rocksdb and failing on the broken `python3.13-rocksdict-0.3.29`, and `mpv` is the one building `swift` from source and failing.

BirdeeHub:
Do I leave these tests disabled and push this because users supply their own pkgs so the version here is not important, or do I re enable them and just wait for nixpkgs to fix those 2?

I guess since the version here doesn't really matter for users, it is fine to wait?

BirdeeHub:
trying master for mpv as requested, it caused nushell to fail though. So now trying again with nushell disabled on master so that I can check mpv.

Edit: ok, so `mpv` is working on master. `claude-code` is actually also working on master. `nushell` seems like it is not though.

BirdeeHub:
I rolled it back to nixpkgs-unstable but it turns out the mpv fixes have not made it there yet, so

TODO: roll this back to nixpkgs-unstable before you merge this.

PR #270, opened 2026-02-03
ci(checks-docs): moved tests to ci flake. renamed docs/ -> ci/

BirdeeHub:
to run the tests, use `nix flake check -Lv ./ci`

to build the docs, use `nix run ./ci`

This also tests that the docs worked as part of the checks now.

`wlib.checks` now contains `wlib.checks.helper.*` and `wlib.checks.wrapper.*` however I am quite confident nobody outside of this repository is interacting with `wlib.checks` as it is not documented and does nothing useful for users other than being ran in github actions to catch bugs.


BirdeeHub:
oh interesting it can't pull even the name for the docs if the system doesn't match...

I feel like there is probably a real fix for that.

Obviously the module isnt supposed to work on that system, so the fix should be something that makes the docgen not try to fill in the default? Or, the name is there anyway and we aren't building it, so maybe allowUnsupportedSystem for the docs?

I should also add supported meta.platforms to the docs for the wrapper modules maybe? But that will be a second PR.

BirdeeHub:
Also, if the cache action doesnt work as well as hoped and makes it always take 20 minutes then I will be removing it again, but I'm going to give it another chance.

Because the tests themselves take more like 1 minute on the runner, so it is barely worth it at the moment.

Edit: nope. Even when it works and only updates 1 thing it is still slower to upload the delta than just running all the tests again currently. We don't really "build" stuff here, other than some config files and wrapper scripts so everything just comes from the nixpkgs cache.

PR #285, opened 2026-02-05
feat(wrapperModules.zsh): init

FluxZA:
Create a wrapper module for zsh

### DIFF
--- a/wrapperModules/z/zsh/module.nix
+++ b/wrapperModules/z/zsh/module.nix
@@ -18,8 +18,9 @@ let
     For details regarding the rc files see <https://zsh.sourceforge.io/Doc/Release/Files.html#Startup_002fShutdown-Files>.
 
     This wrapper provides two methods of defining your local rc files. You can specify a directory
-    which contains the files, or you can specify each file directly. These two options cannot be used
-    together. If they are both defined, only the directory with its contained files will be used.
+    which contains the files, or you can specify each file directly. These two options can be used
+    together, in which case the files from the directory (specified using the `zdotdir` option) are
+    sourced before the individually specified files.
   '';
 in
 {
@@ -31,64 +32,194 @@ in
       type = types.bool;
       default = false;
       description = ''
-        Don't use the global rc files.
+        Set the option for zsh to skip loading system level rc files. The system level zshenv
+        file cannot be skipped.
       '';
+    };
+
+    zshAliases = lib.mkOption {
+      type = types.attrsOf (types.nullOr wlib.types.stringable);
+      default = { };
+      description = ''
+        An attribute set that maps aliases (the top level attribute names in this option) to command
+        strings or directly to build outputs. Aliases mapped to null are ignored. 
+
+        These aliases are created before any of the rc file options are sourced, therefore,
+        aliases specified in those options will override the aliases specified in this option.
+      '';
+      example = {
+        l = null;
+        ll = "ls -l";
+      };
     };
 
     zdotdir = lib.mkOption {
-      type = types.nullOr (types.either types.path types.str);
+      type = types.nullOr wlib.types.stringable;
       default = null;
       description = ''
-        WIP
+        Direct or string path to a directory containing rc files. The following files will be sourced from
+        this directory if they exist: `.zshenv`, `.zshrc`, `.zlogin` and `.zlogout`.
       '';
     };
 
-    zdotFiles = lib.mkOption {
+    zshenv = lib.mkOption {
+      type = types.nullOr (wlib.types.file pkgs);
       default = null;
-      type = types.nullOr (
-        types.submodule {
-          options = {
-            zshenv = lib.mkOption {
-              type = wlib.types.file pkgs;
-              default.content = "";
-              description = ''
-                WIP
-              '';
-            };
-            zshrc = lib.mkOption {
-              type = wlib.types.file pkgs;
-              default.content = "";
-              description = ''
-                WIP
-              '';
-            };
-            zlogin = lib.mkOption {
-              type = wlib.types.file pkgs;
-              default.content = "";
-              description = ''
-                WIP
-              '';
-            };
-            zlogout = lib.mkOption {
-              type = wlib.types.file pkgs;
-              default.content = "";
-              description = ''
-                WIP
-              '';
-            };
-          };
-        }
-      );
+      description = ''
+        Specifies a file which will be sourced as part of the local `.zshenv` file.
+
+        For details on how to specify the file, see the `file` type description in the wrapper lib.
+      '';
+    };
+    zshrc = lib.mkOption {
+      type = types.nullOr (wlib.types.file pkgs);
+      default = null;
+      description = ''
+        Specifies a file which will be sourced as part of the local `.zshrc` file.
+
+        For details on how to specify the file, see the `file` type description in the wrapper lib.
+      '';
+    };
+    zlogin = lib.mkOption {
+      type = types.nullOr (wlib.types.file pkgs);
+      default = null;
+      description = ''
+        Specifies a file which will be sourced as part of the local `.zlogin` file.
+
+        For details on how to specify the file, see the `file` type description in the wrapper lib.
+      '';
+    };
+    zlogout = lib.mkOption {
+      type = types.nullOr (wlib.types.file pkgs);
+      default = null;
+      description = ''
+        Specifies a file which will be sourced as part of the local `.zlogout` file.
+
+        For details on how to specify the file, see the `file` type description in the wrapper lib.
+      '';
     };
   };
 
   config =
     let
+      # Name the dirctory which is created as the zdotdir in the wrapper output
       zdotFilesDirname = "wrapped_zdot";
-      zdotdirEnv =
-        lib.trivial.warnIf (config.zdotdir != null && config.zdotFiles != null)
-          "Using both zdotdir and zdotFiles options is not compatible. Only zdotdir will be used."
-          (if config.zdotdir != null then config.zdotdir else "${placeholder "out"}/${zdotFilesDirname}");
+
+      zshAliases = builtins.concatStringsSep "\n" (
+        lib.mapAttrsToList (k: v: "alias -- ${k}=${lib.escapeShellArg v}") (
+          lib.filterAttrs (k: v: v != null) config.zshAliases
+        )
+      );
+
+      baseZshenv = /* bash */ ''
+        # zsh-wrapped zshenv: DO NOT EDIT -- this file has been generated automatically.
+        # This file is read for all shells.
+
+        # Ensure this is only run once per shell
+        if [[ -v __WRAPPED_ZSHENV_SOURCED ]]; then return; fi
+        __WRAPPED_ZSHENV_SOURCED=1
+
+        # Cover some of the work done by zsh NixOS program if it is not installed
+        if [[ ! (-v __ETC_ZSHENV_SOURCED) ]]
+        then
+          HELPDIR="${pkgs.zsh}/share/zsh/$ZSH_VERSION/help"
+
+          # Tell zsh how to find installed completions.
+          for p in ''${(z)NIX_PROFILES}; do
+              fpath=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions $p/share/zsh/vendor-completions $fpath)
+          done
+        fi
+
+        # Get zshenv from wrapped options if they exist
+        # zdotdir files must be sourced first to maintain documented override rules
+        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
+          if [[ -f "${config.zdotdir}/.zshenv" ]]
+          then
+            source "${config.zdotdir}/.zshenv"
+          fi
+        ''}
+        ${lib.optionalString (config.zshenv != null) /* bash */ ''
+          if [[ -f "${config.zshenv.path}" ]]
+          then
+            source "${config.zshenv.path}"
+          fi
+        ''}
+      '';
+
+      baseZshrc = /* bash */ ''
+        # zsh-wrapped zshrc: DO NOT EDIT -- this file has been generated automatically.
+        # This file is read for all shells.
+
+        # Ensure this is only run once per shell
+        if [[ -v __WRAPPED_ZSHRC_SOURCED ]]; then return; fi
+        __WRAPPED_ZSHRC_SOURCED=1
+
+        # zsh-wrapped defined aliases
+        ${zshAliases}
+
+        # Get zshrc from wrapped options if they exist
+        # zdotdir files must be sourced first to maintain documented override rules
+        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
+          if [[ -f "${config.zdotdir}/.zshrc" ]]
+          then
+            source "${config.zdotdir}/.zshrc"
+          fi
+        ''}
+        ${lib.optionalString (config.zshrc != null) /* bash */ ''
+          if [[ -f "${config.zshrc.path}" ]]
+          then
+            source "${config.zshrc.path}"
+          fi
+        ''}
+      '';
+
+      baseZlogin = /* bash */ ''
+        # zsh-wrapped zlogin: DO NOT EDIT -- this file has been generated automatically.
+        # This file is read for all shells.
+
+        # Ensure this is only run once per shell
+        if [[ -v __WRAPPED_ZLOGIN_SOURCED ]]; then return; fi
+        __WRAPPED_ZLOGIN_SOURCED=1
+
+        # Get zlogin from wrapped options if they exist
+        # zdotdir files must be sourced first to maintain documented override rules
+        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
+          if [[ -f "${config.zdotdir}/.zlogin" ]]
+          then
+            source "${config.zdotdir}/.zlogin"
+          fi
+        ''}
+        ${lib.optionalString (config.zlogin != null) /* bash */ ''
+          if [[ -f "${config.zlogin.path}" ]]
+          then
+            source "${config.zlogin.path}"
+          fi
+        ''}
+      '';
+
+      baseZlogout = /* bash */ ''
+        # zsh-wrapped zlogout: DO NOT EDIT -- this file has been generated automatically.
+        # This file is read for all shells.
+
+        # Ensure this is only run once per shell
+        if [[ -v __WRAPPED_ZLOGOUT_SOURCED ]]; then return; fi
+        __WRAPPED_ZLOGOUT_SOURCED=1
+
+        # Get zlogout from wrapped options if they exist
+        # zdotdir files must be sourced first to maintain documented override rules
+        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
+          if [[ -f "${config.zdotdir}/.zlogout" ]]
+          then
+            source "${config.zdotdir}/.zlogout"
+          fi
+        ''}
+        ${lib.optionalString (config.zlogout != null) /* bash */ ''
+          if [[ -f "${config.zlogout.path}" ]]
+          then
+            source "${config.zlogout.path}"
+          fi
+        ''}
+      '';
     in
     {
       package = lib.mkDefault pkgs.zsh;
@@ -96,17 +227,32 @@ in
       # Required for shebangs to work on MacOS
       wrapperImplementation = "binary";
 
+      # Allow use as a system/user shell
+      passthru.shellPath = "/bin/zsh";
+
       addFlag = lib.mkIf (config.skipGlobalRC) [ "-d" ];
 
-      drv.installPhase = lib.mkIf (config.zdotFiles != null) ''
-        mkdir $out/wrapped_zdot
-        cp ${config.zdotFiles.zshenv.path} $out/${zdotFilesDirname}/.zshenv
-        cp ${config.zdotFiles.zshrc.path} $out/${zdotFilesDirname}/.zshrc
-        cp ${config.zdotFiles.zlogin.path} $out/${zdotFilesDirname}/.zlogin
-        cp ${config.zdotFiles.zlogout.path} $out/${zdotFilesDirname}/.zlogout
+      drv.buildPhase = ''
+        runHook preBuild
+
+        mkdir $out/${zdotFilesDirname}
+        cat >$out/${zdotFilesDirname}/.zshenv <<'EOL'
+        ${baseZshenv}
+        EOL
+        cat >$out/${zdotFilesDirname}/.zshrc <<'EOL'
+        ${baseZshrc}
+        EOL
+        cat >$out/${zdotFilesDirname}/.zlogin <<'EOL'
+        ${baseZlogin}
+        EOL
+        cat >$out/${zdotFilesDirname}/.zlogout <<'EOL'
+        ${baseZlogout}
+        EOL
+
+        runHook postBuild
       '';
 
-      env.ZDOTDIR = zdotdirEnv;
+      env.ZDOTDIR = "${placeholder "out"}/${zdotFilesDirname}";
 
       meta.maintainers = [ wlib.maintainers.fluxza ];
       meta.description = {
--- a/wrapperModules/z/zsh/check.nix
+++ b/wrapperModules/z/zsh/check.nix
@@ -6,19 +6,31 @@
 let
   zshWrapped = self.wrappers.zsh.wrap {
     inherit pkgs;
+    zshAliases = {
+      testAlias = "echo Alias";
+    };
+    # zdotdir files contain both their own alias and the alias specified in the direct file options
+    # below to test override functionality
     zdotdir = ./test-zdotdir;
-    zdotFiles.zshrc.content = "alias test-zdot=\"echo Test Fail\"";
-  };
-  zshFilesWrapped = self.wrappers.zsh.wrap {
-    inherit pkgs;
-    zdotFiles.zshrc.content = "alias test-zdot=\"echo Test Success\"";
+    zshenv.content = "alias testZshenv=\"echo Zshenv\"";
+    zshrc.content = "alias testZshrc=\"echo Zshrc\"";
+    zlogin.content = "alias testZlogin=\"echo Zlogin\"";
   };
 
 in
 pkgs.runCommand "zsh-test" { } ''
   "${zshWrapped}/bin/zsh" --version | grep -q "${zshWrapped.version}"
-  "${zshWrapped}/bin/zsh" -c "echo \$ZDOTDIR" | grep -q "test-zdotdir"
-  "${zshWrapped}/bin/zsh" -c "ls \$ZDOTDIR"
-  "${zshWrapped}/bin/zsh" -ic testZdot | grep -q "Testing Success"
+
+  "${zshWrapped}/bin/zsh" -ic testAlias | grep -q "Alias"
+
+  "${zshWrapped}/bin/zsh" -ic testZshenv | grep -q "Zshenv"
+  "${zshWrapped}/bin/zsh" -ic testZdotenv | grep -q "Zdot env"
+
+  "${zshWrapped}/bin/zsh" -ic testZshrc | grep -q "Zshrc"
+  "${zshWrapped}/bin/zsh" -ic testZdotrc | grep -q "Zdot rc"
+
+  "${zshWrapped}/bin/zsh" -lc testZlogin | grep -q "Zlogin"
+  "${zshWrapped}/bin/zsh" -lc testZdotlogin | grep -q "Zdot login"
+
   touch $out
 ''
--- a/wrapperModules/z/zsh/test-zdotdir/.zlogin
+++ b/wrapperModules/z/zsh/test-zdotdir/.zlogin
@@ -1,0 +1,2 @@
+alias testZlogin="echo Zdot login"
+alias testZdotlogin="echo Zdot login"
--- a/wrapperModules/z/zsh/test-zdotdir/.zshenv
+++ b/wrapperModules/z/zsh/test-zdotdir/.zshenv
@@ -1,0 +1,2 @@
+alias testZshenv="echo Zdot env"
+alias testZdotenv="echo Zdot env"
--- a/wrapperModules/z/zsh/test-zdotdir/.zshrc
+++ b/wrapperModules/z/zsh/test-zdotdir/.zshrc
@@ -1,1 +1,2 @@
-alias testZdot="echo Test Success"
+alias testZshrc="echo Zdot rc"
+alias testZdotrc="echo Zdot rc"


FluxZA:
Hi. The tests are failing because my checks currently don't pass, I have a question regarding that below. I also have not finished documenting options.

Questions regarding the implementation:
- Do you think it is necessary to add an option to skip reading the /etc/zshenv file? I think it would be possible by overriding the configuration flags used to build the base package.
- I have not set options for common zsh options or features like aliases etc. This is because they would only be able to be added if the zdotFiles option is used, not the zdotdir option. Do you think this is fine or should I add them anyway with a caveat.

Questions regarding the checks:
- I wanted to test that the config in the store was being used, but from what I can see the test is running before the build is fully finished? Trying to `ls` the store directory fails claiming that it does not exist, however, when I check it manually after the tests stop, the directory exists and contains the file it should.

BirdeeHub:
> Do you think it is necessary to add an option to skip reading the /etc/zshenv file? I think it would be possible by overriding the configuration flags used to build the base package.

Not really no but for completeness maybe it could be offered one day, but overriding the base package means you rebuild it, and that is expensive.

> I have not set options for common zsh options or features like aliases etc. This is because they would only be able to be added if the zdotFiles option is used, not the zdotdir option. Do you think this is fine or should I add them anyway with a caveat.

Yeah this is a conundrum slightly.

How about this.

What if you secretly make your own zdotdir every time. If they supply zdotdir then you can source the files from it if they exist as it would from each of your files in your zdotdir, but that would allow you to ALSO accept things from the zdotFiles options.

Also you should link the zdotFiles rather than copying them, or just refer to them when you source them from your secret zdotdir (probably easier to just check for existence then source from the script), that way you can include impure paths.

Also

```nix
      type = types.nullOr (types.either types.path types.str);
```

```nix
    type = types.nullOr wlib.types.stringable;
```

> I wanted to test that the config in the store was being used, but from what I can see the test is running before the build is fully finished? Trying to ls the store directory fails claiming that it does not exist, however, when I check it manually after the tests stop, the directory exists and contains the file it should.

It was missing this condition, it might just be things like this.

```nix
      env.ZDOTDIR = lib.mkIf (config.zdotdir != null || config.zdotFiles != null) zdotdirEnv;
```


Because if you build with no arguments right now it refers to a place in placeholder out that doesnt get created. So its probably something of that nature. Your drv.installPhase is most definitely running before your test is, your test needs that whole zshWrapped derivation to resolve before it can get the path to it.

So, whatever is going wrong with the tests, it is not related to sequencing of stuff in the test being ran before stuff in the zshWrapped derivation. It is likely confusing because the test itself is a "build", but it is a runCommand which uses the result of zshWrapped. zshWrapped is computed already.

BirdeeHub:
I just did

```nix
    zdotdir = builtins.path { path = ./test-zdotdir; };
```

and it worked but IDK why that makes any sense.

Also

```nix
    zdotdir = "${./test-zdotdir}";
```

That also works too for some reason.

I think this is lazy trees.......?

```nix
      zdotdirEnv = lib.mkIf (config.zdotdir != null || config.zdotFiles != null) (
        lib.trivial.warnIf (config.zdotdir != null && config.zdotFiles != null)
          "Using both zdotdir and zdotFiles options is not compatible. Only zdotdir will be used."
          (
            if config.zdotdir != null then
              if builtins.isPath config.zdotdir then builtins.path { path = config.zdotdir; } else config.zdotdir
            else
              "${placeholder "out"}/${zdotFilesDirname}"
          )
      );
```

Weird but the above passes the `ls` test

However, you should make your own zdotdir and source the stuff we provide, which hopefully will make this issue go away on its own? We will see.

The path to the test-zdotdir is in fact making it into the derivation. And I suppose your store would have that path. But its context somehow isnt making it.

But all I do is pass them to lib.escapeShellArg and then write them into the buildCommand of the drv. The path you are passing does in fact make it into the drv build command. It really should be provisioning this I feel like but maybe I am not understanding how it works?

---

Im going to swap back to upstream and see if this still happens

---

Edit: Ok, so it also happens on upstream. So it is not lazy-trees

But something is making it not provision that path, despite the fact that that path is very much making it into the derivation and is a nix path, and is only being passed through lib.escapeShellArg (this happens on all the backends it is not specific to the binary one)

But regardless, hopefully this goes away on its own once you change the design to always provide an internal zdotdir and source the users stuff yourself from there. If not, then call builtins.path on it if it is a path. I am not sure why this is needed. It seems like when you do stuff like lib.generators.toLua or builtins.toJSON it does this for you, but if you are just using toString it does not necessarily?

I wonder if this is something that I can do in wlib.types.stringable for convenience for the user?

---

Edit:

Should I do this?

```nix
  stringable = lib.mkOptionType {
    name = "stringable";
    descriptionClass = "noun";
    description = "str|path|drv";
    check = lib.isStringLike;
    merge =
      loc: defs:
      let
        res = lib.mergeEqualOption loc defs;
      in
      if builtins.isPath res then builtins.path { path = res; } else res;
  };
```

FluxZA:
Thank you for the quick feedback!

I like the idea of using a secret zdotdir that sources the given files. If I use the approach of sourcing the zdotFiles, when they aren't impure paths should I check for existence in the store? My understanding of files in vs out of the store is still not the best.

FluxZA:
I think it would be helpful to include that as part of stringable for the users, as some programs might need to test against directories.

BirdeeHub:
A string may be an impure path, like `"/home/flux/mynixos"` or whatever. But if they are path types, they are pure paths, they just might not be provisioned apparently. I think this is because `/home/flux/mynixos` is a path, but not a pure path. But if something is providing something like that, they are already using --impure, so it still should be safe to call builtins.path on it?

builtins.path can be called multiple times. And you dont have to use wlib.types.stringable but where people would use it they probably would expect that thing to get provisioned. So I think I will do that.

FluxZA:
> A string may be an impure path, like `"/home/flux/mynixos"` or whatever. But if they are path types, they are pure paths, they just might not be provisioned apparently. I think this is because `/home/flux/mynixos` is a path, but not a pure path. But if something is providing something like that, they are already using --impure, so it still should be safe to call builtins.path on it?

is that because an absolute path is handled differently to a relative path? I thought that specifying a relative path is a pure path?

Also, what is the difference between provisioning vs a path existing in the store?

BirdeeHub:
And, you shouldnt check that they are in the store no. They explicitly are wlib.types.stringable which means, anything that I can interpolate into a string, which includes an impure paths. If you want to test it with a path not already in the store, you should check that your secret one is there, and for the other options then you should test it with some paths that are "impure" inside the derivation if you want to, i.e. use a relative path within the test or something.

And yeah, ngl this one was a surprise to me as well, I did not know store paths worked like that. Oddly, you can still import from them even if they are like that, but they arent really "realized" yet, which is weird. This has to be some sort of optimization that allows it to not scan stuff in nixpkgs.

BirdeeHub:
> > A string may be an impure path, like `"/home/flux/mynixos"` or whatever. But if they are path types, they are pure paths, they just might not be provisioned apparently. I think this is because `/home/flux/mynixos` is a path, but not a pure path. But if something is providing something like that, they are already using --impure, so it still should be safe to call builtins.path on it?
> 
> is that because an absolute path is handled differently to a relative path? I thought that specifying a relative path is a pure path?
> 
> Also, what is the difference between provisioning vs a path existing in the store?

`"/home/flux"` is a string, it is a string that looks like a path but wont be provided by nix

`"./somedir"` and this is the same. It will look in wherever that is at runtime. Nix has no idea where that is.

Colloquially, the above 2 are known as impure paths. This is because nobody acknowledges the next form of impure path:

`/home/flux` is an impure path. If you dont use --impure and you are using a flake this will throw an error if you try to use it.

`./somedir` this is a store path. you can call `import` on it or check for its existence without an IFD. But apparently it isnt always realized. Who knew? I didnt. I mean, I knew you could turn it into a string and then call unsafeDiscardStringContext on it to make sure that it does not get provisioned. But I kinda thought that otherwise it always would be? Generally one operates with these like they are always there, and are shocked when in some odd scenario like now they are not, and then call `builtins.path` on them to fix it.

FluxZA:
Okay, thanks for clarifying.

BirdeeHub:
To the best of my ability anyway :laughing:

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/286

Gonna add this.

This also takes care of it for `wlib.types.file` as well. its path value is of the stringable type

Im not really sure if this needs to be done at the makeWrapper level or not yet. Maybe just a different default escapeShellArg that also calls this if it is a path first could be used for esc-fn? Regardless it should still be done in that type as well, and for our purposes here is enough.

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,116 @@
...
+    skipGlobalRC = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Don't use the global rc files.
+      '';
+    };
+
+    zdotdir = lib.mkOption {
+      type = types.nullOr (types.either types.path types.str);
```
BirdeeHub:
```suggestion
      type = types.nullOr wlib.types.stringable;
```

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,116 @@
...
+
+      drv.installPhase = lib.mkIf (config.zdotFiles != null) ''
+        mkdir $out/wrapped_zdot
+        cp ${config.zdotFiles.zshenv.path} $out/${zdotFilesDirname}/.zshenv
+        cp ${config.zdotFiles.zshrc.path} $out/${zdotFilesDirname}/.zshrc
+        cp ${config.zdotFiles.zlogin.path} $out/${zdotFilesDirname}/.zlogin
+        cp ${config.zdotFiles.zlogout.path} $out/${zdotFilesDirname}/.zlogout
+      '';
+
+      env.ZDOTDIR = zdotdirEnv;
```
BirdeeHub: make sure to only make this if something was provided (although, now that you will always be providing one, that might not be relevant anymore) but regardless it should behave as much like normal if nothing was provided as possible

BirdeeHub:
Left the comments in the specific places they were mentioning. They may or may not be relevant by the time you finish implementing the suggestion where you source stuff yourself.

Anyway, hope I was helpful. Good luck!

FluxZA:
Thank you.

FluxZA:
One question I forgot to ask. Is there a way to run tests for a specific module only? I tried to look at the ci flake, but I didn't see any way to pass in arguments. Which now that I think of it makes sense as it is a flake.

BirdeeHub:
> One question I forgot to ask. Is there a way to run tests for a specific module only? I tried to look at the ci flake, but I didn't see any way to pass in arguments. Which now that I think of it makes sense as it is a flake.

`nix build ./ci#checks.<system>.wrapperModule-zsh`

Also with zsh that probably means ur gonna have to escape the `#`

`nix flake check -Lv ./ci` runs all the checks outputs of the flake with the checks in it. To run them individually you just build them.

I have the checks in a second flake so that I can do stuff like pull flake-parts to test that at some point, pull random stuff for docs but still run the docgen as part of the tests, mock things that are maybe complicated and need dependencies, etc. while leaving the top-level flake with just 1 input

FluxZA:
Perfect thank you, that makes sense.

FluxZA:
What do you think of this approach to building the files:
```nix
      baseZshenv = /* bash */ ''
        # zsh-wrapped zshenv: DO NOT EDIT -- this file has been generated automatically.
        # This file is read for all shells.

        # Ensure this is only run once per shell
        if [[ -v __WRAPPED_ZSHENV_SOURCED ]]; then return; fi
        __WRAPPED_ZSHENV_SOURCED=1

        # Cover some of the work done by zsh NixOS program if it is not installed
        if [[ ! (-v __ETC_ZSHENV_SOURCED) ]]
        then
          WRAPPER_MOCK_ETC_ZSHENV=1
          HELPDIR="${pkgs.zsh}/share/zsh/$ZSH_VERSION/help"

          # Tell zsh how to find installed completions.
          for p in ''${(z)NIX_PROFILES}; do
              fpath=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions $p/share/zsh/vendor-completions $fpath)
          done
        fi

        # Get zshenv from wrapped options if they exist
        ${lib.optionalString (config.zdotFiles != null) /* bash */ ''
          if [[ -f "${config.zdotFiles.zshenv.path}" ]]
          then
            source "${config.zdotFiles.zshenv.path}"
          fi
        ''}
        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
          if [[ -f "${config.zdotdir}/.zshenv" ]]
          then
            source "${config.zdotdir}/.zshenv"
          fi
        ''}
      '';
    in
    {
      drv.installPhase = ''
        mkdir $out/wrapped_zdot
        cat >$out/${zdotFilesDirname}/.zshenv <<'EOL'
        ${baseZshenv}
        EOL
      '';
```

BirdeeHub:
yeah

You would have to do the thing you did for .zshenv for all of them probably, but basically that yeah.

Looks like you got some (most? all? IDK) of the stuff from nix that one might want to source too with the completions thing.

You should use buildPhase instead of installPhase and you should put the runHook for pre and post build like

```nix
drv.buildPhase = ''
  runHook preBuild;
  thestuff;
  ...;
  runHook postBuild
'';
```

But yes that looks like a fine approach.

Also, you should set `config.passthru.shellPath = "/bin/zsh";` as discussed by [this discussion about bash](https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295)

FluxZA:
I have updated with changes as discussed.

When it comes to testing, I couldn't find a good way to test the `.zlogout` file, I know it works locally, but it is not covered by the check module.

I was trying to think of a way to package the history file with the module, but I couldn't think of an elegant solution. The idea was that the history would not be placed in the local home directory and could be updated on exit, but I couldn't find a good place to put it outside the store, which is not an option due to being read only. Would be interested to hear any ideas you have.

BirdeeHub:
I would think the history file should be handled by the user, they can set the location, and provision it either by copying it in their config if the file is not there, or if they are using home manager they can use `home.files.<name>.onChange` to do the copy. (home manager and nixos have activation scripts, which derivations do not have unless you are loading it as a shell)

---

This looks good actually.

---

One question.

Does `-d` solve the stuff from [this discussion](https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295)? Specifically [the stuff about zenoli's attempt at a zsh wrapper](https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295#discussioncomment-15814531)?

I am still trying to figure out the best way to handle that.... It might even require getting your hands dirty redefining wrapperFunction to make it create some sort of sourceable file for the variables to source from the config and only handle the args with the wrapper script itself....

Do you have any ideas on that which do not require such a thing? I feel like such a thing might be worth a helper for that, so, if it does need that, wait for me to add a helper for doing such a thing maybe unless you wanted to try to tackle that yourself?

---

So, I approve of the work you have done on this so far, but unfortunately, recent news is saying that this might not be all that is needed, due to some modifications nixpkgs made to some of the shell packages, and the way they set up `/etc/profile`

On the bright side, all the work you have done so far is still applicable, and is a big help and should still work in scenarios where you do things such as, launch it from your wezterm drv as the launch command

Maybe as a stop-gap measure we can `config.passthru.ZDOTDIR = "${config.wrapper}/${zdotFilesDirname}";` until we find a nice way to handle that? (honestly we should probably do that anyway?)

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,262 @@
...
+        EOL
+        cat >$out/${zdotFilesDirname}/.zlogout <<'EOL'
+        ${baseZlogout}
+        EOL
+
+        runHook postBuild
+      '';
+
+      env.ZDOTDIR = "${placeholder "out"}/${zdotFilesDirname}";
+
```
BirdeeHub, line 256:
```suggestion

      passthru.ZDOTDIR = "${config.wrapper}/${zdotFilesDirname}";

```

^ One of the things mentioned by this comment https://github.com/BirdeeHub/nix-wrapper-modules/pull/285#issuecomment-3938949902 (we use `config.wrapper` because `placeholder "out"` is a placeholder and only points where you would expect if placed into that drv build, whereas for internal consumption we use `placeholder "out"` to avoid infinite recursion)

FluxZA:
Thanks for the feedback, I had a look at that discussion, and your idea for a helper function seems like it might be the best option. I asked zenoli to test the -d flag for information sake, but it feels more like a workaround than a solution, especially if we are considering all shells.

In the same vein of ideas that work for zsh but not all shells, you can change the build args for zsh to specify where /etc files are read from. This means we could avoid sourcing any system rc files, and I imagine this would include /etc/profile. However, as you mentioned when I asked about this previously, having to rebuild the package is not ideal. I just thought I would throw it out there so all ideas are on the table.

If we decide to use a helper function, I would be happy to give it a try.

BirdeeHub:
Yeah I don't want to incur a rebuild in this repo if possible (makes tests slow, the only reason .override its ok with me in mpv is because they have mpv-unwrapped which pkgs.mpv just uses, and it works similarly with vim). Feel free to do it via the overrides option in your config, but I don't want to do it here if possible.

Im working on some kind of helper at the moment.

Im going to do the sort and return the sorted DAL with an added type field (which option it came from), esc-fn will not be applied yet, it will be in the set for that item, with the default value from config.escapingFunction substituted if it was null. It will not process argv0 and argv0type

And then Im going to figure out what to do from there, maybe that is enough maybe it is not we will see.

But likely you would import the makeWrapper module but with the wrapperImplementation option excluded, and then use that helper to define wrapperFunction such that it just returns a string which is a build script which constructs an initial wrapper via pkgs.makeBinaryWrapper.

It would filter by the type field for `flags` (normalized to be like `addFlag`), `addFlag`, and `appendFlag` from the DAL and then along with your ZDOTDIR in `placeholder "out"`, make the base wrapper using `pkgs.makeBinaryWrapper`

And then in your ZDOTDIR you would filter out the ones you already processed from the DAL, and then stick the instructions for setting the rest of the things in a config file like zshrc or zshenv (the 4 shell functions [here](https://github.com/BirdeeHub/nix-wrapper-modules/blob/8ba8703b62c2cc7e642f58ff03bd14a2f836e07a/modules/makeWrapper/makeWrapperNix.nix#L143-L146) should be sourceable by zsh and may come in handy for that)

So, I will add some amount of helper function(s). You can view it/them and decide what you want to do.

If you would rather not deal with that part, I am ok with accepting it and then adding that part to it myself and adding myself as a secondary maintainer to the module.

If I end up doing it, I will still make the part actually in the zsh module as simple as possible, so that you hopefully will actually be able to understand what I did, I don't want to crash your current understanding of how it works.

I obviously had no such concerns for the neovim one, as there is a [134 line pipe with the result of another pipe ++'d into the middle of it](https://github.com/BirdeeHub/nix-wrapper-modules/blob/8ba8703b62c2cc7e642f58ff03bd14a2f836e07a/wrapperModules/n/neovim/normalize.nix#L12-L146) in that one :rofl: So, yeah, if I do it, I pledge not to do that to your module :rofl:

FluxZA:
I am happy to implement it, I just need to take a look at what is happening in /etc/profile to try fully understand why we are doing what we are doing. I get that path is being redefined, affecting additional packages, but not how it relates to the other config options.

I also need to try get a proper understanding of the DAL, so doing this myself will be a good way to check how well I am doing there 😂.

The final added benefit is that you won't have to worry about creating another neovim situation in someone else's module XD

BirdeeHub:
Luckily, a DAG or DAL sounds more complex than it is to use

It is an `attrsOf` or `listOf` a `wlib.types.spec` type

`wlib.types.spec` is an auto-normalizing type. `wlib.types.spec` is the part that is hard to understand, you don't need to understand the implementation to use it, but you do need to sorta understand what it does.

```nix
{ config, lib, wlib, ... }: {
  options.myspecset = lib.mkOption {
    default = {};
    type = lib.types.attrsOf (wlib.types.spec {
      # yes this is a module, yes this supports type merging
      # If the thing does not have the expected fields, it will place it in the main field.
      options = {
        data = lib.mkOption {
          type = lib.types.str;
          description = "no default makes this the main one";
        };
        extrafield = lib.mkOption {
          type = lib.types.nullOr lib.types.str;
          description = "and then you can have other ones, but they need defaults";
          # You could theoretically set this default via config for this spec submodule as well.
          # It needs to be defined somewhere within the base set of modules passed to the type
          default = null;
        };
      };
    });
  };
  # provide just the value for the main field
  config.myspecset.TESTVAL1 = "hello";
  # or a set where you can set more than just the main field
  config.myspecset.TESTVAL2.data = "hello2";
  config.myspecset.TESTVAL2.extrafield = "hello2";
  config.myspecset.TESTVAL3 = { data = "hello3"; extrafield = "hello3"; };
  config.myspecset.TESTVAL4 = { data = "hello4"; };
  # now:
  # config.myspecset == {
  #   TESTVAL1 = { data = "hello"; extrafield = null; };
  #   TESTVAL2 = { data = "hello2"; extrafield = "hello2"; };
  #   TESTVAL3 = { data = "hello3"; extrafield = "hello3"; };
  #   TESTVAL4 = { data = "hello4"; extrafield = null; };
  # }
}
```

The module passed to `wlib.types.spec` for the DAG and DAL types contains `name` `before` `after` and `data` with `data` as the main field. The main field is where, if you provided just the value instead of a set, the value will be placed. So, in this case, bare values will be placed in `data`. The makeWrapper options also add an `esc-fn` field to the spec type for the DAL, and the flags set also adds a `sep` field.

The sorting functions from `wlib.dag` can sort either form of the data structure. The type is auto-normalizing, so you can just call it on the result!

The thing that makes it a DAG are the `name` `before` and `after` fields, and `data` is a convention relied upon for a few of the functions in `wlib.dag`. For example, `wlib.dag.lmap` and `wlib.dag.gmap` expect a `.data` field. But `wlib.dag.topoSort` itself doesn't expect much of anything other than that the thing is a list or set of sets with _optional_ `name` `before` and `after` fields.

As to the behavior of the sort, it gets to be [quite simple](https://github.com/BirdeeHub/nix-wrapper-modules/blob/0ea455457dac3de9129498f5c50a35153ab56484/lib/dag.nix#L280-L291), thanks to nixpkgs offering `lib.toposort`

Basically, there was a lot of stuff where you would want some collection of either the value, or a set with the value and some other stuff. And then you would need to do a bunch of normalization yourself. This is the type for that, and it automatically normalizes it. Then you do whatever with that. If it has a name before and after field, you can call the sorting function on it.

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,262 @@
...
+    {
+      package = lib.mkDefault pkgs.zsh;
+
+      # Required for shebangs to work on MacOS
+      wrapperImplementation = "binary";
+
+      # Allow use as a system/user shell
+      passthru.shellPath = "/bin/zsh";
+
+      addFlag = lib.mkIf (config.skipGlobalRC) [ "-d" ];
```
BirdeeHub, line 233:
```suggestion
      flags."-d" = config.skipGlobalRC;
```

^ lets people `config.flags."-d".data = lib.mkForce idk;`, lets people grab and use the value from `config.flags."-d".data`, and lets people `before = [ "-d" ];` or `after = [ "-d" ];` (you can do the before and after thing with `addFlag` too but you would need to `{ name = "-d"; data = [ "-d" ]; }` and they would only be able to see or override that easily via your `config.skipGlobalRc` option)

Tbh, this one is up to you now that I think about it, it is a boolean option, it kinda doesn't matter the effect is more or less the same.

Honestly maybe we remove it? Once we are setting everything other than the args and ZDOTDIR from within the config itself I don't know what conflicts could even happen, and people could always pass it themselves still

If nothing else, hopefully that further explains the DAG vs DAL types.

And also the trade-off between which to use. DAGs have better overridability because you can always index into them without having to do any mapping, but DALs have an inherent ordering and lib.mkBefore and lib.mkAfter are concepts which also make sense. The same trade-off is true of just a regular attrsOf and listOf type, except the DAG and DAL types also lets you describe dependent ordering.

BirdeeHub:
To make a zsh-sourceable version of these:

```nix
  setvarfunc = /* bash */ ''wrapperSetEnv() { export "$1=$2"; }'';
  setvardefaultfunc = /* bash */ ''wrapperSetEnvDefault() { [ -z "''${!1+x}" ] && export "$1=$2"; }'';
  prefixvarfunc = /* bash */ ''wrapperPrefixEnv() { export "$1=''${!1:+$3$2}''${!1:-$3}"; }'';
  suffixvarfunc = /* bash */ ''wrapperSuffixEnv() { export "$1=''${!1:+''${!1}$2}$3"; }'';
```

You can edit these by just swapping the `${!` for `${(P)`

```nix
prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { export "$1=''${(P)1:+$3$2}''${(P)1:-$3}" }'';
```

I think there may be some kind of emulate bash option that could be used instead of editing them also idk, but it is such a small change considering the rest of what has to be done.

The fish version of those is so long, none of these shorthands work. I don't know why people use fish other than for the slightly nicer vi mode tbh, that is whoever is making a fish wrapper module's problem XD

```zsh
❯ unset TESTVAR
❯ wrapperPrefixEnv() { export "$1=${(P)1:+$3$2}${(P)1:-$3}" }
❯ wrapperPrefixEnv TESTVAR : AGAIN
❯ wrapperPrefixEnv TESTVAR : HELLO
❯ echo "$TESTVAR"
HELLO:AGAIN
```

^ This is how we ensure uniform escaping rules for all the values under the hood in the `nix` wrapperFunction implementation. We make them all function arguments. If we were to try to interpolate them directly into what we put as the body of the function, we would end up with no end of escaping issues because we are passing from nix and not bash. It also makes the generated wrappers really nice and readable.

FluxZA:
Thank you so much for that explanation, it helped a lot.

To check that we are on the same page, I am waiting for you to finish the helper functions you are working on and then I will go about updating the zsh wrapper to handle the options other than those set in the base wrapper as part of the config?

BirdeeHub:
Yes, it will take a few days though, a bit of re-working to do.

Trying to figure out what to do with the flags set. I kinda want to add an ifs field to it which if `ifs = ",";` for example allows `flags."--arg" = [ "a" "b" "c" ];` to create `--arg=a,b,c` instead of when it is null where it would do `--arg=a --arg=b --arg=c` like it does currently.

If I want to do that, now is probably the time to do that, while I am adding helpers for doing the sort and reimplementing wrapperFunction yourself.

I have a helper for the sort, but im still working on the shape of it. I need one for processing the flags, and maybe one or 2 others. In order to determine what I need, I am reimplementing the current ones using the new helper.

But the sort itself will be something like this

```nix
{
  config,
  lib,
  wlib,
  sortResult ? true,
  ...
}:
let
  liftDag = type: v: {
    inherit type;
    data = v.data or null;
    before = v.before or [ ];
    after = v.after or [ ];
    name = v.name or null;
    esc-fn = v.esc-fn or null;
    ${if type == "flags" then "sep" else null} = v.sep or null;
    ${if type == "flags" then "ifs" else null} = v.ifs or null;
    value = v;
  };
  pushDownDagNames = builtins.mapAttrs (
    n: v:
    v
    // {
      name =
        if isString (v.name or null) then
          v.name
        else if isString n then
          n
        else
          null;
    }
  );
  mapAndLiftDal = from_option: map (liftDag from_option);
  mapAndLiftDag =
    from_option: dag:
    lib.mapAttrsToList (attr-name: v: liftDag from_option v // { inherit attr-name; }) (pushDownDagNames dag);
  unsorted =
    mapAndLiftDal "unsetVar" (config.unsetVar or [ ])
    ++ mapAndLiftDag "env" (config.env or { })
    ++ mapAndLiftDag "envDefault" (config.envDefault or { })
    ++ mapAndLiftDal "prefixVar" (config.prefixVar or [ ])
    ++ mapAndLiftDal "suffixVar" (config.suffixVar or [ ])
    ++ mapAndLiftDal "prefixContent" (config.prefixContent or [ ])
    ++ mapAndLiftDal "suffixContent" (config.suffixContent or [ ])
    ++ mapAndLiftDal "chdir" (config.chdir or [ ])
    ++ mapAndLiftDal "runShell" (config.runShell or [ ])
    ++ mapAndLiftDag "flags" (config.flags or { })
    ++ mapAndLiftDal "addFlag" (config.addFlag or [ ])
    ++ mapAndLiftDal "appendFlag" (config.appendFlag or [ ]);
in
if sortResult then wlib.dag.unwrapSort "makeWrapper" unsorted else unsorted
```

Im not sure if I want to have flags + addFlag and appendFlag as their own outputs, or if I just want to output 1 big DAL

I will have a helper for processing the flags set, and maybe one for taking a list of 3 values and turning it into a set with `var sep env` but I don't know what shape it will be, I will find out by reimplementing the current backends using this

But the idea would be, you get the raw stuff but sorted, and a helper or 2, and then you fold to group the values as desired, likely you pull out flags addFlag and appendFlag and use those plus the argv0 options and config.package config.exePath and config.binName to build the base command which gets launched, and then you map over the rest of them in place into build commands

FluxZA:
Sounds good, I will only have time to look at it next weekend anyway.

I like the idea for the comma separated args 👍

BirdeeHub:
Im also going to address https://github.com/BirdeeHub/nix-wrapper-modules/issues/307 as I add this helper.

Edit: Which I hope to be done with by the end of today (now the 27th as of writing this edit) but it isn't the only thing I have to do today, so we will see. I have been busy this week.

I do have some helpers though, I need to figure out how to expose it and use it though.

BirdeeHub:
I now need to implement the existing wrapperFunction implementations using that, or something very close to that, and figure out how to expose that.

Then it will be ready for you to use, which basically will will entail pulling out the flags using the helper, and mapping `--add-flag` and `--append-flag` into them and calling `esc-fn` on it, grabbing argv0 and argv0type, the path to wrap and the path to output to, and passing that to `pkgs.makeBinaryWrapper`. Then you would copy what I do in `makeWrapperNix` for the rest of the things, but swap `${!` for `${(P)` and put that in a file you source from a zsh config file.

BirdeeHub:
also you now have a merge conflict but it is a really simple one, someone else snuck into the maintainers list.

BirdeeHub:
And keep in mind that all of what I showed as permalinks here have veeery slightly different options in the final version, so go to the newest when you look at the files if they haven't been merged yet

BirdeeHub:
Ok. I have merged the helper things.

Good luck!

You will need to mash the two wrapperImplementation options together, and use pkgs.makeBinaryWrapper for the args and ZDOTDIR and then the nix implementation version for the thing sourced by the zsh config.

FluxZA:
Thanks for all the info, unfortunately my weekend has filled up so I probably won't get to this until sometime this week. Is there anything I should prioritise to give feedback on this weekend?

BirdeeHub:
Take your time it will still be here when you have time for it!

If there is anything to give feedback on, its these 2 files I link in this reply, and if they look sane to you or not, but I am pretty confident that what I cooked up is at the very least not harder than it was before :rofl:

Fair warning, my helpers did not make it too much less verbose, but they do make it easier to think about and work with and read the resulting code.

They handle aggregating and sorting the options into 1 DAL, and they handle splitting up the args, the flags set, and running stuff on 1 or many variants, so, they cover the hard stuff but the verbose part is the actual backend representation of these things, and that still needs to be done by implementer.

I have no idea what purposes someone may have for redefining it, so I covered as much of the backend agnostic machinery as I could without touching the actual final representation of these things.

So, there is an aggregate function to turn an options set into a DAL and optionally sort it for you, and then there is a helper to split it into `args` and `other`, and then there is a helper that turns the flags entries into addFlag entries for you. Then all you have to do is map over everything in place and apply the esc-fn however is relevant for that option, and then concat and provide them to whatever builder format you are doing.

So, like, the `pkgs.makeWrapper` backend then just intersperses the values with their `--add-flag` or whatever arg, and then applies `esc-fn` to everything and `concatStringsSep " "`, and puts it in the build string providing all that to `pkgs.makeWrapper`. And the in-house nix one for the args it basically just applies `esc-fn` and concats them, and then for the other ones, makes a build command which copies whatever they have to do into the wrapper.

You would need to do this

https://github.com/BirdeeHub/nix-wrapper-modules/blob/7025a67a1410015bf225903333f12021fd0bdff0/lib/makeWrapper/makeWrapper.nix#L11-L60

and this

https://github.com/BirdeeHub/nix-wrapper-modules/blob/7025a67a1410015bf225903333f12021fd0bdff0/lib/makeWrapper/makeWrapper.nix#L114-L124

like the `pkgs.makeBinaryWrapper` backend (and also pass `--set ZDOTDIR yourpath` to it as well)

And you would do this

https://github.com/BirdeeHub/nix-wrapper-modules/blob/7025a67a1410015bf225903333f12021fd0bdff0/lib/makeWrapper/makeWrapperNix.nix#L9-L25

and this

https://github.com/BirdeeHub/nix-wrapper-modules/blob/7025a67a1410015bf225903333f12021fd0bdff0/lib/makeWrapper/makeWrapperNix.nix#L78-L118

like the in-house nix backend but write it to a different file which you are going to source from a zsh config (and swap `${!` for `${(P)` in the 4 helper function things like `setvarfunc` and the like so it works in zsh)

You would only do that for the main binary probably.

For the wrapper variants, you would just do it kinda like I did for the neovim one where I call the normal function:

https://github.com/BirdeeHub/nix-wrapper-modules/blob/cf2acacbbb4dc86211be108524891e316bc3e4f4/wrapperModules/n/neovim/makeWrapper/default.nix#L22

The file is created by the main one already, so if you want to source it in a variant, you still can.

---

You definitely chose a difficult one as your first wrapper module :rofl: thanks for sticking with it. 🤣

BirdeeHub:
Ok.

So, there have been some changes to how wrapperFunction is implemented. It is possible  to still do it the way I mentioned. However it is also possible to do it another way as well

I decided to spend 30 minutes or so poking at it

Here is where I got to

It does not work yet, but, if I finish it at some point I will let you know, if not it will maybe be helpful to jump off from

```nix
{
  config,
  lib,
  wlib,
  pkgs,
  ...
}:
let
  inherit (lib) types;
  # defined here so that it can double as module documentation
  descriptionPre = ''
    Wrapping zsh involves some complexity regarding the global rc files. If zsh is installed
    on the system, /etc/zshenv will be used no matter what. This should not be impactful as
    system maintainers should keep the file from causing unexpected behaviour. The remaining
    global rc files can be skipped using the `skipGlobalRC` option if they are causing conflicts
    with your local rc files.

    For details regarding the rc files see <https://zsh.sourceforge.io/Doc/Release/Files.html#Startup_002fShutdown-Files>.

    This wrapper provides two methods of defining your local rc files. You can specify a directory
    which contains the files, or you can specify each file directly. These two options can be used
    together, in which case the files from the directory (specified using the `zdotdir` option) are
    sourced before the individually specified files.
  '';
in
{
  imports = [
    wlib.modules.symlinkScript
    wlib.modules.constructFiles
    (
      (import wlib.modules.makeWrapper)
      // {
        excluded_options.wrapperFunction = true;
        excluded_options.wrapperImplementation = true;
      }
    )
  ];

  options = {

    # I decided not to repeat the same zsh wrapper for all of them
    # rather, if someone wants to use this to wrap other programs in the context of their zsh they can
    wrapperVariants = lib.mkOption {
      type = lib.types.attrsOf (
        lib.types.submoduleWith {
          modules = [
            (
              { name, ... }:
              {
                config.mirror = lib.mkOverride 0 false;
                config.package = lib.mkOverride 1400 (pkgs.${name} or pkgs.hello);
              }
            )
          ];
        }
      );
    };

    skipGlobalRC = lib.mkOption {
      type = types.bool;
      default = false;
      description = ''
        Set the option for zsh to skip loading system level rc files. The system level zshenv
        file cannot be skipped.
      '';
    };

    zshAliases = lib.mkOption {
      type = types.attrsOf (types.nullOr wlib.types.stringable);
      default = { };
      description = ''
        An attribute set that maps aliases (the top level attribute names in this option) to command
        strings or directly to build outputs. Aliases mapped to null are ignored. 

        These aliases are created before any of the rc file options are sourced, therefore,
        aliases specified in those options will override the aliases specified in this option.
      '';
      example = {
        l = null;
        ll = "ls -l";
      };
    };

    zdotdir = lib.mkOption {
      type = types.nullOr wlib.types.stringable;
      default = null;
      description = ''
        Direct or string path to a directory containing rc files. The following files will be sourced from
        this directory if they exist: `.zshenv`, `.zshrc`, `.zlogin` and `.zlogout`.
      '';
    };

    zshenv = lib.mkOption {
      type = types.nullOr (wlib.types.file pkgs);
      default = null;
      description = ''
        Specifies a file which will be sourced as part of the local `.zshenv` file.

        For details on how to specify the file, see the `file` type description in the wrapper lib.
      '';
    };
    zshrc = lib.mkOption {
      type = types.nullOr (wlib.types.file pkgs);
      default = null;
      description = ''
        Specifies a file which will be sourced as part of the local `.zshrc` file.

        For details on how to specify the file, see the `file` type description in the wrapper lib.
      '';
    };
    zlogin = lib.mkOption {
      type = types.nullOr (wlib.types.file pkgs);
      default = null;
      description = ''
        Specifies a file which will be sourced as part of the local `.zlogin` file.

        For details on how to specify the file, see the `file` type description in the wrapper lib.
      '';
    };
    zlogout = lib.mkOption {
      type = types.nullOr (wlib.types.file pkgs);
      default = null;
      description = ''
        Specifies a file which will be sourced as part of the local `.zlogout` file.

        For details on how to specify the file, see the `file` type description in the wrapper lib.
      '';
    };
  };

  config =
    let
      # Name the dirctory which is created as the zdotdir in the wrapper output
      zdotFilesDirname = "wrapped_zdot";

      zshAliases = builtins.concatStringsSep "\n" (
        lib.mapAttrsToList (k: v: "alias -- ${lib.escapeShellArg "${k}=${v}"}") (
          lib.filterAttrs (k: v: v != null) config.zshAliases
        )
      );

      baseZshenv = /* bash */ ''
        # zsh-wrapped zshenv: DO NOT EDIT -- this file has been generated automatically.
        # This file is read for all shells.

        # Ensure this is only run once per shell
        if [[ -v __WRAPPED_ZSHENV_SOURCED ]]; then return; fi
        __WRAPPED_ZSHENV_SOURCED=1

        # Cover some of the work done by zsh NixOS program if it is not installed
        if [[ ! (-v __ETC_ZSHENV_SOURCED) ]]
        then
          HELPDIR="${pkgs.zsh}/share/zsh/$ZSH_VERSION/help"

          # Tell zsh how to find installed completions.
          for p in ''${(z)NIX_PROFILES}; do
              fpath=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions $p/share/zsh/vendor-completions $fpath)
          done
        fi

        # Get zshenv from wrapped options if they exist
        # zdotdir files must be sourced first to maintain documented override rules
        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
          if [[ -f "${config.zdotdir}/.zshenv" ]]
          then
            source "${config.zdotdir}/.zshenv"
          fi
        ''}
        ${lib.optionalString (config.zshenv != null) /* bash */ ''
          if [[ -f "${config.zshenv.path}" ]]
          then
            source "${config.zshenv.path}"
          fi
        ''}
      '';

      baseZshrc = /* bash */ ''
        # zsh-wrapped zshrc: DO NOT EDIT -- this file has been generated automatically.
        # This file is read for all shells.

        # Ensure this is only run once per shell
        if [[ -v __WRAPPED_ZSHRC_SOURCED ]]; then return; fi
        __WRAPPED_ZSHRC_SOURCED=1

        # zsh-wrapped defined aliases
        ${zshAliases}

        # Get zshrc from wrapped options if they exist
        # zdotdir files must be sourced first to maintain documented override rules
        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
          if [[ -f "${config.zdotdir}/.zshrc" ]]
          then
            source "${config.zdotdir}/.zshrc"
          fi
        ''}
        ${lib.optionalString (config.zshrc != null) /* bash */ ''
          if [[ -f "${config.zshrc.path}" ]]
          then
            source "${config.zshrc.path}"
          fi
        ''}
      '';

      baseZlogin = /* bash */ ''
        # zsh-wrapped zlogin: DO NOT EDIT -- this file has been generated automatically.
        # This file is read for all shells.

        # Ensure this is only run once per shell
        if [[ -v __WRAPPED_ZLOGIN_SOURCED ]]; then return; fi
        __WRAPPED_ZLOGIN_SOURCED=1

        # Get zlogin from wrapped options if they exist
        # zdotdir files must be sourced first to maintain documented override rules
        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
          if [[ -f "${config.zdotdir}/.zlogin" ]]
          then
            source "${config.zdotdir}/.zlogin"
          fi
        ''}
        ${lib.optionalString (config.zlogin != null) /* bash */ ''
          if [[ -f "${config.zlogin.path}" ]]
          then
            source "${config.zlogin.path}"
          fi
        ''}
      '';

      baseZlogout = /* bash */ ''
        # zsh-wrapped zlogout: DO NOT EDIT -- this file has been generated automatically.
        # This file is read for all shells.

        # Ensure this is only run once per shell
        if [[ -v __WRAPPED_ZLOGOUT_SOURCED ]]; then return; fi
        __WRAPPED_ZLOGOUT_SOURCED=1

        # Get zlogout from wrapped options if they exist
        # zdotdir files must be sourced first to maintain documented override rules
        ${lib.optionalString (config.zdotdir != null) /* bash */ ''
          if [[ -f "${config.zdotdir}/.zlogout" ]]
          then
            source "${config.zdotdir}/.zlogout"
          fi
        ''}
        ${lib.optionalString (config.zlogout != null) /* bash */ ''
          if [[ -f "${config.zlogout.path}" ]]
          then
            source "${config.zlogout.path}"
          fi
        ''}
      '';

      # TODO: makeWrapper here for things other than args and ZDOTDIR
      # do it like makeWrapperNix except swap ${! for ${(P)
      # wrap either to config.constructFiles.generated_zshenv.path or "$2"
      wrapperInit = "";
      # TODO: make the entrypoint wrapper that sets args and ZDOTDIR
      # use pkgs.makeBinaryWrapper.
      # wrap from wrapperPaths.input to wrapperPaths.placeholder
      wrapperEntry = "";
    in
    {
      package = lib.mkDefault pkgs.zsh;

      # Allow use as a system/user shell
      passthru.shellPath = config.wrapperPaths.relPath;

      addFlag = lib.mkIf (config.skipGlobalRC) [ "-d" ];

      constructFiles = {
        generated_zshenv = {
          relPath = "${zdotFilesDirname}/.zshenv";
          content = baseZshenv;
          builder = ''
            mkdir -p "$(dirname "$2")"
            ${wrapperInit}
            cat "$1" >> "$2"
          '';
        };
        generated_zshrc = {
          relPath = "${zdotFilesDirname}/.zshrc";
          content = baseZshrc;
        };
        generated_zlogin = {
          relPath = "${zdotFilesDirname}/.zlogin";
          content = baseZlogin;
        };
        generated_zlogout = {
          relPath = "${zdotFilesDirname}/.zlogout";
          content = baseZlogout;
        };
      };

      buildCommand.makeWrapper =
        wrapperEntry + "\n" + wlib.makeWrapper.wrapVariants { inherit config pkgs; };

      env.ZDOTDIR = "${placeholder config.outputName}/${zdotFilesDirname}";

      meta.maintainers = [ wlib.maintainers.fluxza ];
      meta.description = {
        pre = descriptionPre;
      };
    };
}
```

BirdeeHub:
@FluxZA I finished it. I broke it up into several files too.

Can/should I list both of us as maintainer of it? Is that OK with you? Do you want that? You worked on it too, I just moved it around and added the last bit.

https://github.com/BirdeeHub/nix-wrapper-modules/blob/9dd11689016f0065792f4da6eca73933a229843a/wrapperModules/z/zsh

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/347

Closing this PR because the zsh module has now merged

PR #299, opened 2026-02-15

BirdeeHub:


BirdeeHub:
whoa the tests have been running 25 minutes something is broken in nixpkgs I guess (it appears deno is hanging forever and nothing I added runs deno to build, so it is an issue with one of the packages from nixpkgs being used by one of the modules)

BirdeeHub:
looks like I finally get them to both be using nixpkgs-unstable again

PR #302, opened 2026-02-17
WIP: feat(wrapperModules.xdg): init

ameertaweel:
Following the discussion in https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295.

It should work with any shell, not just Bash.

I still have to write tests for it, but I would appreciate comments about this current form.

### DIFF
--- a/wrapperModules/x/xdg/module.nix
+++ b/wrapperModules/x/xdg/module.nix
@@ -1,0 +1,248 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+let
+  types = lib.types;
+  pathAsStr = types.coercedTo types.path builtins.toString types.str;
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    baseDirs.enable = lib.mkEnableOption "management of XDG base directories";
+
+    baseDirs.cacheHome = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/.cache";
+      description = ''
+        Absolute path to directory holding application caches.
+
+        Sets `XDG_CACHE_HOME` for the user if `xdg.enable` is set `true`.
+      '';
+    };
+
+    baseDirs.configHome = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/.config";
+      description = ''
+        Absolute path to directory holding application configurations.
+
+        Sets `XDG_CONFIG_HOME` for the user if `xdg.enable` is set `true`.
+      '';
+    };
+
+    baseDirs.dataHome = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/.local/share";
+      description = ''
+        Absolute path to directory holding application data.
+
+        Sets `XDG_DATA_HOME` for the user if `xdg.enable` is set `true`.
+      '';
+    };
+
+    baseDirs.stateHome = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/.local/state";
+      description = ''
+        Absolute path to directory holding application states.
+
+        Sets `XDG_STATE_HOME` for the user if `xdg.enable` is set `true`.
+      '';
+    };
+
+    userDirs.enable = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to manage {file}`$XDG_CONFIG_HOME/user-dirs.dirs`.
+
+        The generated file is read-only.
+      '';
+    };
+
+    userDirs.package = lib.mkPackageOption pkgs "xdg-user-dirs" { nullable = true; };
+
+    # Well-known directory list from
+    # https://gitlab.freedesktop.org/xdg/xdg-user-dirs/blob/master/man/user-dirs.dirs.xml
+
+    userDirs.desktop = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Desktop";
+      description = "The Desktop directory.";
+    };
+
+    userDirs.documents = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Documents";
+      description = "The Documents directory.";
+    };
+
+    userDirs.download = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Downloads";
+      description = "The Downloads directory.";
+    };
+
+    userDirs.music = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Music";
+      description = "The Music directory.";
+    };
+
+    userDirs.pictures = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Pictures";
+      description = "The Pictures directory.";
+    };
+
+    userDirs.publicShare = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Public";
+      description = "The Public share directory.";
+    };
+
+    userDirs.templates = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Templates";
+      description = "The Templates directory.";
+    };
+
+    userDirs.videos = lib.mkOption {
+      type = types.nullOr pathAsStr;
+      default = "$HOME/Videos";
+      description = "The Videos directory.";
+    };
+
+    userDirs.extraConfig = lib.mkOption {
+      type = types.attrsOf (pathAsStr);
+      default = { };
+      defaultText = lib.literalExpression "{ }";
+      example = lib.literalExpression ''
+        {
+          MISC = "$HOME/Misc";
+        }
+      '';
+      description = ''
+        Other user directories.
+
+        The key ‘MISC’ corresponds to the user-dirs entry ‘XDG_MISC_DIR’.
+      '';
+    };
+
+    userDirs.createDirectories = lib.mkEnableOption "automatic creation of the XDG user directories";
+
+    userDirs.setSessionVariables = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to set the XDG user dir environment variables, like
+        `XDG_DESKTOP_DIR`.
+
+        ::: {.note}
+        The recommended way to get these values is via the `xdg-user-dir`
+        command or by processing `$XDG_CONFIG_HOME/user-dirs.dirs` directly in
+        your application.
+        :::
+      '';
+    };
+  };
+
+  config =
+    let
+      baseDirs = (
+        lib.filterAttrs (n: v: !isNull v) {
+          XDG_CACHE_HOME = config.baseDirs.cacheHome;
+          XDG_CONFIG_HOME = config.baseDirs.configHome;
+          XDG_DATA_HOME = config.baseDirs.dataHome;
+          XDG_STATE_HOME = config.baseDirs.stateHome;
+        }
+      );
+
+      userDirs =
+        (lib.filterAttrs (n: v: !isNull v) {
+          DESKTOP = config.userDirs.desktop;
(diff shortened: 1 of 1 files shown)


BirdeeHub:
That is an interesting idea actually

So, you take a wrapped shell, and you supply it here before passing it to `users.users.<name>.shell` or something?

But theoretically you could do it with anything else that runs at startup?

Personally I might have waited for services for this? But actually maybe it is better as a script, it would be pretty easy to turn this into a service later and now it works without a service runner.

ameertaweel:
I've been porting my NixOS config to remove `home-manager` entirely. One thing that bit me is XDG directories setup as NixOS itself doesn't support configuring per-user stuff. Without `home-manager`, I can't set per-user env vars, and I can't create per-user `systemd` services. I arrived at this solution after a looot of thinking. Simplicity is the ultimate goal of is.

BirdeeHub:
> NixOS itself doesn't support configuring per-user

Yeah it is a bit wild. It has `systemd.user` stuff, but it is not parameterized by user it applies to all of them.

It would be super cool if they had a `systemd.packages` but for per-user. Home manager has one! `systemd.user.packages`! But, that would require them to provision files to the user directory I think and they do try to avoid that? If there is a way for it to not require that, then they have no reason not to have one haha

~Also, why do they not have a `systemd.user.packages` for nixos? It would apply to all of them just like the `systemd.user.services` options they have, so it would not exactly be per user still, but why is it missing? Did I miss it?~ Edit: `systemd.packages` does both I think actually.

BirdeeHub:
https://github.com/NixOS/nixpkgs/blob/fa56d7d6de78f5a7f997b0ea2bc6efd5868ad9e8/nixos/modules/system/boot/systemd/user.nix#L256C1-L257C58

HMMMMMM

This is an interesting comment....

Wonder if they could use that to add such a `systemd.users.<name>.packages` option...

Im still happy to do it as a wrapper module with a script, and maybe later we can have an option to make it generate a service which runs that script instead of wrapping the thing again. But until then, having it wrap the thing again works and is fine.

BirdeeHub:
Suggestion:

Wrap `pkgs.xdg-user-dirs` instead.

Then run it in your bashrc.

Then this would be `wrapperModules.xdg-user-dirs` and it would actually be wrapping what it says.

It was always silly to me that `xdg-user-dirs` doesn't give any way to actually create the directories. We could do that. Might be cleaner than double wrapping your shell.

Then, if you wanted to accept the same options in your bash wrapper module, you could

```nix
options.xdg-user-dirs = lib.mkOption {
  type = wlib.types.subWrapperModule wlib.wrapperModules.xdg-user-dirs;
  description = ''
    settings for making xdg user dirs on first login
  '';
};
# and then you can run the stuff from `config.xdg-user-dirs.wrapper` in your bashrc
```

wrapperModules/x/xdg/module.nix
```diff
@@ -0,0 +1,248 @@
...
+              (lib.optionalString config.baseDirs.enable createBaseDirs)
+              (lib.optionalString (config.userDirs.enable && config.userDirs.createDirectories) createUserDirs)
+              (lib.optionalString config.userDirs.enable createUserDirsConf)
+              (lib.optionalString config.userDirs.enable createUserDirsDirs)
+            ];
+          }
+        ];
+
+      # NOTE: Any shell works here
+      package = lib.mkDefault pkgs.bashInteractive;
```
BirdeeHub, line 244: Idea:

Wrap `pkgs.xdg-user-dirs` instead.

Then run it in your bashrc.

Then this would be `wrapperModules.xdg-user-dirs` and it would actually be wrapping what it says.

It was always silly to me that `xdg-user-dirs` doesn't give any way to actually create the directories. We could do that. Might be cleaner than double wrapping your shell.

Then, if you wanted to accept the same options in your bash wrapper module, you could

```nix
options.xdg-user-dirs = lib.mkOption {
  type = wlib.types.subWrapperModule [ wlib.wrapperModules.xdg-user-dirs { inherit pkgs } ];
  default = { };
  description = ''
    settings for making xdg user dirs on first login
  '';
};
# and then you can run the stuff from `config.xdg-user-dirs.wrapper` in your bashrc
```

If you didn't want to you don't even actually have to wrap the `xdg-user-dirs` or `xdg-user-dirs-update` binaries themselves to do that you could just stick an extra script in there via a drv hook and leave the user to wrap the actual binaries however they want. You might also want to wrap those, or achieve the functionality BY wrapping one or both of those instead, idk up to you, whatever feels best.

This would still work with any shell, and it keeps it within the system, and has a clearer separation of concerns IMO. The current iteration wants to be a helper module, but it also wants to set the package option. Wrapping `pkgs.xdg-user-dirs` is probably better.

The options you have seem fine at a glance, haven't tried it yet. Ill look at those more closely when it is more ready. But it seems to be wrapping the wrong thing.

When I first looked at it, it was not immediately obvious what was strange, but I think this identifies both what that was and a good solution to it.

In addition, this will seamlessly fit in with when we have service options, this wrapper module can then offer those services for those who want to pass it to an option like `systemd.packages` without it feeling weird to be passing a `bash` drv to it for that. You would instead be passing your `xdg-user-dirs` drv to install the services for it. Feels better to me.

ameertaweel: Just to make sure I understand. What your saying is:

- Add an extra script to `xdg-user-dirs` via a drv hook.
- The script will support creating the XDG directories and exporting the environment variables.
- Call this script in my `.bashrc`.
- Wrap Bash to use my `.bashrc` via the `--rcfile` Bash argument.
- Now we have a `wrapperModules.xdg-user-dirs` that's independent of any specific shell.

Is my understanding correct?

BirdeeHub: Yes, basically, although the sequencing is not the way I would say it. The order I would have listed it in would be:

- Add an extra script to xdg-user-dirs via a drv hook.
- The script will support creating the XDG directories. (If you want to also wrap the binaries pkgs.xdg-user-dirs provides and run the wrapped binaries as part of this script, go for it, but the script just needs to make the directories.)
- Now we have a wrapperModules.xdg-user-dirs that's independent of any specific shell and contains a script to create the dirs.

You now need to run that script sometime at runtime.

One method for that:

- Call this script in my .bashrc.
- Wrap Bash to use my .bashrc via the --rcfile Bash argument.

And because `wrapperModules.xdg-user-dirs` is a wrapper module, you can add a `subWrapperModule` option to your bash wrapper module that just uses that and then runs the script it provides, so that it feels like you are just setting the options directly within your bash wrapper module (if you want)

Another method for that:

Call that script from a service. Which, eventually, when we have services options, will be as simple as passing the wrapped xdg-user-dirs derivation to systemd.packages, but currently would mean defining a service yourself which runs the script from the wrapped xdg-user-dirs derivation

zenoli: My 2 cents:

- Having a dedicated module that can be integrated in whatever shell you want using `subWrapperModule` is the best solution imo.
- From what I've read, `xdg-user-dirs` is for managing user dirs:
```sh
XDG_DESKTOP_DIR="$HOME/Desktop"
XDG_DOCUMENTS_DIR="$HOME/Documents"
XDG_DOWNLOAD_DIR="$HOME/Downloads"
XDG_MUSIC_DIR="$HOME/Music"
XDG_PICTURES_DIR="$HOME/Pictures"
XDG_PUBLICSHARE_DIR="$HOME/Public"
XDG_TEMPLATES_DIR="$HOME/Templates"
XDG_VIDEOS_DIR="$HOME/Videos"
```

but not the `CONFIG/DATA/STATE/CACHE` variants. So we might be mixing concepts if we extend the functionality  there.

But do we even need `xdg` as a standalone wrapper?
I think the cleanest solution is just to provide a reusable subWrapperModule that can be integrated in all shell wrappers to come. This way, we provide a standardized way of integrating these options, and they can be maintained centrally. 

@BirdeeHub couldn't we provide this as an additional [helper module](https://birdeehub.github.io/nix-wrapper-modules/md/helper-modules.html)? Or would this the wrong place?

BirdeeHub: If you were to provide it as an additional helper module, you would just use a regular `submodule` type and enable some random stdenv phase nobody uses and then build it.

```nix
{ config, wlib, lib, ... }: {
  options.myopts = lib.mkOption {
    # you can use a submodule type, or just like, a regular set of options here.
    type = lib.types.submodule { options = { ... }; };
    default = {};
  };
  config.drv.doDist = true;
  config.drv.distPhase = ''
    runHook preDist
    build your script which can be pointed at from wherever in the wrapper module with ${passthru "out"};
    runHook postDist
  '';
}
```

Then people would import said module if they wanted to add such a script to their derivation to use.

I would not rely solely on `runShell` to do it for a generalized helper module, as it does not work with `wrapperImplementation = "binary"`, it should include the script for people to do what they want with it.

---

Personally I quite like doing it under `xdg-user-dirs` though compared to those options.

it is still a user dir. It is in the user directory. There is no tool that exists under the xdg-utils or xdg-user-dirs umbrella for creating the CONFIG/DATA/WHATEVER directories. There is not even a tool for setting the variables as far as I am aware. I don't feel like this is conflating concepts to include a script for creating such things in a wrapped xdg-user-dirs derivation. To be honest, Im not even sure what else I would want a wrapped xdg-user-dirs derivation to even do. It will also allow us to effortlessly use services options when we add them, which is something you would want standalone.

Personally, I am in favor of wrapping xdg-user-dirs and then using it as a `subWrapperModule` in other modules if desired. (`subWrapperModule` is just a [wrapped `submodule` type](https://github.com/BirdeeHub/nix-wrapper-modules/blob/0ea455457dac3de9129498f5c50a35153ab56484/lib/types.nix#L37-L55))

Doing it as a wrapper for xdg-user-dirs even lets us run xdg-user-dirs-update and stuff as part of the steps (or have it run the thing when you run xdg-user-dirs-update, probably a better dependency order), so that it works fully as one might expect. And then we can throw in also creating `${XDG_CONFIG_HOME:-~/.config}` and the like as a bonus if we want

If it does provide services at one point, having it as a standalone thing and not a helper module will be much better. Because you would then want to have it standalone to pass as a standalone service.

I think it fits best as a wrapper for xdg-user-dirs for those reasons.

ameertaweel: Sorry for the late reply. I didn't get the chance to work on this for a while. I started making the changes to make this a wrapper for `xdg-user-dirs`. But then I noticed an issue:

I agree that a wrapper for `xdg-user-dirs` that provides a script for creating the directories is nice. But where does the logic for creating the environment variables go? In the current form of this PR, setting the `env` option in the Bash wrapper sets the environment variables for Bash, which makes sense. If I set `env` in the `xdg-user-dirs` wrapper, and use that in my `.bashrc`, Bash won't see the environment variables.

Of course I can split the logic for creating directories and setting environment variables into different wrappers but it feels unclean to me.

Do you have any suggestions on how to go about this?

BirdeeHub: Hmmmm

Well

4 possible ways are coming to mind, idk which is best or what to do.

1, we could have the wrapper generate a sourceable bash script and have the main wrapper source that instead of just a single file, then other things could source that. Or just have both.

2, we could have the wrapper export a helper module via readOnly option that sets env in the module which imports it.

3, programs where one is likely to set these, such as display managers and shells, could take an option somewhat like https://github.com/BirdeeHub/nix-wrapper-modules/discussions/297#discussioncomment-16146073 where it grabs the values from it? Not sure about this one.

4. Actually, so now that we have the core `buildCommand` option, it is a little easier to just run stuff, maybe it could be a helper module that sets env and then adds a script to create xdg dirs? And then later it could also set some service options? Or, you could have it both wrap xdg-user-dirs, but also be a functor like the makeWrapper module is that allows you to set "helper module mode" for it, then it could be both those things.

Also when services and this https://github.com/BirdeeHub/nix-wrapper-modules/discussions/370 are things, more options might appear too.

PR #310, opened 2026-02-28
type(wrapperModules.cava): init

rachitvrma:
Add wrapper for [Cava: Audio Visualizer](https://github.com/karlstav/cava).

### DIFF
--- a/wrapperModules/c/cava/module.nix
+++ b/wrapperModules/c/cava/module.nix
@@ -27,6 +27,6 @@ in
       "-p" = configFile;
     };
     meta.maintainers = [ wlib.maintainers.rachitvrma ];
-    meta.platforms = [ "x86_64-linux" ];
+    meta.platforms = lib.platforms.linux;
   };
 }


ci/docs/redirects.nix
```diff
@@ -6,6 +6,7 @@
         "alacritty"
         "atool"
         "btop"
+        "cava"
```
BirdeeHub: this is not necessary, redirects are only needed for things that existed before the page locations were changed

I moved all of the pages once. So, I added redirects as to not break old links to the site.

Good attention to detail, but not something you should add.

wrapperModules/c/cava/check.nix
```diff
@@ -0,0 +1,10 @@
+{ pkgs, self, ... }:
+let
+  cavaWrapped = self.wrappers.cava.wrapped {
```
BirdeeHub:
```suggestion
  cavaWrapped = self.wrappers.cava.wrap {
```

(this is what is making the test fail at the moment)

wrapperModules/c/cava/check.nix
```diff
@@ -0,0 +1,10 @@
...
+    inherit pkgs;
+  };
+in
+pkgs.runCommand "cava-test" { } ''
+  "${cavaWrapped}/bin/cava -v"
```
BirdeeHub: does it give any useful printout information about the config we provided we could check too?

rachitvrma: No nothing like that.

wrapperModules/c/cava/check.nix
rachitvrma: After making the changes suggested. 

```
error: The option `pkgs' was accessed but has no value defined. Try setting the option.
```

BirdeeHub: Ah. I was not paying attention

I meant to suggest `.wrap`

My bad. I was doing several things at once when I made the suggestion.

You changed it to `.apply { inherit pkgs; };` instead of what I meant to suggest which was `.wrap { inherit pkgs; };`.

That also works, but it does mean that you still need to `.wrapper` if you use `.apply` instead of `.wrap`. It runs, so all good, but it is slightly cleaner to use `.wrap` here.

rachitvrma: Yep. This is my first contribution to something nix related, and my first contribution to open source at all.

BirdeeHub:
It appears that cava does not build on mac? Is that expected? You may need to exclude the test for them [like this](https://github.com/BirdeeHub/nix-wrapper-modules/blob/0ea455457dac3de9129498f5c50a35153ab56484/wrapperModules/m/mako/check.nix#L14) and set `meta.platforms = lib.platforms.linux;` (which doesn't do anything on its own necessarily, but may show up in the docs eventually and should be used by the test to disable itself like shown at that link)

wrapperModules/c/cava/check.nix
BirdeeHub: Ah, sorry you got my distracted suggestion then XD Thanks for stopping by to help out!

I was working on upgrading the makeWrapper stuff to be easy to redefine onesself if required, so I had a lot of things to think about and did not give your PR the proper amount of attention.

rachitvrma:
Checks pass when `meta.platforms = [ "x86_64-linux" ]`

BirdeeHub:
Can you use `lib.platforms.linux` instead? It is a list of all the linux platforms.

rachitvrma:
> Can you use `lib.platforms.linux` instead? It is a list of all the linux platforms.

Thanks for the guidance.

BirdeeHub:
A quick look at the package says it only has 2 flags, one of them being -p to point to the config file, and it looks like it does it all through there. And it doesn't offer any services either.

So, this should be sufficient! Looks good to me!

Issue #307, opened 2026-02-24
Wrapped executables in destinations other than `$out/bin`


iclanzan:
Many packages expose scripts that don’t necessarily reside under `./bin`. But there doesn’t seem to be a way to wrap them while placing the wrapped script in the same place as the original and as a consequence anything that refers to that script breaks.

The specific problem that I am running into is trying to package gnome applications that also come with search providers which are stored under `$out/share/<app-name>/search_provider`.

BirdeeHub:
Currently there is not an option for this in the makeWrapper module.

The easiest current workaround would be just literally move it afterwards

```nix
{ config, ... }: {
  drv.buildPhase = ''
    runHook preBuild
    mv $out/bin/${config.binName} $out/share/${config.binName}/search_provider
    runHook postBuild
  '';
}
```

If you are already wrapping the main package, then you use wrapperVariants and move that one in the hook instead

Both top level and the variants have binName (the name in bin it gets added with) and exePath (the thing you are wrapping) options.

Hopefully we can add some kind of output path option for the makeWrapper module. Im working on those right now to add some other things for redefining them, I could maybe add that while I am at it.

Problem is, if I add an outputDir option, then binName is slightly mis-named... And then you would need to make `binName = "search_provider"` and `outputDir = "share/<app-name>"` in your example for whichever set of makeWrapper options was building that one... I might just have to live with that, that is likely pretty annoying for people if I deprecate it (I could give it a warning and keep it working for a long time for people to update it, but like, still) I could instead make an outputPath option and have it be mutually exclusive with binName but that is even worse, it is nice having binName be something you can use without checking if it is null or not.

iclanzan:
Your `drv.buildPhase` idea works for now, thanks!

> And then you would need to make binName = "search_provider" and outputDir = "share/<app-name>" in your example

This sounds reasonable to me.

It would be a breaking change, but I think something like this would be cool:

```nix
wrapperVariants."share/<app-name>/search_provider" = {};
```

Where the attribute name is used as what is currently the `execPath` and it would output the wrapped binary in the same spot.

BirdeeHub:
As much as I like that idea, unfortunately, that has 2 problems.

1, I would really like it to be possible for it to read from and output to different locations.

[...]

PR #311, opened 2026-02-28
feat(wlib.makeWrapper): library for redefining wrapperFunction in wlib

BirdeeHub:
Exposed the (4) functions of the makeWrapper library via `wlib.makeWrapper`, (minus the module itself which is still where it was and can still be imported and modified, although the rest of the functions were deprecated, due to now being in wlib.makeWrapper)

also added `ifs` field to flags specs to further change how `flags.name = [ a list ]` is rendered

Added (3) functions for aggregating and sorting makeWrapper options DAL, and for processing the flags set correctly, without dictating the backend representation farther than the text of the options themselves dictate.

Added `config.binDir` for deciding what directory the `config.binName` binary ends up in, added for both top level and `wrapperVariants.${name}`, addressing https://github.com/BirdeeHub/nix-wrapper-modules/issues/307


BirdeeHub:
Ok, well, I built my config with it. It seems to work fine.

Ive looked it over.

I will look it over 1 or 2 more times and then I guess  it is ready?

:crossed_fingers: casual rewrite of like 1/3rd of the useful code

BirdeeHub:
had some doc edits

BirdeeHub:
I think these changes make it not a ton shorter for reimplementers than it was before, but they do make it much easier to think about and manage all those options

PR #312, opened 2026-02-28
type(wrapperModules.yt-dlp): init

rachitvrma:
Add yt-dlp wrapper module

### DIFF
--- a/wrapperModules/y/yt-dlp/module.nix
+++ b/wrapperModules/y/yt-dlp/module.nix
@@ -12,7 +12,6 @@ let
     oneOf [
       bool
       int
-      str
       wlib.types.stringable
     ];
 


BirdeeHub:
Since you are doing your own file rendering, lets make placeholder "out" work inside the generated file

Something like

```nix
# since the string goes directly into the drv, now placeholder "out" will work in it
config.drv.renderedSettings = lib.concatStringsSep "\n" (lib.remove "" (renderSettings config.settings));
# passAsFile so it can be as long as we want
config.drv.passAsFile = [ "renderedSettings" ];
# copy it into $out somewhere, preferably parameterized by binName
# (passAsFile makes varnamePath from varname, this is a feature of derivations.)
config.drv.buildPhase = ''
  runHook preBuild
  cp $renderedSettingsPath "$out/${config.binName}-settings.conf"
  runHook postBuild
'';
# point to it for your flag.
config.flags."--config-location" = "${placeholder "out"}/${config.binName}-settings.conf";
```

wrapperModules/y/yt-dlp/module.nix
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  config = {
+    package = pkgs.yt-dlp;
+    flags = {
+      "--config-location" = "${placeholder "out"}/${config.binName}-settings.conf";
+    };
+    drv = {
+      renderedSettings =
+        lib.concatStringsSep "\n" (lib.remove "" (renderSettings config.settings)) + "\n";
```
BirdeeHub:
```suggestion
        lib.concatStringsSep "\n" (lib.remove "" (renderSettings config.settings));
```

What is the extra newline for?

rachitvrma: I saw a lot of wrappers using it. Home-Manager used it in a lot of config generators, too, to adjust extraConfig. So, I thought it was a standard.

BirdeeHub:
It seems like there are some values which in the config file are like `-o` with 1 dash for example.

It seems to follow the normal command line logic for that, but I don't know if it always accepts both 1 and 2 dashes or not.

If it accepts `--o` then this is fine.

If not, should you check if the thing has more than 1 char, and if it is 1, do 1 dash and more, do 2? Or should you just require them to specify `config.settings."--option"`?

wrapperModules/y/yt-dlp/module.nix
BirdeeHub: Hmmm

I mean, maybe there is some edgecase with echoing to the file?

But you arent doing that you are just copying it.

And if you had an extraConfig option, you would include it in the list you are concatenating.

IDK I didnt see any reason for it to be there but Im not going to fight you if you like it lol

rachitvrma:
```nix
renderSingleOption =
    name: value:
    let
      isShort = builtins.stringLength name == 1;
      prefix = if isShort then "-" else "--";
    in
    if lib.isBool value then
      if value then
        "${prefix}${name}"
      else if isShort then
        throw "Short values cannot be negated: -${name}"
      else
        "--no-${name}"
    else
      "${prefix}${name} ${toString value}";
```


What about this implementation? However, the single dash options have no `--no-${option_short_form}` alternative.

Edit: home-manager's yt-dlp module has a warning set to not use short-form options at all. If something like that is possible here, maybe we could implement that?

BirdeeHub:
Just don't include it?

You do `lib.remove ""` that removes empty strings from the list, so you can just return an empty string as the negation?

And for the set forms, it looks like that would just mean an extra newline, you could maybe filter it there too, but an extra newline is fine idk, but for the list one, that would filter it out.

Maybe you should move that logic into renderSettings function so it is more explicit that it all goes together?

BirdeeHub:
```nix
  renderSettings = settings: builtins.concatStringsSep "\n" (lib.remove "" (lib.mapAttrsToList (
    name: value:
    if lib.isList value then
      (map (renderSingleOption name) value)
    else
      [ (renderSingleOption name value) ]
  ) settings));
  ```
  
Something like that, and then renderSingleOption can just return "" to omit, (or null and do `lib.remove null`, but you might have to then update the signature so idk up to you on that)

Edit: wait, thats not quite it. 1s

Here we go:

```nix
  renderSettings = settings: lib.pipe settings [
    (lib.mapAttrsToList (
      name: value:
      if lib.isList value then
        (map (renderSingleOption name) value)
      else
        [ (renderSingleOption name value) ]
    ))
    builtins.concatLists
    (lib.remove "")
    (builtins.concatStringsSep "\n")
  ];
```

wrapperModules/y/yt-dlp/module.nix
```diff
@@ -0,0 +1,59 @@
...
+    };
+  };
+
+  config = {
+    package = pkgs.yt-dlp;
+    flags = {
+      "--config-location" = "${placeholder "out"}/${config.binName}-settings.conf";
+    };
+    drv = {
+      renderedSettings = lib.concatStringsSep "\n" (lib.remove "" (renderSettings config.settings));
```
BirdeeHub:
```suggestion
      renderedSettings = renderSettings config.settings;
```

wrapperModules/y/yt-dlp/module.nix
```diff
@@ -0,0 +1,59 @@
...
+    ];
+
+  renderSingleOption =
+    name: value:
+    if lib.isBool value then
+      if value then "--${name}" else "--no-${name}"
+    else
+      "--${name} ${toString value}";
+
+  renderSettings = lib.mapAttrsToList (
```
BirdeeHub:
```suggestion
  renderSettings = settings: lib.pipe settings [
    (lib.mapAttrsToList (
      name: value:
      if lib.isList value then
        (map (renderSingleOption name) value)
      else
        [ (renderSingleOption name value) ]
    ))
    builtins.concatLists
    (lib.remove "")
    (builtins.concatStringsSep "\n")
  ];
```

BirdeeHub (changes requested):
This would allow us to then return "" from renderSingleOption to not include something (and be cleaner)

(and then you would do the thing with the renderSingleOption you said, but return "" instead of throwing)

wrapperModules/y/yt-dlp/module.nix
```diff
@@ -0,0 +1,75 @@
...
+    else
+      "${prefix}${name} ${toString value}";
+
+  renderSettings =
+    settings:
+    lib.pipe settings [
+      (lib.mapAttrsToList (
+        name: value:
+        if lib.isList value then
+          lib.concatStringsSep "\n" (map (renderSingleOption name) value)
```
BirdeeHub:
```suggestion
          (map (renderSingleOption name) value)
```

This will get concatenated at the end of the pipe.

We make a list of lists

We concat that list of lists into a list

We remove ""

We concat with "\n"

rachitvrma: Applied the suggested changes.

BirdeeHub: 1s I will try it out :)

wrapperModules/y/yt-dlp/module.nix
```diff
@@ -0,0 +1,75 @@
...
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+  configAtom =
+    with types;
+    oneOf [
+      bool
+      int
+      str
```
BirdeeHub:
```suggestion
      wlib.types.stringable
```

It stood up to

`:bl outputs.wrappers.yt-dlp.wrap { inherit pkgs; settings = { TESTARG = false; TESTING = [ "a" ]; o = "hihihi"; a = false; }; }`

Although that is obviously not a valid config XD

And when I tried to put null or other things in there it yelled at me via the module system, so that is also good.

However, I cannot put a raw derivation in there. But your renderSingleOption supports that.

So we just need to change the type to "and anything else that I can do `"${}"` to" instead of just lib.types.str

After that, I think this is good.

PR #314, opened 2026-03-01

BirdeeHub:

### DIFF
--- a/ci/flake.lock
+++ b/ci/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1771207753,
-        "narHash": "sha256-b9uG8yN50DRQ6A7JdZBfzq718ryYrlmGgqkRm9OOwCE=",
+        "lastModified": 1772419343,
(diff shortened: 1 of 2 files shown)


BirdeeHub:
IDK what a `jeepney` is but its a python package and it appears to be a dependency of `yt-dlp`

So, if you are using `yt-dlp` on macos, don't update nixpkgs at the moment it won't build apparently.

Nothing I can do about that here other than disable the test, or wait, or point at a different nixpkgs commit.

Im going to try it again tomorrow or something.

Your config is going to use whatever `pkgs` you give it, so, this version doesn't matter to users anyway it is just for tests and docs.

Edit: Apparently `jeepney` talks to dbus

Why is it even trying to use that on macos?? Does macos have dbus? I thought it did not?

Its probably for some keyring something-or-other on linux and they forgot to exclude it on mac or something.

Edit 2:

Waiting is the correct move.

https://github.com/NixOS/nixpkgs/commit/b53b848bce00e8f27769d9f5cee29691098a9492

Someone fixed it already. So this update will just have to wait [until that hits nixpkgs-unstable](https://nixpk.gs/pr-tracker.html?pr=493943) in the next day or 2.

PR #319, opened 2026-03-03
type(wrapperModules.aria2): init

rachitvrma:
Wrapper for the aria2 downloader.

### DIFF
--- a/wrapperModules/a/aria2/module.nix
+++ b/wrapperModules/a/aria2/module.nix
@@ -12,10 +12,6 @@ let
       formatValue = v: if builtins.isBool v then (if v then "true" else "false") else toString v;
     in
     "${n}=${formatValue v}";
-
-  configFile = pkgs.writeText "aria2Wrapped.conf" (
-    lib.concatStringsSep "\n" (lib.mapAttrsToList formatLine config.settings)
-  );
 in
 {
   imports = [ wlib.modules.default ];
@@ -31,21 +27,25 @@ in
           str
         ]);
       default = { };
-      description = ''
-        Settings to be wrapped with aria2 binary.
-        See {manpage}'aria2c(1)' 
-      '';
+      description = "Settings to be wrapped with aria2 binary";
     };
   };
   config = {
     package = pkgs.aria2;
+    binName = "aria2c";
+    outputs = (config.package.outputs or [ "out" ]) ++ [ "conf" ];
     flags = {
-      "--conf-path" = configFile;
+      "--conf-path" = "${placeholder "conf"}/${config.binName}-settings.conf";
     };
     flagSeparator = "=";
     drv = {
+      renderedSettings = lib.concatStringsSep "\n" (lib.mapAttrsToList formatLine config.settings);
+      passAsFile = [ "renderedSettings" ];
+
       buildPhase = ''
         runHook preBuild
+        mkdir -p $conf
+        cp $renderedSettingsPath "$conf/${config.binName}-settings.conf"
         rm $bin/bin/aria2c
         cp $out/bin/aria2c $bin/bin
         runHook postBuild
@@ -53,5 +53,4 @@ in
     };
     meta.maintainers = [ wlib.maintainers.rachitvrma ];
   };
-
 }


rachitvrma:
```nix
{
  pkgs,
  self,
}:
let
  aria2Wrapper = self.wrappers.aria2.wrap {
    inherit pkgs;
    settings = {
      file-allocation = "none";
      log-level = "warn";
      max-connection-per-server = 4;
      min-split-size = "5M";
      on-download-complete = "exit";
      auto-file-renaming = false;
    };
  };
in
pkgs.runCommand "aria2-test" { } ''
  "${aria2Wrapper}/bin/aria2c" -v | grep "${aria2Wrapper.version}"
''
```

Check results in:
```
error: builder for '/nix/store/8jan3da7h36gpg1jn6xkrpd9vh93kgy4-aria2-test.drv' failed to produce output path for output 'out' at '/nix/store/8jan3da7h36gpg1jn6xkrpd9vh93kgy4-aria2-test.drv.chroot/nix/store/8y4zi0g9m64gjyagliqfvvlxmlqrhbgp-aria2-test'
```

BirdeeHub:
Hey! Thanks! I can review this soon, but not tonight unfortunately.

BirdeeHub:
> ```nix
> {
>   pkgs,
>   self,
> }:
> let
>   aria2Wrapper = self.wrappers.aria2.wrap {
>     inherit pkgs;
>     settings = {
>       file-allocation = "none";
>       log-level = "warn";
>       max-connection-per-server = 4;
>       min-split-size = "5M";
>       on-download-complete = "exit";
>       auto-file-renaming = false;
>     };
>   };
> in
> pkgs.runCommand "aria2-test" { } ''
>   "${aria2Wrapper}/bin/aria2c" -v | grep "${aria2Wrapper.version}"
> ''
> ```
> 
> Check results in:
> 
> ```
> error: builder for '/nix/store/8jan3da7h36gpg1jn6xkrpd9vh93kgy4-aria2-test.drv' failed to produce output path for output 'out' at '/nix/store/8jan3da7h36gpg1jn6xkrpd9vh93kgy4-aria2-test.drv.chroot/nix/store/8y4zi0g9m64gjyagliqfvvlxmlqrhbgp-aria2-test'
> ```

Also, this is just because you did not do

`touch $out`

or something like that at the end of the test.

That means you told it that it failed. It did not produce a derivation.

wrapperModules/a/aria2/module.nix
```diff
@@ -0,0 +1,57 @@
...
+    };
+  };
+  config = {
+    package = pkgs.aria2;
+    flags = {
+      "--conf-path" = configFile;
+    };
+    flagSeparator = "=";
+    drv = {
+      buildPhase = ''
```
BirdeeHub, line 45: Also, I think I may have thought of a way to handle this "multi output with out not as the default" thing properly, but it will be at least a week or 2 until I can add it properly.

I will update it here too when I add it, it should get shorter and will feel more robust. It would also help with things like the zsh wrapper a little bit, where they have to redefine wrapperFunction because of stuff nixpkgs does to some of the shell drvs and I will do the update for them there too, if it has been submitted by then

In the meantime this is fine (although you shuld use `config.binName` to target it, something I also plan to streamline with my idea but your module won't need to use for that after the change)

wrapperModules/a/aria2/module.nix
```diff
@@ -0,0 +1,57 @@
...
+}:
+let
+  formatLine =
+    n: v:
+    let
+      formatValue = v: if builtins.isBool v then (if v then "true" else "false") else toString v;
+    in
+    "${n}=${formatValue v}";
+
+  configFile = pkgs.writeText "aria2Wrapped.conf" (
```
BirdeeHub: lets do the `passAsFile` trick you did for `yt-dlp` here as well

and add the copy command to our (already existing) buildPhase

https://github.com/BirdeeHub/nix-wrapper-modules/blob/abb9fe7e20adcf3c854c03d101e5d18c6d49457c/wrapperModules/y/yt-dlp/module.nix#L64-L72

That way people can place extra values into the derivation using the other `drv.<idk>` values and then target them with `placeholder "out"` or similar

I am not sure how relevant this capability is to aria2 in particular, but it is easy to do, and might be useful for somebody, and you have already done it once and so know how to do it already.

They do have a very simple config file format which is nice for us I guess even though it is "custom" i.e. not json yaml or toml or something like that.

---

UNLESS

If doing this means you need to copy or link the file into BOTH `$bin` and `$out` for it to work correctly for both outputs then we should leave it as is.

But I am pretty sure you don't I think you can just pick one of them. So, yeah, try it, if it doesn't work let me know and we will go with this current version instead

rachitvrma:
```nix
{
  flake.wrappers.aria2 =
    {
      wlib,
      lib,
      pkgs,
      config,
      ...
    }:
    let
      formatLine =
        n: v:
        let
          formatValue = v: if builtins.isBool v then (if v then "true" else "false") else toString v;
        in
        "${n}=${formatValue v}";

      /*
        configFile = pkgs.writeText "aria2Wrapped.conf" (
          lib.concatStringsSep "\n" (lib.mapAttrsToList formatLine config.settings)
        );
      */
    in
    {

      options = {
        settings = lib.mkOption {
          type =
            with lib.types;
            attrsOf (oneOf [
              bool
              float
              int
              str
            ]);
          default = {
            continue = true;
            dir = "/home/krish/Downloads";
            input-file = "/home/krish/Downloads/.aria_downloads.conf";
            file-allocation = "none";
            log-level = "warn";
            max-connection-per-server = 4;
            min-split-size = "5M";
            on-download-complete = "exit";
            auto-file-renaming = false;
          };
          description = "Settings to be wrapped with aria2 binary";
        };
      };
      config = {
        package = pkgs.aria2;
        binName = "aria2c";
        flags = {
          "--conf-path" = "${placeholder "bin"}/${config.binName}-settings.conf";
        };
        flagSeparator = "=";
        drv = {
          renderedSettings = lib.concatStringsSep "\n" (lib.mapAttrsToList formatLine config.settings);
          passAsFile = [ "renderedSettings" ];
          /*
            buildPhase = ''
              runHook preBuild
              cp $renderedSettingsPath "$out/${config.binName}-settings.conf"
              runHook postBuild
            '';
          */
          buildPhase = ''
            runHook preBuild
            cp $renderedSettingsPath "$bin/${config.binName}-settings.conf"
            rm $bin/bin/aria2c
            cp $out/bin/aria2c $bin/bin
            runHook postBuild
          '';
        };
      };
    };

}
```

Is this what you meant?

BirdeeHub: Yeah exactly. Does that work if you do `aria2.wrapper.out` as well or just `aria2.wrapper` and/or `aria2.wrapper.bin`?

If it works with all of those, then lets go with that, otherwise what you had before was fine.

Doing so lets people refer to `${placeholder "out/bin/man"}` from within the settings option, which may be useful for somebody combined with some hook like `drv.postBuild` or `drv.installPhase` or whatever

I am not sure how applicable that capability that is to aria2 specifically. Can you provide a script for something via any of the options? That would make such a capability fairly valuable.

---

basically, because it currently uses `pkgs.writeText`, `placeholder "out"` in it would refer to THAT derivation, and not the final wrapper derivation.

If it makes the code too weird to avoid this, and/or there is already a specific nixpkgs builder for a filetype but it has no option to generate to a string instead of a file, I don't worry about it too much, but if it is basically the same either way, putting the text directly into the final drv and doing passAsFile is better.

rachitvrma: Building aria2 and aria2.bin does generate aria2c-settings.conf but not aria2.out. 

I don't have a concrete script example in mind where `${placeholder "out"}` would be needed through one of the options. In this case, `out` only contains `/share/bash-completion`, which just provides bash completions for `aria2c`, so `${placeholder "out"}` would mostly just refer to that directory and the wrapped binary.

BirdeeHub: When you say, it does not generate, do you mean, it generates it to $bin, which the wrapper script points to, and that is on your computer but not in the ./result directory? Or do you mean, it does not generate it at all?

Im going to run it actually nix is perfectly reproducible for a reason XD

BirdeeHub: Ok, so I tried it. It does work regardless of which you pick.

It does not generate the generated conf file to `./result` for the `out` output

However, it does still generate a wrapper script which contains the flag which points to the `bin` output which does have the generated file, and it does still build that `bin` output in the store.

(I did a nix gc, then copied what you had in that comment into my config, and did `nix build .\#aria2.out` to test that)

Lets go with the one that makes placeholders work.

If we wanted to get really fancy, we could copy it to its own `conf` output (you would need to then also set `outputs = (config.package.outputs or [ "out" ]) ++ [ "conf" ];`) and then it wouldnt be in either of them when you looked and you could build it separately XD but lets not overcomplicate it further, unless you think that sounds interesting and cool to do, copying it to `bin` is fine.

rachitvrma: You mean something like this?
```nix
{
  flake.wrappers.aria2 =
    {
      wlib,
      lib,
      pkgs,
      config,
      ...
    }:
    let
      formatLine =
        n: v:
        let
          formatValue = v: if builtins.isBool v then (if v then "true" else "false") else toString v;
        in
        "${n}=${formatValue v}";
    in
    {

      options = {
        settings = lib.mkOption {
          type =
            with lib.types;
            attrsOf (oneOf [
              bool
              float
              int
              str
            ]);
          default = {
            continue = true;
            dir = "/home/krish/Downloads";
            input-file = "/home/krish/Downloads/.aria_downloads.conf";
            file-allocation = "none";
            log-level = "warn";
            max-connection-per-server = 4;
            min-split-size = "5M";
            on-download-complete = "exit";
            auto-file-renaming = false;
          };
          description = "Settings to be wrapped with aria2 binary";
        };
      };
      config = {
        package = pkgs.aria2;
        binName = "aria2c";
        outputs = (config.package.outputs or [ "out" ]) ++ [ "conf" ];
        flags = {
          "--conf-path" = "${placeholder "conf"}/${config.binName}-settings.conf";
        };
        flagSeparator = "=";
        drv = {
          renderedSettings = lib.concatStringsSep "\n" (lib.mapAttrsToList formatLine config.settings);
          passAsFile = [ "renderedSettings" ];

          buildPhase = ''
            runHook preBuild
            mkdir -p $conf
            cp $renderedSettingsPath "$conf/${config.binName}-settings.conf"
            rm $bin/bin/aria2c
            cp $out/bin/aria2c $bin/bin
            runHook postBuild
          '';
        };
      };
    };

}
```

BirdeeHub: yeah lol does that look better or worse to you? This way, neither one will have it as a file in their directory, but both can refer to it and use it, and you can build it separately like `nix build .\#aria2.conf` which is basically the same behavior as it would have been with `writeText`, except built as part of this derivation so placeholders work 

Either of the last 2 you put here work. It is entirely up to you as to which you prefer.

rachitvrma: Yeah, this looks cool. I never would have thought that out. Thanks for the suggestion.

BirdeeHub: I had to do something like that for the docs but with __structuredAttrs (note the unsafeDiscardReferences on the generated output, which is why I had to make it a separate output)

But yeah, basically, once you start thinking about a wrapper module as if it was a derivation which exposes module options, and has a makeWrapper generator in it, you start to get even what the more complicated wrappers are doing. Im pretty pleased with the mdbook module tbh its kinda a weird one to wrap.

https://github.com/BirdeeHub/nix-wrapper-modules/blob/2c0a3b795766b25d15ad1ec9eef3a06160154c9a/ci/docs/default.nix#L47-L69

wrapperModules/a/aria2/module.nix
```diff
@@ -0,0 +1,56 @@
...
+          int
+          str
+        ]);
+      default = { };
+      description = "Settings to be wrapped with aria2 binary";
+    };
+  };
+  config = {
+    package = pkgs.aria2;
+    binName = "aria2c";
```
BirdeeHub, line 35: It should detect this correctly? It is fine to specify it, as it removes the dependency tree of this value, but if it is detecting exePath as bin/aria2c it will detect binName as aria2c

But, this is fine, technically better even, you don't have to change this. I just want to make sure my code works rq and is in fact detecting this like it should

Edit: it is. All good. This is fine, and my code works still. 👍

wrapperModules/a/aria2/module.nix
rachitvrma: I think I gotta read all these wrapper modules. They look interesting, tbh. I have only very recently started out with nix, and on top of that spent very less time with wrappers.

BirdeeHub: Most of them are simple, some of them are interesting. the xplr one is kinda cool. It can also do lua and fennel code. the mdbook one as I mentioned. and then some of them do their own generation, like the niri one is pretty extensive, generating kdl. And the vim one is interesting because it uses wlib.makeCustomizeable which is a generalized lib.makeOverrideable. atool is interesting but short, uses wrapperVariants. And then of course neovim where I redefined both builderFunction and wrapperFunction and also there is a deferredModule option and a 2 layer DAG that gets sorted as 1 thing as part of a 130 line pipe with another pipe in the middle of it lol

Some of the ones I wrote are probably the weirdest so far, as is to be expected. But Im sure stuff like firefox will also be crazy XD Although maybe not who knows

rachitvrma: There's a firefox and fish builtin wrapper: wrapFirefox and wrapFish. Maybe, start from there. I saw that you have a starship wrapper in your birdeeSystems repo. That'd be cool to put in here as well.

EDIT: I have been trying to wrap fish using that function, but it gets weird fast.

BirdeeHub:
This looks good to me! going to merge it now I suppose :)

wrapperModules/a/aria2/module.nix
BirdeeHub: For the starship one, I was trying to figure out how to make a thing that can be sourced by nushell so I can have all 4 of the shells as options to create a sourceable file

but I don't really use nushell

So I never got around to figuring that out XD But that would be why it never ended up here.

Otherwise, its just another toml wrapper, its like 20 lines anyone can do that XD So I left it up to see if I get to it before someone else submits it basically.

I would be fine with just another toml wrapper, but if that is what it is to be, then Im not going to be the maintainer of it until I get around to adding sourceable things to mine and then add it there and add myself as secondary XD

BirdeeHub: I also have a bemenu module, but I haven't yet swapped from dmenu to bemenu because I don't know how to do a recency script for it, so I haven't tested my bemenu module XD But Im pretty sure that one works too. I just don't know if it needs more or not.

Im sure I will get to all these things at some point, but I have stuff to do other than this, so I have to prioritize.

On that note, but still about the repo

Priority 1 for me right now for this nix-wrapper-modules project is figuring out the "multi-outputs with out not as the default" thing that has you copying `bin/aria2c` from `out` to `bin` rather than using some kind of option or mechanism for that. And as a corrolary to that, people redefining wrapperFunction needing to manually construct the path to target and output to.

Once I do that, I might have an idea to improve the handling of enabling and disabling stdenv phases

And then I will return to improving docgen for a little bit, and either I finish that, or work on services and then finish that.

Once I do that, I need to start the process of getting it into nix-community, which should not be hard, and then I will be working on getting issue templates which ping the right people, and proper caching for tests. And a proper contributing guide, and maybe start seeing who else can be allowed to merge stuff, (until it is in nix-community only I will have merge privileges though, so I have to do that sometime sooner than later before it gets out of hand), etc.

But that is my list of current TODOs for just this repo in particular :rofl:

But once I do those, I can mostly sit back and review submissions and occasionally submit more wrapper modules for random things, at least until I get my next grand idea whatever that is XD

rachitvrma: I have been trying to write a fish wrapper not using the wrapFish function....

BirdeeHub: Shells are hard.

They work until you put them as your default user shell and then things go wonky, at least for zsh and maybe bash

We figured out how to do it for zsh and bash but it takes a bit of work, you can go check out the PR for the zsh wrapper to see why, and also the discussion that is referenced a few times from the PR

I am assuming using the wrapFish function from nixpkgs would work with nixpkgs.

one would hope. If it doesn't, you will have to do the thing being done in the zsh module too but for fish XD but if it works, then, nice, you get to skip that XD

BirdeeHub: And also, nixpkgs has all sorts of wrapper functions for stuff.

Some are good, some are bad (wrapNeovim), some of them you can modify afterwards, others you cant, half of them are undocumented, some of them are separate functions, some of them are provided as attributes on the package itself. Of the ones provided as attributes on the package itself, you can only call like half of those multiple times and usually they don't extend still. Its crazy XD

I didn't know there was a wrapFish until you told me there was a wrapFish

Honestly, I probably should check if there is a wrapZsh (although, I like the current approach we are taking for zsh so, its fine)

Edit: there is no wrapZsh I don't know how to feel

but yeah that is a lot of the motivation for this repo to be honest. To unify all the configuration mess nixpkgs has in the form of wrapper generators into a cohesive and usable and searchable and documented thing. In the process of doing so, its probably going to eat most of home manager's job. But for now, it has like 30 programs so it will take a while :rofl:

rachitvrma: The main problem with wrapping fish, and it might sound a bit personal, is wrapping `.theme` files to load with `fish_config theme choose <name of theme>`. As of yet, I haven't found a way to do that. 

wrapFish works pretty nicely, tho. I was thinking of using wrapFish to wrap fish using your default modules. That way I can export `self.packages.${pkgs.system}.fish` without having to use wrapFish multiple times. 

EDIT: it's mainly because I don't know what the nixpkgs `wrapFish` outputs, tbh, that I am thinking of this approach. I think the main intention of `wrapFish` was to use fish in a development environment, and of course, as mentioned on their page: to test plugins and other stuff. 

EDIT2: Would you be open to PRs containing aesthetics adding shell tools wrappers, like eza and vivid? They are really small shell tools and add to terminal aesthetics, but since home-manager provides modules for them, I thought having wrapper modules could be a good idea as well (?)

BirdeeHub: [wlib.makeCustomizable](https://birdeehub.github.io/nix-wrapper-modules/lib/wlib.html#function-library-wlib.makeCustomizable) in combination with overrides may be helpful in jamming it into a wrapper module. This works regardless of if you import `wlib.modules.default` actually but does not conflict with it if you do.

See [the vim module](https://github.com/BirdeeHub/nix-wrapper-modules/blob/733053067a3a8baf96a7ba05e970b038105fd567/wrapperModules/v/vim/module.nix#L66-L86) for inspiration

Unfortunately, or fortunately I'm not sure, I don't really have an opinion on themes, that's all you.

PR #329, opened 2026-03-09
feat(wrapperModules.htop): init

AlexLov:
Add wrapper for `htop` app pretty similar to `btop`'s wrapper.

### DIFF
--- a/wrapperModules/h/htop/module.nix
+++ b/wrapperModules/h/htop/module.nix
@@ -22,13 +22,12 @@ let
   mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
   toHtopConf = lib.generators.toKeyValue { inherit mkKeyValue; };
 
-  htopConfig = pkgs.writeText "htoprc" (
-    lib.concatLines [
-      # header_layout must be the first in file (or at least just above) so column_meter* parameters can work
-      (toHtopConf (lib.filterAttrs (n: _: n == "header_layout") config.settings))
-      (toHtopConf (lib.filterAttrs (n: _: n != "header_layout") config.settings))
-    ]
-  );
+  configPath = "${placeholder config.outputName}/${config.binName}rc";
+  htopConfig = lib.concatLines [
+    # header_layout must be the first in file (or at least just above) so column_meter* parameters can work
+    (toHtopConf (lib.filterAttrs (n: _: n == "header_layout") config.settings))
+    (toHtopConf (lib.filterAttrs (n: _: n != "header_layout") config.settings))
+  ];
 in
 {
   imports = [ wlib.modules.default ];
@@ -59,7 +58,15 @@ in
   };
 
   config.package = lib.mkDefault pkgs.htop;
-  config.envDefault.HTOPRC = htopConfig;
+
+  config.drv.htopConfig = htopConfig;
+  config.drv.passAsFile = [ "htopConfig" ];
+  config.envDefault.HTOPRC = configPath;
+  config.drv.buildPhase = ''
+    runHook preBuild
+    cp "$htopConfigPath" ${configPath}
+    runHook postBuild
+  '';
 
   meta.maintainers = [ wlib.maintainers.alexlov ];
 }


wrapperModules/h/htop/module.nix
```diff
@@ -0,0 +1,65 @@
...
+      if value then "1" else "0"
+    else if builtins.isInt value then
+      toString value
+    else
+      throw "Unrecognized type ${builtins.typeOf value} in htop settings";
+
+  mkKeyValue = lib.generators.mkKeyValueDefault { inherit mkValueString; } "=";
+  toHtopConf = lib.generators.toKeyValue { inherit mkKeyValue; };
+
+  htopConfig = pkgs.writeText "htoprc" (
```
BirdeeHub: This looks alright to me, but we are building the file ourselves, and yet `${placeholder "out"}` would not be usable in the generated file.

It is somewhat of a tricky technicality to get right, but it is often quite useful, and if we are generating the file ourselves, I prefer we make it work. (If the only nixpkgs generator builds a file, thats fine, just use that, but when it gives us a string, we should do this)

In your case, that means, you would take just the string from this htopConfig

```suggestion
  configPath = "${placeholder config.outputName}/${config.binName}rc";
  htopConfig = (
```

And then instead of just setting envDefault, we can put the file contents into the final drv, thus making placeholders work.

```nix
config.drv.htopConfig = htopConfig;
config.drv.passAsFile = [ "htopConfig" ];
config.envDefault.HTOPRC = configPath;
config.drv.buildPhase = ''
  runHook preBuild
  cp "$htopConfigPath" ${configPath}
  runHook postBuild
'';
```

There are actually still some existing modules that I should move to do it this way as well.

I did not realize there was a lib.generators.toGitINI and lib.generators.toINI when I was making some of them.

And also I have been realizing over the last few days that this is not in CONTRIBUTING.md yet either and that it really should be.

If I figure out an easier way to do this, I will grep for passAsFile and do a treewide update (although, whatever I do wouldn't break this either)

AlexLov: I'm fairly new to nix and do not understand what are all these needed for. Anyway, I did the suggested changes.
But would appreciate some information about what is all these changes about and what problem do they solve.

BirdeeHub: > what problem do they solve

In your `settings` option, you might want to refer to somewhere within the wrapper derivation created by the module.

You would do this by using `"${placeholder "out"}"` (or another output other than `"out"`)

However, `pkgs.writeText` makes a derivation.

So, this means, within your `settings` option, `"${placeholder "out"}` would refer to THAT derivation, which the user cannot add stuff to, rather than the wrapper derivation created by the module itself, which the user CAN add stuff to.

In your own configs, this is likely not important. When making something for everybody, we should make it work (when we don't have to do backflips to do so anyway)

I will eventually add an option for this to make this less of a strenuous task.

When I do so, I will update this module as well, because I will search for `passAsFile` and update everywhere at once that I can.

---

How the changes work

You declare your string as a derivation attribute on the final wrapper derivation via use of `config.drv.htopConfig`

This would normally just create a `$htopConfig` variable containing the string in the derivation build script that can be used.

Inside that string, `${placeholder "out"}` now refers to the wrapper derivation, because it is now directly a part of the wrapper derivation.

BUT

Environment variables and command line arguments have a maximum length!

So, we use `config.drv.passAsFile = [ "htopConfig" ];`

This will create a `$htopConfigPath` variable as well! This avoids the maximum length problem.

We then copy that file into the derivation output (for consistency, the same one the wrapper will be output to, which is probably `"out"`).

---

To be clear, these are all features of derivations. We are simply setting them via module option. This is how `pkgs.writeText` achieves writing files of arbitrary length as well.

BirdeeHub:
Awesome. Thank you! I will run it again, and then merge it!

wrapperModules/h/htop/module.nix
AlexLov: Thank you for the detailed explanation! There is still a lot to learn about nix for me

BirdeeHub: Also, to add, some of the nixpkgs format generators only can create a derivation (file) and not a string.

For those, it is fine, it can be a later problem if it becomes one (there is usually a way around it anyway), but we should make the effort when we already have a generator that gives us a string in nixpkgs for that type of file (check `lib.generators`, the ones in `pkgs.formats` usually use those to create the file)

On that note, I actually have some that still need this to be done to them, that I missed when first creating the repo. I didn't know there was an ini generator that made a string and only noticed it recently and have been busy with more important items and also just, life.

I might wait until I add the option for doing this easier to do those, I might do it sooner.

I did the git one already though, I will push that today.

BirdeeHub:
BTW, the passAsFile thing

I updated your module to use the new constructFiles options!

https://github.com/BirdeeHub/nix-wrapper-modules/blob/ab1afeebe69df6fb54b0ef5b2d0eee8201c485d3/wrapperModules/h/htop/module.nix#L59-L63

AlexLov:
That looks great and clean!

PR #343, opened 2026-03-15
feat(modules.makeWrapper): allow null values in env and envDefault to remove entries

vinnymeller:
basically due to the fact that claude code is vibecoded slop (but slop that work pays for):

1. when new features are released, if `DISABLE_TELEMETRY` (defaults to "1" in the module here) is set, that gets in the way of them fetching some server-side config to _actually_ enable it for a given user
2. it doesn't check any actual VALUE of `DISABLE_TELEMETRY`. if i override `DISABLE_TELEMETRY` to equal "0", it still won't work (https://github.com/anthropics/claude-code/issues/33403#issuecomment-4043741041)

only option is to completely remove the `DISABLE_TELEMETRY` env var, which I dont believe is currently possible with the way its currently set up? or at least not that i could figure out

alternatively could just remove `DISABLE_TELEMETRY` from claude code's `envDefault` but idk I felt weird being the one to magically change telemetry from being off by default to on by default (if anyone other than me even uses that module)

this tries to basically make it just work the same as `flags`

### DIFF
--- a/lib/makeWrapper/default.nix
+++ b/lib/makeWrapper/default.nix
@@ -368,7 +368,9 @@ in
         ++ mapAndLiftDal "addFlag" (config.addFlag or [ ])
         ++ mapAndLiftDal "appendFlag" (config.appendFlag or [ ]);
     in
-    if sortResult then wlib.dag.unwrapSort "makeWrapper" unsorted else unsorted;
+    builtins.filter (v: !(v.type == "env" || v.type == "envDefault") || v.data or null != null) (
+      if sortResult then wlib.dag.unwrapSort "makeWrapper" unsorted else unsorted
+    );
 
   /*
     splitDal receives the dal as returned by `aggregateSingleOptionSet` and splits it into 2 lists, `args` and `other`.
--- a/ci/checks/env-null.nix
+++ b/ci/checks/env-null.nix
@@ -1,0 +1,38 @@
+{
+  pkgs,
+  self,
+}:
+
+let
+  wrappedPackage = self.lib.wrapPackage {
+    inherit pkgs;
+    package = pkgs.hello;
+    env.MY_NULL_ENV = null;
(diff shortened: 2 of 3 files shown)


lib/makeWrapper/default.nix
```diff
@@ -498,7 +500,7 @@ in
                 v.attr-name
                 d
               ]) v.data)
-          else if v.data or null == null || v.data or false == false then
+          else if v.data == false then
```
BirdeeHub: Why do we need to change this part?

In case people add in a flag item before calling this, we should leave this so it still supports the full signature.

lib/makeWrapper/default.nix
```diff
@@ -368,7 +368,9 @@ in
         ++ mapAndLiftDal "addFlag" (config.addFlag or [ ])
         ++ mapAndLiftDal "appendFlag" (config.appendFlag or [ ]);
     in
-    if sortResult then wlib.dag.unwrapSort "makeWrapper" unsorted else unsorted;
+    builtins.filter (v: v.data or null != null) (
```
BirdeeHub: Should it only filter it out of `env` and `envDefault`? You could check `type` to see

BirdeeHub:
It seems like a reasonable idea, I had a few comments but we can do something like this yes.

vinnymeller:
think the answer to both is that i was trying to be clever, this should do i think

BirdeeHub:
It looks good to me, let me look it over a few more times because its in the core, but it should be fine like this.

And, honestly, I think none of the other ones can do null, so the first filter was probably technically fine, although idk, it seemed safer to only filter the things we wanted to filter.

But we would definitely still want to still have fixArgs handle it. So, yeah, thats back in there now, thank you thats good.

Edit: If it turns out someone actually needs the unfiltered list we can add a boolean option for it like sortResult, but at least for now, Im going to go with not doing that because I am not sure why they would.

PR #347, opened 2026-03-16
feat(wrapperModules.zsh): init

BirdeeHub:
@FluxZA about https://github.com/BirdeeHub/nix-wrapper-modules/pull/285 I finished it. I broke it up into several files too.

Can/should I list both of us as maintainer of it? Is that OK with you? Do you want that? You worked on it too, I just moved it around and added the last bit.

Also, if you would rather be the one to submit it, feel free to grab this and put it in your PR, it was just too much to put in a comment. Just want to know if I should add you as one of the 2 maintainers of it, or what you prefer around that. Is it possible to retroactively add someone as a co-author on a PR? I don't know how to make it give you credit on the github sidebar from here, but you'd be in the maintainers list and it would show you in the docs still.

---

Also, should we be doing .zprofile as well? Its basically the same thing as .zlogin, right? What should we do there. I can add that?

For context on `.zprofile` https://zsh.sourceforge.io/Intro/intro_3.html it says

> `.zlogin` is sourced in login shells. It should contain commands that should be executed only in login shells. `.zlogout` is sourced when login shells exit. `.zprofile` is similar to `.zlogin`, except that it is sourced before `.zshrc`. `.zprofile` is meant as an alternative to `.zlogin` for ksh fans; the two are not intended to be used together, although this could certainly be done if desired. `.zlogin` is not the place for alias definitions, options, environment variable settings, etc.; as a general rule, it should not change the shell environment at all. Rather, it should be used to set the terminal type and run a series of external commands (fortune, msgs, etc). 

So, tbh, not really sure what to do, do I do it for completeness anyway? Do I avoid it because its basically just bloat? Im currently leaning towards not doing that, especially because technically the user could use constructFiles and add it themselves

```nix
  config.constructFiles.zprofile = {
    relPath = "${config.zdotFilesDirname}/.zprofile";
    content = "";
  };
```

---

And any boolean options for toggling system config locations from being sourced that I can add to the zshenv or initial binary wrapper which I missed?

---

Is there anything else needed for proper nixos and home manager compatibility? What does it look like to use this thing? I will be messing with this a while probably. But at the very least, nothing we add will get removed in any unintended scenario I am aware of.

I want to not deal with plugins and external integrations and stuff like that, that is easy enough to add later to the config files via the options, I want to make sure it blends right into nixos and home manager as easily as possible before worrying about that.

I am pretty sure all the nixos stuff is figured out? Home manager I am less sure, I might need to source an extra thing or 2 if home manager was installed standalone vs bundled into your nixos config

### DIFF
--- a/wrapperModules/z/zsh/module.nix
+++ b/wrapperModules/z/zsh/module.nix
@@ -155,6 +155,22 @@ in
       this directory if they exist: `.zshenv`, `.zshrc`, `.zlogin` and `.zlogout`.
     '';
   };
+  options.hmSessionVariables = lib.mkOption {
+    type = types.nullOr wlib.types.stringable;
+    description = ''
+      Absolute path of the `hm-session-vars.sh` script to be loaded.
+
+      For standalone home manager setups to source `~/.nix-profile/etc/profile.d/hm-session-vars.sh`
+
+      This allows home manager to provide things such as `home.sessionVariables` to the shell.
+    '';
+    example = lib.literalMD ''
+      ```nix
+      "''${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
+      ```
+    '';
+    default = "~/.nix-profile/etc/profile.d/hm-session-vars.sh";
+  };
   options.zshenv = lib.mkOption {
     description = ''
       Specifies a file which will be sourced as part of the local `.zshenv` file.
--- a/wrapperModules/z/zsh/zshenv.nix
+++ b/wrapperModules/z/zsh/zshenv.nix
@@ -15,9 +15,9 @@ let
     if [[ -v __WRAPPED_ZSHENV_SOURCED ]]; then return; fi
     __WRAPPED_ZSHENV_SOURCED=1
 
-    ${lib.optionalString (config.homeManagerProfileDir != null) ''
-      if [[ -f ${config.homeManagerProfileDir}/etc/profile.d/hm-session-vars.sh ]]; then
-        . ${config.homeManagerProfileDir}/etc/profile.d/hm-session-vars.sh
+    ${lib.optionalString (config.hmSessionVariables != null) ''
+      if [[ -f ${config.hmSessionVariables} ]]; then
+        . ${config.hmSessionVariables}
       fi
     ''}
 


wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,166 @@
...
+}:
+let
+  inherit (lib) types;
+  rcfile = lib.types.submodule {
+    options = {
+      path = lib.mkOption {
+        type = lib.types.nullOr wlib.types.stringable;
+        default = null;
+        description = ''
+          `.path` is sourced before `.content`
```
zenoli: Just by looking at the descriptions, it is hard to know what these options do.

Ignoring descriptions, I would guess that you can define a path to a file to be sourced, and also some content, which will also be sourced. The current description would only describe that if both are defined the content is sourced first?

If so, the description should read:

"Path to an rc file to be sourced. If `.content` is defined as well, the config at `.path` will be sourced before `.content`"

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,166 @@
...
+    }
+  );
+  config.meta.maintainers = [
+    wlib.maintainers.fluxza
+    wlib.maintainers.birdee
+  ];
+  options.zdotFilesDirname = lib.mkOption {
+    type = lib.types.str;
+    default = "${config.binName}-dot-dir";
+    description = "Name of the dirctory which is created as the zdotdir in the wrapper output";
```
zenoli:
```suggestion
    description = "Name of the directory which is created as the zdotdir in the wrapper output";
```

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,166 @@
...
+    description = ''
+      Specifies a file which will be sourced as part of the local `.zshenv` file.
+
+      To be sourced after its corresponding path in `config.zdotdir` if specified.
+    '';
+    default = { };
+    type = rcfile;
+  };
+  options.zshAliases = lib.mkOption {
+    type = types.attrsOf (types.nullOr wlib.types.stringable);
```
zenoli, line 184: When would we ever want something else than `attrsOf str` here?

If I see `attrsOf str` for me as a user it is immediately clear how to use it. The `stringable` and `null` types makes it less straight forward (I first have to look up the definition of stringable and semantics of using null, with only few added benefits)

I don't see the need for supporting `null` values (I guess you want a  way to disable aliases without having to remove/uncomment them?

BirdeeHub:
Do I need the home manager thing from [commit #2](https://github.com/BirdeeHub/nix-wrapper-modules/pull/347/changes/c86e1caeee723d23132805a8e99e6d2df1b3afa0) of this PR?

wrapperModules/z/zsh/module.nix
BirdeeHub: If you put a script created via `pkgs.writeShellScript` as the value of an alias, with `lib.types.str` you need to put it in `"${}"` first. With `wlib.types.stringable` you do not. Otherwise, you can basically treat it like a string.

It is basically a `lib.types.str` which takes anything that can be `"${interpolated}"` as a string instead of just a string.

---

`null` is something that has been asked for on several similar options.

Basically, if you set it to something, `null` can then be chosen from another module or `.wrap` call to then disable that thing later.

See: https://github.com/BirdeeHub/nix-wrapper-modules/pull/343 as an example of people asking for `null` in similar scenarios. Mostly for overriding values already set in other modules.

BirdeeHub:
> I was not aware of the constructFiles option and do not fully understand it and the `outPath`, `path`, `relPath` sub-options. This part was def more straight forward to read in the original PR. Cannot tell if these options make things more complicated than it need be or it is simply a skill issue on my side.

They are VERY new. They collapse quite a bit of boilerplate from making placeholder paths work nicely both inside and outside of the wrapper module.

Basically, you set `relPath` and `content` and it handles all the `passAsFile` stuff for you. That is what makes placeholders work in the `.content` options, as they are built by the final derivation. Then they also give an easy handle to grab them by, `.path` if you are using it inside the module to put into the derivation, `.outPath` if you are grabbing something like ZDOTDIR from outside of the derivation. The previous implementation with `wlib.types.file` did not do this, placeholders in them would not point to the expected location. It made it look nicer, and fit more nicely with `wlib.types.file`, but also less of a complete solution.

i.e. `constructFiles` turns:

```nix
  config = {
    package = pkgs.yt-dlp;
    flags = {
      "--config-location" = "${placeholder config.outputName}/${config.binName}-settings.conf";
    };
    drv = {
      renderedSettings = renderSettings config.settings;
      passAsFile = [ "renderedSettings" ];
      buildPhase = ''
        runHook preBuild
        mkdir -p ${placeholder config.outputName}
        cp $renderedSettingsPath "${placeholder config.outputName}/${config.binName}-settings.conf"
        runHook postBuild
      '';
    };
    meta.maintainers = [ wlib.maintainers.rachitvrma ];
  };
```

into

```nix
  config = {
    package = pkgs.yt-dlp;
    flags."--config-location" = config.constructFiles.renderedSettings.path;
    constructFiles.renderedSettings = {
      relPath = "${config.binName}-settings.conf";
      content = renderSettings config.settings;
    };
    meta.maintainers = [ wlib.maintainers.rachitvrma ];
  };
```

As in this PR https://github.com/BirdeeHub/nix-wrapper-modules/commit/575386b194529b92cc7708a44b8bba21a025c4ee

The reason there is a `.path` AND a `.outPath` is because `.path` uses a placeholder, and `.outPath` uses `config.wrapper`. Placeholders resolve to the derivation they end up in, not the one they came from. So, the placeholder is suitable for grabbing inside of the module to reference the final drv path, and `outPath` uses `config.wrapper` and you would grab it from outside.

They are provided by the (new) constructFiles helper module https://birdeehub.github.io/nix-wrapper-modules/modules/constructFiles.html

wrapperModules/z/zsh/zshenv.nix
```diff
@@ -0,0 +1,210 @@
...
+    then
+      HELPDIR="${pkgs.zsh}/share/zsh/$ZSH_VERSION/help"
+
+      # Tell zsh how to find installed completions.
+      for p in ''${(z)NIX_PROFILES}; do
+          fpath=($p/share/zsh/site-functions $p/share/zsh/$ZSH_VERSION/functions $p/share/zsh/vendor-completions $fpath)
+      done
+    fi
+
+    ${lib.optionalString (config.homeManagerProfileDir != null) ''
```
BirdeeHub: > The only thing that came to mind was to support [home.sessionVariables](https://github.com/nix-community/home-manager/blob/master/modules/programs/zsh/default.nix#L470) which might be useful. But it could also be easily added later on.

This is the `home.sessionVariables` stuff done correctly, right? I think? I had to make an option for it because we don't really have that path? Should the option take the final path, or just the dir like they have it and I copied here?

Should I make it default to `~/.nix-profile` or `null`? Those are my 2 options for default as far as I can tell.

wrapperModules/z/zsh/module.nix
BirdeeHub: If that is helpful for people first reading it, all for it

BirdeeHub:
I guess, the last question(s) left is/are

Did I do the `home.sessionVariables` stuff correctly, and related, should I make `config.homeManagerProfileDir` default to `~/.nix-profile` or `null`? Those are my 2 options for default as far as I can tell, and it only does anything if the user has standalone home manager (otherwise the /etc/ config from nixpkgs `programs.zsh.enable = true` takes care of it)

In the zshenv I have

```nix
    ${lib.optionalString (config.homeManagerProfileDir != null) ''
      if [[ -f ${config.homeManagerProfileDir}/etc/profile.d/hm-session-vars.sh ]]; then
        . ${config.homeManagerProfileDir}/etc/profile.d/hm-session-vars.sh
      fi
    ''}
```

and then I have the option

```nix
  options.homeManagerProfileDir = lib.mkOption {
    type = types.nullOr wlib.types.stringable;
    default = "~/.nix-profile";
    description = ''
      For standalone home manager setups to source the `''${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh`
      file, provide the path to the `home.profileDirectory`.

      By default, it should be `~/.nix-profile`
    '';
  };
```

And should `homeManagerProfileDir` take what it does now, or the whole path?

---

[I have converted my config to dogfood it for a bit](https://github.com/BirdeeHub/birdeeSystems/commit/24027bccf26061e6d583cdef5afaa64f316dcb6d)

I really need to finish making my config more dendritic, but that isnt important.

BirdeeHub:
> should I make config.homeManagerProfileDir default to ~/.nix-profile or null

`~/.nix-profile`

It is better to do 1 extra existence check for when people do not set it to null when using home manager as a nixos module, as opposed to sessionVariables and such not appearing in standalone.

> And should homeManagerProfileDir take what it does now, or the whole path?

This is what home manager uses, should be fine.

Im also more sure about the ordering of the files being ran. We have .content from nix at the end but we also have .runShell from nix at the beginning. So, before and after from nix is covered, and the order of dir then path is fine.

I am also more sure of what is going on and that I am sourcing the correct file for that.

BirdeeHub:
I think this is good.

The only question left, at least for now, is if @FluxZA still wants to be one of the maintainers on it.

FluxZA:
Hi All, sorry for the radio silence on my part, a bunch of life stuff hit all at the same time. I am happy for you to commit this as your PR and I will stay a maintainer. I haven't yet had a chance to go through all of it, so I can add anything, but I will make sure to familiarise myself with it this week.

BirdeeHub:
Of course!

I'll give you some time to get familiar and stuff too before I merge it, I need to look over the docs again I think.

I added some features and wanted to put them to the test, and I wanted a zsh module that actually works. Just setting ZDOTDIR was so tempting but I'm glad we waited until we did to do this.

I think I need to add some more, like, installation instructions to the description, not a ton though, but it definitely works now I was able to use it in my config to replace basically all my zsh config

You still have to do at least some of the following in nixos to use it as the default shell for all the things

```nix
  # nixOS options
  environment.pathsToLink = [ "/share/zsh" ]; # <- completion stuffs
  programs.zsh.enable = true; # <- just for /etc/set-environment basically
  users.defaultUserShell = config.wrappers.zsh.wrapper; # <- default for all
  users.users.${username}.shell = config.wrappers.zsh.wrapper; # <- default for 1
  environment.systemPackages = [ config.wrappers.zsh.wrapper ]; # <- you also need to install it regularly alongside.
```

But otherwise, it just works, as far as I can tell so far.

And of course, if you aren't using it as the default shell, you don't have to do anything special, just use the package like you would.

zenoli:
> You still have to do at least some of the following in nixos to use it as the default shell for all the things

Could we provide a `mkInstallShellWrapper' or similar which sets these nixos configs to provide a standardized way this wrappes is supposed to be integrated with nixos as a user shell?

BirdeeHub:
> > You still have to do at least some of the following in nixos to use it as the default shell for all the things
> 
> Could we provide a `mkInstallShellWrapper' or similar which sets these nixos configs to provide a standardized way this wrappes is supposed to be integrated with nixos as a user shell?

Maybe? But, 2 of the 5 lines were zsh specific. So, I am not sure exactly how to do so in a way that is both generic and does the whole job.

I think that should be in another PR if we do that. Maybe when we do bash and fish so we have 2 more examples, we will figure out how to generically install them? Unless someone finds a way first I guess.

zenoli:
Yes, I agree. Let's generalize this once we have a better understanding on how to generalize it

wrapperModules/z/zsh/module.nix
zenoli: Makes sense. Thanks.

Off topic:
I guess there is a difference in the merging behavior between`stringable` and `oneOf str path ...`?

BirdeeHub: > I guess there is a difference in the merging behavior between stringable and oneOf [ str path ... ]?

It merges like `lib.types.str` does. It does not stringify the value for you, it leaves the value as it was, it checks using `lib.isStringLike` to see if it is a thing that can be interpolated or not, and accepts or rejects the value based on that.

It is useful, because a ton of stuff takes drv or paths or a string (either a path or a command), and it is annoying to have to stringify all the drvs when the option is going to stringify it anyway.

wrapperModules/z/zsh/variants.nix
zenoli, line 1: No idea what this is used for 🤷 
Would you mind giving some insights?

wrapperModules/z/zsh/zshenv.nix
```diff
@@ -0,0 +1,209 @@
...
+      if [[ -f ${lib.escapeShellArg config.zshenv.path} ]]
+      then
+        source ${lib.escapeShellArg config.zshenv.path}
+      fi
+    ''}
+  '';
+
+  split = wlib.makeWrapper.splitDal (wlib.makeWrapper.aggregateSingleOptionSet { inherit config; });
+
+  wrapcmd = partial: "echo ${lib.escapeShellArg partial} >> \"$2\"";
```
zenoli, line 57: I tried to follow the conversation in the original PR where I think you  outlined what needed to be redefined in the wrapperFunction.

Is there a "high-level" explanation of the problem you are solving here?

wrapperModules/z/zsh/module.nix
```diff
@@ -0,0 +1,213 @@
...
+  options.homeManagerProfileDir = lib.mkOption {
+    type = types.nullOr wlib.types.stringable;
+    default = "~/.nix-profile";
+    description = ''
+      For standalone home manager setups to source the `''${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh`
+      file, provide the path to the `home.profileDirectory`.
+
+      By default, `home.profileDirectory` should be `~/.nix-profile`
+    '';
+  };
```
zenoli: Just for comparison, in my wrapper I modeled it slightly differently:

```nix
    hmSessionVariables = {
      enable = lib.mkEnableOption "home-manager sessionVariables";
      scriptLocation = lib.mkOption {
        type = lib.types.str;
        description = ''
          Absolute path of the `hm-session-vars.sh` script to be loaded.
        '';
        default = "~/.nix-profile/etc/profile.d/hm-session-vars.sh";
      };
    };
```

I think specifying the script location makes the purpose of the option more clear. If a user would want to opt-out of this and not have home.sessionVariables be loaded, he would look for an option named "sessionVariables" and find nothing. We could also keep homeManagerProfileDir but also introduce another option based on homeManagerProfileDir called sessionVariablesScript.
Maybe group both options in a home-manager option set then?

```nix
options.homeManager = {
    profileDir = mkOption ...
    sessionVariablesScript = mkOption ...
}

File: wrapperModules/z/zsh/zshenv.nix (hunk 4 above)
Comment (zenoli) [reply wrapperModules/z/zsh/zshenv.nix]: Maybe have it enabled by default makes more sense. It is minimal bloat if you don't want it but saves users migrating from home-manager some headaches.

If we make it opt-in I fear that people will miss that this exists, especially because the current option name `homeManagerProfileDir` does not make it clear that the impact of setting it is to make it load some variables.

Thinking about it, I would even prefer something like:

```nix
options.homeManagerIntegration = {
  enable = ...
  profileDir = ...
  sessionVariablesScript = ...
```

but maybe we are already moving away form the original simplicity and is better suited in my opinionated wrapper.
But it would make the intent more clear: Enable stuff to make it compatible with home manager.

zenoli:
Thanks for the clarifications on `constructFiles`. I like it.

In my wrapper I currently generate the rcfiles using symlinkJoin and writeTextFile and simply set ZDOTDIR to the store path of the symlinkjoin derivation. This has the disadvantage that I cannot simply inspect the `result` folder and inspect zdotdir. 

`constructFiles` will directly build the files in the same folder as the wrapper, right?

wrapperModules/z/zsh/zshenv.nix
BirdeeHub: The one from the discussion where extraPackages was getting removed!

we need to generate the wrapper options into zshenv

so like

config.env or config.prefixVar and stuff

If you set anything nixos sets via that, it would get overridden

All those options need to go into zshenv

However ZDOTDIR itself must be set beforehand, and the cli args must be provided upon invocation.

So we use pkgs.makeBinaryWrapper to generate the main wrapper, and then put everything else into zshenv

This is us redefining wrapperFunction to make 2 files instead of 1 more or less.

BirdeeHub:
> Thanks for the clarifications on `constructFiles`. I like it.
> 
> In my wrapper I currently generate the rcfiles using symlinkJoin and writeTextFile and simply set ZDOTDIR to the store path of the symlinkjoin derivation. This has the disadvantage that I cannot simply inspect the `result` folder and inspect zdotdir.
> 
> `constructFiles` will directly build the files in the same folder as the wrapper, right?

yes it builds it into `${placeholder output}/${relPath}` (by default `output` is `config.outputName` which is usually `"out"`) (and if `.relPath` is an empty string it will omit the `/`, which will of course cause issues unless you make it its own output via `.output`, but hey why not let someone do `pkgs.writeText` as a wrapper module if they want)

https://github.com/BirdeeHub/nix-wrapper-modules/pull/347#discussion_r2949638913

I agree. I changed the default to `~/.nix-profile`, so that it is enabled by default.

https://github.com/BirdeeHub/nix-wrapper-modules/pull/347#discussion_r2949596831

Totally happy to do it this way. I will try out the double option way, and if its too much, do the absolute path. Although, before even trying it, I think I like just the absolute path, and I am ok with your name for it.

I think I might just use those 2 options you put?

```nix
  options.hmSessionVariables = {
    enable = lib.mkOption {
      type = types.bool;
      default = true;
      description = ''
        For standalone home manager setups to source `~/.nix-profile/etc/profile.d/hm-session-vars.sh`

        This allows home manager to provide things such as `home.sessionVariables` to the shell.
      '';
    };
    scriptLocation = lib.mkOption {
      type = types.str;
      description = ''
        Absolute path of the `hm-session-vars.sh` script to be loaded.
      '';
      example = lib.literalMD "`\${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh`";
      default = "~/.nix-profile/etc/profile.d/hm-session-vars.sh";
    };
  };
```

Edit:

Actually.... after using it...

just the absolute path or `null` The extra boolean is annoying. Either you want it null, or you don't care at all and won't touch it, or you will set it. There isnt really a scenario where it makes sense to enable it separately of setting it. 

```nix
  options.hmSessionVariables = lib.mkOption {
    type = types.nullOr wlib.types.stringable;
    description = ''
      Absolute path of the `hm-session-vars.sh` script to be loaded.

      For standalone home manager setups to source `~/.nix-profile/etc/profile.d/hm-session-vars.sh`

      This allows home manager to provide things such as `home.sessionVariables` to the shell.
    '';
    example = lib.literalMD ''
      ```nix
      "''${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
      ```
    '';
    default = "~/.nix-profile/etc/profile.d/hm-session-vars.sh";
  };
```

```nix
{ config, osConfig ? null, ...  }: {
  wrappers.zsh.hmSessionVariables = if osConfig == null then "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh" else null;
}
```

wrapperModules/z/zsh/variants.nix
BirdeeHub: So, normally config.wrapperVariants repeats the same options, for other files in the bin directory or elsewhere, or to create a second version of a wrapper with slightly different options.

You can set mirror = false so that it does not mirror the options.

I am using type merging to make it be false for all of them, because we cannot mirror the same wrapping thing there really. (mkOverride 1400 is 100 higher priority than mkOptionDefault but 400 lower than mkDefault)

I am also using type merging to add back the wrapperImplementation option we excluded from the main level of the options set.

It would allow us to add other programs and wrap them within the context of the zsh wrapper, i.e. with the zsh wrapper's placeholders

BirdeeHub:
Alright I think I am happy with it. Going to merge it once it finishes these checks!

wrapperModules/z/zsh/zshenv.nix
zenoli: Ah so with this wrapper extraPackages will directly be set in .zshenv and I won't have to handle it myself (as I did by intoducing a `runtimePackages` option)?

One thing I noticed when extending PATH in zshenv:

If you nest multiple shells, PATH will get convoluted because the extraPackages/runtimePackage are added on each invocation on zsh. I worked around this by checking if the new paths are already a prefix of PATH before extending them:

```zsh
      runtime_packages="${pkgs.lib.makeBinPath config.runtimePackages}"
      if [[ $PATH != ''${runtime_packages}:* ]]; then
        export PATH="''${runtime_packages}:$PATH"
      fi
 ```
 
 (I used `export PATH` here but the problem remains even if I remove it as it is already exported I think?)

Comment (BirdeeHub) [reply wrapperModules/z/zsh/zshenv.nix:57]: we have a variable that makes sure the file is only sourced once for each of the generated files.

https://github.com/BirdeeHub/nix-wrapper-modules/blob/8613abd9db99d92cd85dbe37cdca08e4ee82225f/wrapperModules/z/zsh/zshenv.nix#L15-L16

That should be fine. If it isnt let me know. We may have to call export on it for some sort of niche scenario, so that would be the first fix to try if you do experience an issue with that would be to call export on those.

Im about to test that out a bit more myself. I only tested a little bit of that tbh, launching zsh twice in a row

Edit:

Wait we do have this bug. I will try export.

Edit 2: nope export no good. Then it wont run anything the second time you launch it.

Should we NOT export path to begin with? We can do that!

```nix
      setvarfunc = /* zsh */ ''wrapperSetEnv() { typeset -g "$1=$2"; }'';
      setvardefaultfunc = /* zsh */ ''wrapperSetEnvDefault() { [ -z "''${(P)1+x}" ] && typeset -g "$1=$2"; }'';
      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { typeset -g "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { typeset -g "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
```

wrapperModules/z/zsh/variants.nix
zenoli: Is this related to https://github.com/BirdeeHub/nix-wrapper-modules/pull/347#discussion_r2949680618 ?

>I am also using type merging to add back the wrapperImplementation option we excluded from the main level of the options set.

So wrapper variants for zsh here need to re-define options but also have access to the modified wrapperImplementation from the other discussion?

wrapperModules/z/zsh/zshenv.nix
zenoli: I think this guard only protects against re-sourcing .zshenv from within the same shell but when spawning a new shell `__WRAPPED_ZSHENV_SOURCED` won't be set as it is not exported. Exporting it would probably fix it but I would not do this because:

1. it would deviate from how this pattern is applied in hm/nixos and they might have had their reasons to not export it.
2. I can see it become annoying especially when tweaking with the config and spawning new shells. If it is exported zshenv will be sourced only when opening a new shell window/tmux pane etc.

BirdeeHub: for now, it happening twice is better than it not happening at all, but, I will find a way to guard it

BirdeeHub:
```nix
      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3$2"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$2$3"* ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
```

BirdeeHub: https://github.com/BirdeeHub/nix-wrapper-modules/pull/355

BirdeeHub: doing export on the thingy just makes it so you lose your prompt if you call zsh twice so this is the better method

BirdeeHub: fixed and merged

BirdeeHub: Also do you also want to be on the maintainers list? Feel free to open a PR adding yourself to this or whatever really

It basically just gives you bragging rights (name goes in docs) and sometimes I will ping you if I have a question or need help XD

zenoli:
```nix
prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3$2"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$2$3"* ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
 ```
      
     Shouldn't it be `*"$2$3" in  the suffixvarfunc? It looks inconsistent with the pattern in prefixvarfunc? But maybe it is because of the `VARNAME=` stuff at the beginning and hence we cannot match from the beginning?
     
> Also do you also want to be on the maintainers list? Feel free to open a PR adding yourself to this or whatever really

No no its fine :-)

wrapperModules/z/zsh/variants.nix
BirdeeHub: They use the normal wrapping function. The base options for zsh itself use the custom one.

So, mirroring the args by default seemed unwise

BirdeeHub: https://github.com/BirdeeHub/nix-wrapper-modules/blob/54a14082c5eae3db70ec7d9af28f9d8d7b093471/wrapperModules/z/zsh/zshenv.nix#L208-L209

wrapperModules/z/zsh/zshenv.nix
BirdeeHub: I did `*"$2$3"*` for suffix because if the thing already was present anywhere in it, the previous one will win anyway, why add it again?

Honestly I probably should have done `*"$3"*`

BirdeeHub: Hmmm

Although, maybe some env vars actually read right to left? and if it was empty prior, the sep would not be there?

`"$3"*` and `*"$3"` it is.

BirdeeHub: If it turns out to be a problem still, I will do `*"$3"*` for both and give an extra flag of some kind for forcing them or doing some other kind of conditional you can use in the DAL

PR #356, opened 2026-03-18
fix(wrapperModules.zsh): improve prefix and suffix matching

BirdeeHub:
related to https://github.com/BirdeeHub/nix-wrapper-modules/pull/355

### DIFF
--- a/wrapperModules/z/zsh/zshenv.nix
+++ b/wrapperModules/z/zsh/zshenv.nix
@@ -102,8 +102,8 @@ let
     let
       setvarfunc = /* zsh */ ''wrapperSetEnv() { export "$1=$2"; }'';
       setvardefaultfunc = /* zsh */ ''wrapperSetEnvDefault() { [ -z "''${(P)1+x}" ] && export "$1=$2"; }'';
-      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3$2"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
-      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$2$3"* ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
+      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
+      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$3" ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
     in
     builtins.concatStringsSep "\n" (
       lib.optional (config.env or { } != { }) setvarfunc


wrapperModules/z/zsh/zshenv.nix
```diff
@@ -102,8 +102,8 @@ let
     let
       setvarfunc = /* zsh */ ''wrapperSetEnv() { export "$1=$2"; }'';
       setvardefaultfunc = /* zsh */ ''wrapperSetEnvDefault() { [ -z "''${(P)1+x}" ] && export "$1=$2"; }'';
-      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3$2"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
-      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$2$3"* ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
+      prefixvarfunc = /* zsh */ ''wrapperPrefixEnv() { [[ "''${(P)1}" != "$3"* ]] && export "$1=''${(P)1:+$3$2}''${(P)1:-$3}"; }'';
+      suffixvarfunc = /* zsh */ ''wrapperSuffixEnv() { [[ "''${(P)1}" != *"$3" ]] && export "$1=''${(P)1:+''${(P)1}$2}$3"; }'';
```
zenoli, line 106: @BirdeeHub I just noticed:
extraPackages are suffixed to PATH not prefixed. 
Is there a good reason to this?

In zsh for example, I integrated my custom direnv wrapper adding the wrapper to `extraPackages`. 

If I use my zshwrapper on a machine that has direnv already in path, this direnv binary will take precedence over my direnv wrapper which is not what I want.

There might be good reasons for suffixing them by default in general. But it would make sense to be able to opt-out of suffixing and use prefixing instead. Or should I just use prefixVar directly and not use extraPackages instead? (this is possible right?)

Aside from that, integrating my zsh wrapper into the official zsh wrapper was pretty seamless so far :-)

BirdeeHub: extraPackages suffixes to the path, because then dev shells win

This is very useful in things like the neovim wrapper where you want it to provide stuff like LSPs but then have a shell provide its own and have the wrapped program pick up on that.

You should make another option like extraPackages but for prefixing them, unless you have a better idea of how to handle the option and change which it does per item that we can find out how to add.

zenoli: The dev shells argument make sense. I think the situations is different when we are wrapping shells though:

- I launch my shell and my zsh-wrappers' extraPackages are added to PATH
- I enter a devshell and the devshell packages get prefixed to PATH so they will win anyway

So in the case of shell wrappers, dev shells win independent of extraPackages are prefixed or suffixed in the zsh wrapper.

I introduced an extra option that prefixes them in my wrapper.

The easiest way to add this upstream would be to add an option `extraPrefixPackages` in addition to `extraPackages` but I don't like this name and the asymmetry (i.e. no `extraSuffixPackages`).

A cleaner solution would be to make extraPackages take either a list or a submodule  which consists of two sub-options "append" and "prepend" - each accepting a list of packages to be suffixed/prefixed accordingly.

If one provides a list directly the default behavior would be to suffix them to stay backwards compatible.

We could do the same pattern for "runtimeLibraries" (and any other options which could be generalized this way).

I am mainly dumping my thoughts here to see to discuss whether this is even a reasonable idea.

BirdeeHub: Yeah we should figure out how to do both in a nice way.

We could do a number of things, Ill brainstorm and give some examples here later today or something.

> I think the situations is different when we are wrapping shells though:

    I launch my shell and my zsh-wrappers' extraPackages are added to PATH
    I enter a devshell and the devshell packages get prefixed to PATH so they will win anyway


Only if the dev shell doesnt then launch the shell again IDK if it does or not.

PR #357, opened 2026-03-18
fix: zathura "options" option -> "settings" option

rachitvrma:
"options" option confuses with module-making option rather than a module option

### DIFF
--- a/wrapperModules/z/zathura/module.nix
+++ b/wrapperModules/z/zathura/module.nix
@@ -29,7 +29,7 @@ in
         ]);
       default = { };
       internal = true;
-      apply = x: if x != { } then lib.warn "'options' has been renamed to 'settings'" else x;
+      apply = x: if x != { } then lib.warn "nix-wrapper-modules zathura: 'options' has been renamed to 'settings'" else x;
     };
     settings = lib.mkOption {
       type =


BirdeeHub:
Would you mind adding a deprecation warning for that?

You would declare an `options` option that, if it contains any values, will warn telling people to change their config, and then alias to the new `settings` option.

That way, for people who worked around this issue by doing `config.options`, they would get a nice warning instead of it suddenly breaking.

But, yeah, most stuff is named settings for that reason, we should change it I agree. I didnt notice this when I was reviewing it.

BirdeeHub:
```nix
  config.settings = lib.mkIf (config.options != { }) config.options;
```
```nix
  options = {
    options = lib.mkOption {
      type =
        with lib.types;
        attrsOf (oneOf [
          bool
          str
          int
          float
        ]);
      default = { };
      internal = true;
      apply = x: lib.warnIf (x != { }) "DEPRECATION WARNING options -> settings" x;
    };
    settings = lib.mkOption {
```

(obviously that example warning message is crap and the real one should be slightly better)

BirdeeHub:
can you put the name zathura somewhere in the warning? It will help people

wrapperModules/z/zathura/module.nix
```diff
@@ -19,6 +19,19 @@ in
...
+        with lib.types;
+        attrsOf (oneOf [
+          bool
+          str
+          int
+          float
+        ]);
+      default = { };
+      internal = true;
+      apply = x: if x != { } then lib.warn "'options' has been renamed to 'settings'" else x;
```
BirdeeHub:
```suggestion
      apply = x: if x != { } then lib.warn "nix-wrapper-modules zathura: 'options' has been renamed to 'settings'" else x;
```

BirdeeHub:
I accidentally hit merge before I meant to but it is fine

PR #360, opened 2026-03-20
feat(wrapperModules.swayidle): init

rachitvrma:
add: Wrapper module for the swayidle program

### DIFF
--- a/wrapperModules/s/swayidle/module.nix
+++ b/wrapperModules/s/swayidle/module.nix
@@ -72,6 +72,16 @@ in
               Executes command when logind signals that the session should be unlocked
             '';
           };
+          idlehint = mkOption {
+            type = types.nullOr types.ints.positive;
+            default = null;
+            apply = v: if v == null then null else toString v;
+            description = ''
+              Set IdleHint to indicate an idle logind/elogind session after <timeout>
+              seconds. Adding an idlehint event will also cause swayidle to call SetIdleHint(false) when run, on
+              resume, unlock, etc.
+            '';
+          };
         };
       };
       default = { };


wrapperModules/s/swayidle/module.nix
```diff
@@ -0,0 +1,121 @@
...
+
+  options = {
+    timeouts = mkOption {
+      type = types.listOf (types.submodule timeoutModule);
+      default = [ ];
+      description = "List of idle timeout actions";
+    };
+
+    events = mkOption {
+      type = types.submodule {
```
BirdeeHub, line 43: Should this be a freeform submodule like `freeformType = lib.types.attrsOf (lib.types.nullOr wlib.types.stringable);`?

In case we missed some events?

I think we missed [`idlehint` event](https://man.archlinux.org/man/extra/swayidle/swayidle.1.en#EVENTS), if we made it a freeform type, it would let people submit nonexistent arguments, but it would also let people provide arguments we did not cover.

Alternatively, you could just add `idlehint`, and people can use extraArgs for any not covered

(use `apply = toString;` on the option declaration since it is a number I think and then you dont have to change any mapping either)

Although, I guess technically the user could already declare such an option themselves, it is a submodule, and then it would get mapped because of the way you did these, so, thats kinda cool.

Other than this one small question, it looks ready to go!

Honestly, I somewhat prefer just adding `idlehint` event, because then it tells people if they put an invalid one or not. And they could still declare an option there for a new one.

Unless I am misunderstanding and it is not actually provided in the same way as the other events?

rachitvrma: I think idlehint is provided the same way. I just couldn't find a way to test this option.

PR #365, opened 2026-03-23
feat(wrapperModules.fastfetch): init

rachitvrma:
add wrapper module for the
[fastfetch](https://github.com/fastfetch-cli/fastfetch) system information tool


BirdeeHub:
Yeah I think that is all fastfetch needs. Looks good to me I think!

PR #403, opened 2026-04-01
feat(wrapperModules.mpv): use --config-dir instead of separate flags for each file

iynaix:
Currently the module uses separate `--include` and `--input-conf` flags. While this works, it makes it very awkward to extend this module to write additional files such as script-opts settings. (The --script-opts flag is used to pass a long string of script-opt settings, instead of pointing to a script-opt directory: https://mpv.io/manual/stable/#options-script-opt) 

Switching it to use the `--config-dir` option will allow being able to write other files into the already created `mpv-config` directory and still have those work. This change is internal and does not affect functionality for existing users.

### DIFF
--- a/wrapperModules/m/mpv/check.nix
+++ b/wrapperModules/m/mpv/check.nix
@@ -4,7 +4,7 @@
 }:
 
 let
-  mpvWrapped =
+  mpvWithoutConfigDirectory =
     (self.wrappers.mpv.apply {
       inherit pkgs;
       scripts = [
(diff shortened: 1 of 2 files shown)


BirdeeHub:
If people were currently using the normal config dir location to put scripts, this would break their workflow.

You can provide scripts via `--script` flag.

```nix
config.flags."--script" = [];
```

The module's scripts existing `scripts` option is meant to receive scripts from `pkgs.mpvScripts`. It uses the `pkgs.mpv.override` machinery to do this. It actually cannot be passed scripts generated via `constructFiles` either. 

All other scripts should be added directly via `config.flags."--script" = [];`, and the values in that list can be a derivation, nix path, or just a string path to somewhere at runtime hopefully, doesn't matter.

It should, however, mention that in the docs for the existing `scripts` option. It mentions nothing of this whatsoever. And it should set `config.flagSeparator` or `config.flags."--script" = {...}: { sep = "="; };` for the user, so that users passing it themselves need only do `config.flags."--script" = [ idk ];`

So, what should probably be added instead of this change is

```nix
config.flags."--scripts" = { ... }: { ifs = ":"; };
```

And some stuff in the docs about  how the existing `scripts` option works and what it is for.

```nix
config.flags."--script" = [ ];
# and also, the ifs thing above makes this work too
config.flags."--scripts" = [ ];
```

If you want, we can make a separate `config.scriptFiles` option, to illustrate that the one that takes `pkgs.mpvScripts` takes a derivation but `scriptFiles` takes just a file

Another thing that could be done, but isn't really necessary, is we could reimplement [what the nixpkgs one does](https://github.com/NixOS/nixpkgs/blob/nixos-25.11/pkgs/applications/video/mpv/wrapper.nix) but using our module options instead, and use `pkgs.mpv-unwrapped` as the package.

iynaix:
That's a fair point, I did not consider the case where users would put scripts in the normal config location. To me, having the scripts in the normal config directory defeats the entire purpose of having a wrapper, but I can see people doing that. I'll maintain my own wrapper then.

BirdeeHub:
I mean, there is plenty that can be done with this one, I just think that we should be able to choose if the normal directory works or not.

Right now, the normal directory works, and we can choose whether we want it to work or not by providing an empty directory to it. Since we are using the --script --include and --input-conf options, it doesn't matter where --config-dir points. So we could have a boolean option.

The thing that is stopping me from immediately doing that, is that, there is an `pkgs.mpvScripts`, but it expects the value to be at a very specific sub path, it doesnt just take the value directly.

So it would need to have both options.

And if it needs both options, `options.flags."--script" = [ ];` is already a single option which takes a list of scripts.

But having such an option explicitly doesnt seem like the worst idea, simply because it would make usage clearer.

And regardless, you don't need to make an entirely separate wrapper for this, simply add the script files to that option and it should work.

---

If you wanted to make a PR adding such an extra `scriptFiles` option or something like that, that is cool and makes sense, but I don't think just changing it to `--config-dir` by default like this is the best idea, without an option to choose if that is the way it is done or not.

iynaix:
I've added a `scriptFiles` option that will use `--config-dir` if set, but falls back to the default behavior of reading from `$XDG_CONFIG_DIR/mpv` if unset and so will not affect current users.

BirdeeHub:
I would really rather what has been shown be done with `--script` and `--script-opts`, because it can be

But I recognize the need/desire for the ability to set stuff in the actual `--config-dir`, and also exclude the normal config.

Give me some time, I think we can do both in a satisfying way

https://mpv.io/manual/master/#mp-options-functions

They have a very well defined way to pass these things, determine the names of the plugins, etc.

iynaix:
I see where you are coming from, but that wouldn't work for me personally. I use https://github.com/CogentRedTester/mpv-sub-select that reads from a json file in the script-opts directory.

BirdeeHub:
Right. That was my conclusion from further research, is that, there is some things that definitely need a directory, but it is nice to let people use the normal ~/.config directory still if they want, alongside what we provide as well.

There should be both.

I'm gonna do both.

They let us do both, and clearly put a lot of effort into letting us do so. So, may as well.

BirdeeHub:
I have the --scripts and --script-opts part figured out in a local attempt I think, now I need to add something for the dir and make it optional

BirdeeHub:
since the other one got merged, Im going to close this one now?

PR #405, opened 2026-04-01
WIP: feat(core.install): integration module options

BirdeeHub:
Work in progress, but, the start of addressing:
https://github.com/BirdeeHub/nix-wrapper-modules/discussions/370

### DIFF
--- a/lib/core.nix
+++ b/lib/core.nix
@@ -4,6 +4,8 @@
   lib,
   wlib,
   extendModules,
+  moduleType,
+  _prefix ? [ ],
   # NOTE: makes sure builderFunction gets name from _module.args
   name ? null,
   ...
@@ -172,7 +174,192 @@ in
   '';
   config.meta.maintainers = [ wlib.maintainers.birdee ];
   config._module.args.pkgs = config.pkgs;
+  config.install.modules = {
+    homeManager =
+      { config, ... }:
+      {
+        config.home.packages =
+          let
+            cfg = args.config.install.getWrapperConfig config;
+          in
+          lib.mkIf cfg.enable [ cfg.wrapper ];
+      };
+    nixos =
+      { config, ... }:
+      {
+        config.environment.systemPackages =
+          let
+            cfg = args.config.install.getWrapperConfig config;
+          in
+          lib.mkIf cfg.enable [ cfg.wrapper ];
+      };
+    darwin =
+      { config, ... }:
+      {
+        config.environment.systemPackages =
+          let
+            cfg = args.config.install.getWrapperConfig config;
+          in
+          lib.mkIf cfg.enable [ cfg.wrapper ];
+      };
+  };
   options = {
+    install = lib.mkOption {
+      description = ''
+        This submodule contains options which create a generic integration module.
+
+        You can then import this submodule as a module in the module systems you specify in `config.install.modules`
+
+        i.e. in nixos or home manager `{ imports = [ inputs.nix-wrapper-modules.wrappers.<name>.install ]; }`
+
+        home manager, nixos, and nix darwin will by default also install the package.
+
+        It will create the options at `config.install.optionLocation`
+
+        This will default to `[ "wrappers" (lib.last _prefix) ]`,
+        so before importing it like this you will want to either set `config.install.optionLocation` yourself,
+        or use `wlib.getInstallModule` to set prefix before importing the module.
+
+        In this way, your wrapper module can specify installation instructions for any module system evaluation which sets the `class` attribute
+      '';
+      type = lib.types.submodule (
+        { config, ... }:
+        {
+          _file = wlib.core;
+          options = {
+            optionLocation = lib.mkOption {
+              type = lib.types.listOf lib.types.str;
+              default = lib.optionals (_prefix != [ ]) [
+                "wrappers"
+                (lib.last _prefix)
+              ];
+              description = ''
+                The location for the generated submodule option containing the wrapper module
+
+                This will be set for you if you use the provided flake-parts module or `wlib.getInstallModule`, otherwise you must set it manually.
+
+                This is to prevent the option location depending on the `pkgs` of the target evaluation, as it would if we used `config.binName`.
+              '';
+            };
+            mkWrapperExtension = lib.mkOption {
+              type = lib.types.functionTo lib.types.raw;
+              readOnly = true;
+              default =
+                key: module:
+                lib.setAttrByPath config.optionLocation {
+                  imports = lib.toList module;
+                  key = builtins.unsafeDiscardStringContext "${wlib.core}:${
+                    if lib.isStringLike module then "${module}." else ""
+                  }install.mkWrapperExtension:${toString key}";
+                };
+              description = ''
+                Function that takes a wrapper module to place at the path indicated by `config.install.optionLocation`.
+
+                It then returns the set to pass to config within an `install.modules` option
+
+                To provide this alongside the config for the target module system, you can use `lib.mkMerge`
+
+                ```nix
+                { ... }@top: { # <- in a wrapper module
+                  config.install.modules.nixos = { config, pkgs, lib, ... }: {
+                    config = lib.mkMerge [
+                      # pass another wrapper module
+                      # but available for only this module system (nixos in this example)
+                      (top.config.install.mkWrapperExtension "unique key name" (
+                        { wlib, pkgs, lib, config, ... }: {
+                          options.enableService = lib.mkEnableOption "installation of service units";
+                        }
+                      ))
+                      # config for nixos module (in this example)
+                      {
+                        systemd.packages = let
+                          # get the config from the wrapper modules option namespace
+                          cfg = top.config.install.getWrapperConfig config;
+                        in lib.mkIf cfg.enableService [ cfg.wrapper ];
+                      }
+                    ];
+                  };
+                }
+                ```
+
+                Tip: Instead of using `mkWrapperExtension`, you might wish to consider simply declaring the option in the wrapper module,
+                and just checking its value when relevant in your `install.modules` modules
+
+                This function is generally most useful when the option declaration itself must know something about the target module system.
+              '';
+            };
+            getWrapperConfig = lib.mkOption {
+              type = lib.types.functionTo lib.types.raw;
+              default =
+                let
+                  res = lib.attrByPath config.optionLocation null;
+                in
+                if res == null then
+                  throw "could not find the wrapper config! `getWrapperConfig` works only in the modules exported by `<this_wrapper_module>.install` and the configurations which import it!"
+                else
+                  res;
+              readOnly = true;
+              description = ''
+                This option indexes into config and returns the wrapper submodule at `config.install.optionLocation` within it
+
+                ```nix
+                { ... }@top: { # <- in a wrapper module
+                  config.install.modules.nixos = { config, lib, ... }: let
+                    wrappercfg = top.config.install.getWrapperConfig config;
+                  in {
+                    config = lib.mkIf wrappercfg.enable {
+                      # ...
+                    };
+                  };
+                }
+                ```
+              '';
+            };
+            modules = lib.mkOption {
+              type = lib.types.lazyAttrsOf lib.types.deferredModule;
+              # don't set default, instead set via config, so that you can add some default behavior to some of them
+              description = ''
+                A set of modules for integrating the wrapper module with an external module system. Place a module at the attribute name corresponding with the class of the target module evaluation
+              '';
+            };
+            __functor = lib.mkOption {
+              type = lib.types.raw;
+              internal = true;
+              readOnly = true;
+              default =
+                _:
+                {
+                  _class,
+                  pkgs ? null,
+                  ...
+                }:
+                {
+                  _file = wlib.core;
+                  imports = lib.toList (config.modules.${_class} or [ ]);
+                  options =
+                    assert
+                      config.optionLocation != [ ]
+                      || throw "to import `<wrapper_module>.install`, you must set `config.install.optionLocation` to a non-empty list of strings!";
+                    lib.setAttrByPath config.optionLocation (
+                      lib.mkOption {
+                        type = moduleType;
+                        default = { };
+                      }
+                    );
+                  config = config.mkWrapperExtension "wrappers-install-add-enable-option" {
+                    _file = wlib.core;
+                    config._module.args.name = lib.mkOverride (lib.modules.defaultOverridePriority - 1) (
+                      lib.last config.optionLocation
+                    );
+                    options.enable = lib.mkEnableOption "the wrapper module.";
+                    config.pkgs = lib.mkIf (pkgs != null) pkgs;
+                  };
+                };
+            };
+          };
+        }
+      );
+    };
     meta = {
       maintainers = lib.mkOption {
         description = "Maintainers of this module.";
--- a/ci/docs/md/getting-started.md
+++ b/ci/docs/md/getting-started.md
@@ -215,7 +215,7 @@ With a single, simple function, you can use any wrapper module directly as a mod
 # in a nixos module
 { ... }: {
   imports = [
-    (inputs.wrappers.lib.mkInstallModule { name = "tmux"; value = inputs.wrappers.lib.wrapperModules.tmux; })
+    (inputs.wrappers.lib.getInstallModule { name = "tmux"; value = inputs.wrappers.lib.wrapperModules.tmux; })
   ];
   wrappers.tmux = {
     enable = true;
@@ -230,8 +230,7 @@ With a single, simple function, you can use any wrapper module directly as a mod
 # in a home-manager module
 { config, lib, ... }: {
   imports = [
-    (inputs.wrappers.lib.mkInstallModule {
-      loc = [ "home" "packages" ];
+    (inputs.wrappers.lib.getInstallModule {
       name = "neovim";
       value = inputs.wrappers.lib.wrapperModules.neovim;
     })
@@ -261,7 +260,7 @@ With a single, simple function, you can use any wrapper module directly as a mod
 }
 ```
 
-See the [`wlib.mkInstallModule`](../lib/wlib.html#function-library-wlib.mkInstallModule) documentation for more info!
+See the [`wlib.getInstallModule`](../lib/wlib.html#function-library-wlib.getInstallModule) documentation for more info!
 
 ### `flake-parts`
 
@@ -294,32 +293,51 @@ It offers a template! `nix flake init -t github:BirdeeHub/nix-wrapper-modules#fl
       # Import the flake-parts module:
       imports = [ wrappers.flakeModules.wrappers ];
 
-      perSystem =
-        { pkgs, ... }:
-        {
-          # wrappers.pkgs = pkgs; # choose a different `pkgs`
-          wrappers.control_type = "exclude"; # | "build" (default: "exclude")
-          wrappers.packages = {
-            alacritty = true; # <- set to true to exclude from being built into `packages.*.*` flake output
-          };
-        };
+      # provide wrapper modules to flake.wrappers
       flake.wrappers.alacritty = { pkgs, wlib, ... }: {
         imports = [ wlib.wrapperModules.alacritty ];
         settings.terminal.shell.program = "${pkgs.zsh}/bin/zsh";
         settings.terminal.shell.args = [ "-l" ];
       };
+      flake.wrappers.xplr = wrappers.lib.wrapperModules.xplr;
       flake.wrappers.tmux =
         { wlib, pkgs, ... }:
         {
           imports = [ wlib.wrapperModules.tmux ];
           plugins = with pkgs.tmuxPlugins; [ onedark-theme ];
         };
-      flake.wrappers.xplr = wrappers.lib.wrapperModules.xplr;
+
+      flake.wrappers.tmux-modified = {
+        # using flake.wrappers will also make importable forms
+        # available in config.flake.wrapperModules!
+        imports = [ self.wrapperModules.tmux ];
+        # these will add to the above config which added the onedark-theme plugin
+        modeKeys = "vi";
+        statusKeys = "vi";
+        vimVisualKeys = true;
+      };
+
+      # no need for getInstallModule with flake-parts!
+      flake.nixosModules = builtins.mapAttrs (_: v: v.install) self.wrappers;
+      flake.homeModules = self.nixosModules;
+      # you don't have to export them from there specifically,
+      # this just shows that you can access `.install` directly when using the flake-parts module
+
+      # (optionally) Control which packages get built!
+      perSystem =
+        { pkgs, ... }:
+        {
+          # wrappers.pkgs = pkgs; # (optionally) choose a different `pkgs`
+          wrappers.control_type = "exclude"; # | "build" (default: "exclude")
+          wrappers.packages = {
+            tmux-modified = true; # <- set to true to exclude from being built into `packages.*.*` flake output
+          };
+        };
     };
 }
 ```
 
-The above flake will export the partially evaluated submodule from `outputs.wrappers` as it shows.
+The above flake will export the partially evaluated submodules from `outputs.wrappers` as it shows.
 
 However, it also offers the values in importable form from `outputs.wrapperModules` for you!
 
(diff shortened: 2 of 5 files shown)


zenoli:
In case you want some feedback on the usability, I'm willing to give this a try on the weekend for my two usecases: integrating zsh into nixos and integrating direnv into zsh (these would both be solvable by this, right?)

BirdeeHub:
yeah I actually have the zsh one in this PR as an example!

Do let me know what you think, I am not entirely set on the design and I am still deciding exactly how it should be, but hopefully this illustrates at least what I was talking about in that discussion.

BirdeeHub:
I also broke it apparently

BirdeeHub:
OK I thought I broke it last night, turns out I was just importing it twice.

I still have to test out some things though to see if this is viable.

I am concerned about if importing the module in a home manager config, and then grabbing that module out of that home manager config and using the install module from that in another config, if that will cause any options added to the wrapper module by the install module to be declared multiple times or not.

So, there is more testing to do and stuff of niche scenarios like that.

Do let me know how it goes.

I have in my config right now a branch with this new version where I am messing with it

https://github.com/BirdeeHub/birdeeSystems/compare/main...installmods

Edit: It does appear to be an issue to import it after importing it........

BirdeeHub:
https://github.com/NixOS/nixpkgs/blob/306d3438ad768b4056c4f1a21a53742695016649/lib/types.nix#L1245

^ this module does not have key set in it

https://github.com/NixOS/nixpkgs/blob/306d3438ad768b4056c4f1a21a53742695016649/lib/types.nix#L1302

^ and neither does this one.

This is actually is somewhat of a problem for the case when you try to grab the wrapper module out of a home manager or nixos config which imports it and then use the install module out of that and try to use that in another home manager or nixos config.

So, I suppose I will be making a nixpkgs PR, because those should probably have key set, the internal module in lib.evalModules has it set, these should have it set too.

https://github.com/NixOS/nixpkgs/pull/506165

BirdeeHub:
To be honest, I am still not 100% sold on this design, although it is very nice actually in terms of what it allows a wrapper module to specify about how it is installed. It requires lib.evalModules to be called to grab the module for the other system though... it would be nice to see how much of that I am able to confine into the submodule level and leave the top level of the generated module as lazy as possible. That being said, as-is, it is still quite lazy, at least none of these are attrsOf, it doesnt depend on pkgs, and there is no interdependence that would be relevant for that. And also, we are literally trying to define modules for another module system inside our module, so I really am not sure how possible it would be to get around that but I will try? Basically I would need to write a really clever `wlib.installModule` to get around that one, and it would still have to do it at one point or another in the process, and I think what I have added to the core module might still be good to have or be helpful in that process even with that.

But regardless, that PR to nixpkgs is still something which should be done

zenoli:
I tested this today.
Her is my feedback purely based on UX without knowing any of the technical limitations.

### TL;DR:
For users I think this is great!
For wrapper module authors I would whish for the process of writing "install" modules to become a bit more streamlined.

---

### My current setup:

I have a repo `my-zsh` where I add options and my custom config to the zsh wrapper (`wlib.wrapperModules.zsh`).
So far I used `mkInstallModule` to expose a `flake.homeModules.zsh` and `flake.nixosModules.zsh`.
Then in my nixos config (another repo) I imported these modules, which gave me `wrapper.enable` options.
I then had to  configure zsh as my user shell in my nixos config

```nix
    users.users.<name>.shell = lib.getExe config.wrappers.zsh.wrapper;
```

### With `install.modules`

From a user's perspective, I think the `install` option is a great improvement.

In my nixos config I simply need to import `inputs.zsh-flake.wrappers.zsh.install` in nixos and enable it in my config:

```nix
wrappers.zsh = {
    enable = true;
    asSystemDefault = true;
};
```

No more explicitly using `mkInstallModule`. 
Also, it is opt-in. I can still hook it up how I want by not setting `asSystemDefault`.

For wrapper authors, I think it is still a bit boilerplate-heavy at the moment.
I added an option to the nixos module to only make zsh the default shell for a given user:

```nix
      install.modules.nixos = { config, lib, ... }: 
      let
        cfg = top.config.install.getWrapperConfig config;
      in
      {
        config = lib.mkMerge [
          (top.config.install.addWrapperModule "${./default.nix} zsh as userShell" {  # <-- no idea how to pick the key
            _file = ./default.nix;
            options.userShell = lib.mkOption {
              type = lib.types.nullOr lib.types.str;
              description = "zsh as userShell";
              default = null;
            };
          })
          (lib.mkIf cfg.enable {
            users.users = lib.mkIf (cfg.userShell != null) {
              "${cfg.userShell}".shell = lib.getExe cfg.wrapper;
            };
            programs.zsh.enable = lib.mkIf (cfg.asSystemDefault || cfg.userShell != null) true;  # <-- not sure how this is merged with the module defined in this PR
          })
        ];
      };

```


Ideally, from the use cases I experienced so far, I would want for most of this boilerplate to be abstracted a way from me.
I do not understand the implications of picking a key for `addWrapperModule` and why I need `mkMerge` and pass it the option and config snippets as two list items. I'm sure it due to some nixos module system behavior I do not understand, but I'd hope that it is something I don't have to care for when I write wrappers using this feature.

I was expecting to just have to provide a module with options and configs that will be embedded into my nixos (home manager) config once installed.

Something like:

```nix
install.modules.nixos = { config, lib, ... }: {
    options.foo = mkOption ...
    config.programs.zsh.enable = lib.mkIf ...
}
```

...and all the things necessary to make it work (mkMerge, setting unique keys etc.) to be magically taken care of for me :-)

BirdeeHub:
```nix
install.modules.nixos = { config, lib, ... }: {
    options.foo = mkOption ...
    config.programs.zsh.enable = lib.mkIf ...
}
```

^ I mean, you can do that.

But then foo would just be set like `config.foo = "idk"` in nixos, you would need to pick some kind of namespace for the options that wrapper module creates in nixos.

`addWrapperModule` adds another module to the wrapper submodule option itself, via nixos or home manager.

The option is at `config.install.optionLocation`, `addWrapperModule` puts it there for you.

As for why `key` is needed, that is for deduplication...

If you do some really cursed stuff, like grabbing the wrappers from another evaluated system config and then using the install modules from those in another config....

```nix
imports = lib.mapAttrsToList (n: v: v.install) self.nixosConfigurations.mysystemconfig.config.wrappers;
```

Well, then the nixos module gets added twice, which adds its option to the wrapper module twice. So we force you put literally anything unique there as a key, so that it deduplicates it in that scenario. (however currently that is not all that is needed, as lib.types.submodule sets _module.args.name twice in that scenario, hence the PR to nixpkgs)

zenoli:
> ^ I mean, you can do that.
> 
> But then foo would just be set like `config.foo = "idk"` in nixos, you would need to pick some kind of namespace for the options that wrapper module creates in nixos.
> 
> `addWrapperModule` adds another module to the wrapper submodule option itself, via nixos or home manager.
> 
> The option is at `config.install.optionLocation`, `addWrapperModule` puts it there for you.

Soo.... 

in the example of integrating zsh into nixos...
...I could also write 

```nix
options.wrappers.zsh.foo = mkOption ...
```

but
 
```nix
addWrapperModule "some-unique-key" ({ wlib, ... }: { 
    options.foo = mkOption ....
})
```

already takes care to add the option alongside the other zsh wrapper options.

And then the "native" nixos modules are specified in a separate module and everything is merged together using `lib.mkMerge`. Is this all there is to it?

BirdeeHub:
Yeah pretty much.

this

```nix
addWrapperModule "some-unique-key" ({ wlib, ... }: { 
    options.foo = mkOption ....
})
```

returns (assuming `config.install.optionLocation` was `[ "wrappers" "zsh" ]`)

```nix
{ 
  wrappers.zsh = {
    key = builtins.unsafeDiscardStringContext "${wlib.core}:install.addWrapperModule:some-unique-key";
    imports = [ themoduleyoupass ];
  };
}
```

You want to supply that to `config = that`.

There are of course, several ways to merge that with your other values in config, but mkMerge is the most general, easy recommendation.

Im still not entirely decided on this design, as there are still edgecases to handle (or, decide if they are able to be handled and what to do about that)

but that is how they are currently working.

zenoli:
Got it. 
It makes sense now.

The only thing that still strikes me as "weird" is that we have to provide those unique keys. I'm not familiar enough with the internals of the module system to be able to propose a solution, but it "feels" like if the problem is just to ensure a unique key,  there should be a way for `addWrapperModule` to generate it internally and omit the `key` argument. (I guess this is non-trivial and there is a reason the `key` arg has to exist. I just don't know why)

BirdeeHub:
~~Well..... maybe? with `builtins.unsafeGetAttrPos`? never used it before for something real, unsure if it would work properly for this. I feel like we would then run into scenarios where a function generates several modules, and then they share a key or something.~~

~~Although, I suppose then they could supply their own key in the module? Rather than it being forced? The problem is, reaching that edgecase would be a REALLY non-trivial error case. Like, the most non-trivial, with no helpful error message whatsoever, just stuff not being imported for some reason you don't understand.~~

---

But the problem that makes it non-trivial to generate this value is this

You (might in some scenario related to this, outlined above) pass the same module twice. To the same function twice.

How does it know they are the same module, and not different ones? Because if it just adds a unique key, then, it would add a different key to both copies of the same module, and then they would not deduplicate still.

It literally does not matter what the value of key is, as long as there is not another module you want to keep imported in that particular wrapper module with that same key, but the ones you do want to de-duplicate have the same key

As for it being weird...

the situation it solves is weird too...

It solves a really niche problem...

I am not the happiest about it being there either

Its a good part of why this is still a draft

I feel like there should be some way around most of these problems, that I have not seen yet. Some slightly different design that solves these problems in a more elegant way. I haven't come up with it yet though. To be honest, I haven't even fully solved them in an inelegant way yet.

lib/core.nix
```diff
@@ -172,7 +173,162 @@ in
...
+              type = lib.types.functionTo lib.types.raw;
+              readOnly = true;
+              default =
+                key: module:
+                lib.setAttrByPath config.optionLocation {
+                  imports = lib.toList module;
+                  key = builtins.unsafeDiscardStringContext "${wlib.core}:${
+                    if lib.isStringLike module then "${module}." else ""
+                  }install.addWrapperModule:${toString key}";
+                };
```
zenoli, line 254: So this creates a set:

```nix
wrappers.<name> = {
  imports = [module];
  key = ...
}
```

which gets then assigned to `config` in the `__functor` below.

Do `imports` get resolved also when being set at arbitrary positions in the config tree?
I thought they must be top level?

```nix
{
  imports = ...
  options = ...
  config = ...
}
```

BirdeeHub: submodules are modules. You can import in them.

subWrapperModules are wrapper modules. Which are just modules with `wlib.core` imported and `wlib` passed as a specialArg, and `modulesPath` specialArg set to the nix-wrapper-modules root dir

`wlib.types.subWrapperModule` is a `lib.types.submoduleWith` alias, so you dont even have to pass them the function form of a module to make use of the whole module, because they have `shorthandOnlyDefinesConfig` defaulting to `false` unlike `lib.types.submodule` where it defaults to `true`

`moduleType` is also a submoduleWith type.

lib/core.nix
```diff
@@ -172,7 +173,162 @@ in
...
+              ];
+              description = ''
+                The location for the generated submodule option containing the wrapper module
+
+                This will be set for you if you use the provided flake-parts module or `wlib.mkInstallModule`, otherwise you must set it manually.
+
+                This is to prevent the option location depending on the `pkgs` of the target evaluation, as it would if we used `config.binName`.
+              '';
+            };
+            addWrapperModule = lib.mkOption {
```
zenoli: Maybe `wrapperModule` is the wrong term here because `wrapperModules` refer to the "wrapper modules" this repo provides (i.e. `wlib.wrappers.<wrapperModule>`). This module is a plain nixos/hm/etc module that integrates the wrapperModule, no?

Hence, I would use something else here.

- `integrationModule`
- `wrapperIntegrationModule`

Also, imo the prefix `add` is not necessary. The option does not "add" the module, it merely provided ist.
It gets "added" in the functor.

BirdeeHub: `modules.nixos` is a plain nixos module

`addWrapperModule` adds a module to the wrapper module which is currently imported inside a `submoduleWith` option in said nixos module

(but that module it adds to the wrapper module is not imported inside the wrapper module normally, only if you import the install module of that wrapper module in that particular system does it get imported in that particular submodule option containing the wrapper module)

BirdeeHub: I am of course happy to change the name still to something more instructive if you have one, but it is definitely adding a module to the wrapper module which is imported inside that nixos module as a submodule option, so it is adding a module to the wrapper module, or, adding a wrapper module. IDK. addWrapperModule was what I came up with, there is probably better, do let me know

wrapperIntegrationModule is ok. integrationWrapperModule? config.install.putitintherightplaceplease? (joking with that last one obviously) honestly add felt about as good, and was shorter? extraWrapperModule? extendSubWrapperModule? `config.install.extendWrapper`?

BirdeeHub: AI SLOP FUNCTION NAME BRAINSTORM COURTESY OF GIPPITY BEING FED THIS CONVERSATION

* `extendWrapperModule` ← probably the cleanest
* `wrapperModuleExtension`
* `wrapperModuleExtender`
* `wrapperModuleAugment`
* `augmentWrapperModule`
* `wrapperModulePatch`
* `patchWrapperModule`

### If you want to emphasize composition / merging

* `composeWrapperModule`
* `mergeIntoWrapperModule`
* `wrapperModuleOverlay`
* `overlayWrapperModule` (Nix-y, but slightly overloaded with overlays conceptually)

### If you want to emphasize scope (submodule context)

Since this happens inside a `submoduleWith`, you can lean into that:

* `extendWrapperSubmodule`
* `wrapperSubmoduleExtension`
* `wrapperSubmoduleModule` (a bit clunky but precise)
* `injectWrapperSubmodule`

### If you want to emphasize “hooking into install flow”

Because it only applies when used via the install module:

* `wrapperInstallExtension`
* `installWrapperExtension`
* `wrapperIntegrationExtension`
* `extendWrapperIntegration`

### Short + pragmatic (closest to what you had)

* `extendWrapper` ← good balance of brevity + meaning
* `extraWrapperModule`
* `wrapperExtension`

---

### What to avoid (based on your own confusion)

* `addWrapperModule` → implies side-effecting behavior, not declarative provision
* `integrationModule` → too generic, loses target
* `wrapperIntegrationModule` → still ambiguous direction
* anything with just `module` → doesn’t distinguish *which* module is affected

---

### Strong recommendation

If you want something that will still make sense in 6 months:

> **`extendWrapperModule`**

It encodes:

* target: `WrapperModule`
* action: `extend` (fits Nix mental model: overlays, mkMerge, etc.)
* avoids imperative “add”

If you want slightly shorter:

> **`wrapperModuleExtension`**

Same meaning, more data-oriented naming.

BirdeeHub: `wrapperSubmoduleModule` really seems like the winner here of course :rofl: :rofl: :rofl: :rofl: :rofl: :rofl:

(more seriously, `config.install.extendWrapperModule` or `config.install.extendWrapper`)

zenoli: > (more seriously, config.install.extendWrapperModule or config.install.extendWrapper)

Maybe I am missing something but to me it's unintuitive to have verb in the name because as far as I understand this option merely returns some module/submodule. It doesn't DO anything with it. Hence I would rather name it `wrapperModuleExtension/wrapperExtension` instead of `extendWrapperModule/extendWrapper` if you were to go for this one.

If this option was responsible for integrating it in the "consuming" module system (which I think is done in the `__functor` then `extendWrapperModule` or even `add...` would be adequate.

BirdeeHub: I see what you are saying actually. `config.install.wrapperExtension` seems ok to me actually

or maybe `mkWrapperExtention` because that is common in nixpkgs for like `mkIf` and stuff where it just returns a value with some specific type that if put into the module system will have some effect. Actually, I think I just talked myself out of mkWrapperExtension? unsure.

BirdeeHub: I still need to solve the _module.args.name problem though, and despite this repository being enjoyable to mess with I do have stuff to do, and I have some ideas I think but I havent gotten to trying them out, so there will be some time between now and when this is ready

I need to fix up the mpv module, Im most of the way done with that.

lib/core.nix
zenoli: hmm ok.. So in the functor you define an option at the same location (`wrappers.<name>`) of type `moduleType`:
https://github.com/BirdeeHub/nix-wrapper-modules/blob/3da8aa9a55743ac64a2288514200c9a223fe33c9/lib/core.nix#L317

I guess this is how you indicate that whatever is configured there is a module and hence the module system understands the `import = [...]` key there. Is this the idea?

`moduleType` is some kind lowever level detail of the module system? Couldn't find much about it.

lib/core.nix
zenoli: > I see what you are saying actually. `config.install.wrapperExtension` seems ok to me actually
> 
> or maybe `mkWrapperExtention` because that is common in nixpkgs for like `mkIf` and stuff where it just returns a value with some specific type that if put into the module system will have some effect. Actually, I think I just talked myself out of mkWrapperExtension? unsure.

I think using `mk...` makes sense (no pun intended).
I'm also in favor of having "module" somewhere in the name because it indicates the type of the option (`lib.types.raw` does not make this obvious for me).



I'm still debating whether `Integration` is better than `Extension` but in the end, probably both are fine for me.
`Extension` is  adequate as you are free to extend the wrapperModules in whatever way you want, although the main reason for having this option, is to add "integration capability" to the module. `mkIntegrationModule` would make this purpose more explicit but `mkWrapperExtension` (or `mkWrapperModuleExtension`) is correct too.

lib/core.nix
BirdeeHub: > moduleType is some kind lowever level detail of the module system? Couldn't find much about it.

It is a submoduleWith type with the current module evaluation's state in it

https://github.com/NixOS/nixpkgs/blob/d83c818aea5e339a71744e01e15fe54af93aa0da/lib/modules.nix#L395-L397

https://github.com/NixOS/nixpkgs/blob/d83c818aea5e339a71744e01e15fe54af93aa0da/lib/modules.nix#L250

https://github.com/NixOS/nixpkgs/blob/d83c818aea5e339a71744e01e15fe54af93aa0da/lib/modules.nix#L94

lib/core.nix
BirdeeHub: mkWrapperExtension is probably the one then

Unless I find (or am told of) a better design that doesnt have it at all or something

lib/core.nix
zenoli: I see. Thx

BirdeeHub:
So, I think I fixed the _module.args.name problem at least.

Interface is otherwise still the same.

Still trying to think of a better way, but at least it solves the problem it was trying to solve.

BirdeeHub:
> integrating direnv into zsh

Also, I somehow just read this from your very first comment correctly.

Unless you set `wlib.evalModules { class = "zsh"; modules = [ yourzshmodule ]; }`, and then imported the direnv module into it then, no? So, it wouldn't help you integrate wrapper module to wrapper module. It does still make the option though? But if class isnt set, it just makes the option.

That being said... _devenv_ actually accepts modules, right? So, if you were passing your zsh to the devenv module evaluation used by the direnv config, and it sets class, then you could declare an install module for devenv within your zsh module.

These install modules would be for anything which is another module system which does set `_class` such as nixos or home manager.

Wrapper modules do not set class because it doesnt make sense, for particularly that reason.

`_class` is meant to ensure that the module you import will be compatible with that system.

For wrapper modules, each one would need to have its own `_class = "zsh";` or `_class = "direnv";` or whatever. But the overall class of the evaluation is set at the evaluator and not the individual class. Setting `_class` inside the module only says, will this module be compatible with that evaluator. So theoretically, we should use `_class = "wrapper";` but that isn't true, not all wrapper modules are compatible with wrapper modules of other programs.

So, for integrating zsh into your _direnv wrapper module_, you would have to make a `subWrapperModule` option in the direnv wrapper module. Or the other way around. But for passing zsh into the _devenv_ module system, I think so?

zenoli:
Thanks for clearing things up.
When I tested it, I quickly noticed that the 'direnv --> zsh' use case is not suitable here, so I remained with my current solution based on subWrapperModule as you mentioned (which is quite nice).

My feedback, after I tested it, was solely for the nixos/hm integration

BirdeeHub:
So, I am not really thinking of anything actually better for `mkWrapperExtension`

The best I have is:

Instead of `mkWrapperExtension`

we have just a set of extra wrapper modules. But then... is that, like, do the names have to match up with install.modules then? Do we have to have it be like, a set of sets maybe so that they can be grouped with their correct install module and then also have a name? And then they can't access the config from that other module system in it...

So, that doesn't really feel better.

To be fair, most of the time, people will just add a new option to their wrapper module rather than messing around with this anyway :rofl:

I really wish I didn't have to have name as the first parameter, and have name not really matter whatsoever other than being different. It is somewhat confusing. But this makes it more technically sound and forms some kind of guarantee that this will be handled. Because if it isn't forced, nobody would do it, practically guaranteed. Theres like 20 people in the whole world who know how `lib.evalModules` works top to bottom. And I am not even quite one of them, despite knowing a lot about it...

I could maybe add a second helper that takes care of the `mkMerge` thing? But I am having trouble finding a satisfying way to do it that is any shorter, and it isnt necessary in every case to add an extra wrapper module anyway.

That being said, I still like what it enables, and I have been using it and it works nicely, and it allows for something like a dendritic config for that particular program without even using flake-parts for that, which is somewhat satisfying to use.

So, like, not fully decided, but it is growing on me more, especially now that I fixed the (admittedly somewhat niche, but still important) duplicate name issue

BirdeeHub:
I can't think of anything better...

zenoli:
I haven't been following the recent changes. I can try it again and provide feedback if you wish.

BirdeeHub:
I mean it hasn't really changed much in terms of usage. I just solved the last issue with reimporting imported things via submodules in cursed ways by using `__prefix` to set `_module.args.name` with a priority 1 higher than normal. I also rebased it onto main a bunch of times because I use this branch for my config and I have departed from my original branch on accident XD

I just still haven't though of a better way to handle key.

I have thought of several other ways.

But not a better one.

And I could just, not handle it. But then no one will handle it. And I will be annoyed. It is not that big of a requirement to name the thing.

And the feature is nice. I like it. I have been using it in my config and I like using it.

So, I will be giving this a once over again soon after I review the fish module again probably, and I guess just going ahead with it?

If you have more feedback, happy to hear it, but I think the names of at least the options in core are fine now, maybe I can improve on `wlib.installModule`? If you have any suggestions there, name or implementation, that would be good.

But unless you have an actionable better way to handle the name thing, I ran out of ideas and am happy with this approach for that lol

BirdeeHub:
Im going to wait until after the april 30th deprecations to merge this at least still though

BirdeeHub:
actually, I did find a problem with the implementation of `wlib.installModule`, and fixed it

Now it sets prefix, rather than setting optionLocation option.

This means if people set the option inside the module they won't get conflicts, and if they do set it, it will use their setting, because it only uses prefix because thats what the default checks

I also renamed it to `wlib.getInstallModule`

BirdeeHub:
Also, usage of this in flake parts is like

```nix
inputs: # <- just needed to import the flake parts modules
{ config, ... }: {
  imports = [
    inputs.flake-parts.flakeModules.modules
    inputs.wrappers.flakeModules.wrappers
  ];
  config.flake.modules = let
    installMods = builtins.mapAttrs (_: v: v.install) config.flake.wrappers;
  in {
    homeManager = installMods;
    nixos = installMods;
    darwin = installMods;
    generic = installMods;
  };
}
```

And then you have modules for the various things you can use exported from your flake

But without flake parts, if you had a set of unevaluated modules like in `wlib.wrapperModules`

```nix
installMods = builtins.mapAttrs (name: value: wlib.getInstallModule { inherit name value; }) wlib.wrapperModules;
```

and then you export/import those instead

BirdeeHub:
Also, I just realized I still have to do the flake outputs

I was exporting home and nixos modules via mkInstallModule and now I should use getInstallModule to export those

BirdeeHub:
oof and I forgot the docs

BirdeeHub:
Alright. I guess it is time to merge this thing.

PR #408, opened 2026-04-04
feat(wrapperModules.fish): init

ormoyo:
Add wrapperModule: fish


BirdeeHub:
Hi there!

So, first thing that comes to mind.

We ended up running into some issues when wrapping zsh that might be relevant here as well.

Basically, nixpkgs has an /etc/set-environment script

And home manager likely has things in its profile paths as well to source.

Both of those have the ability to overwrite things we provide via the wrapper.

https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295#discussioncomment-15814531

^ context for the issues we ended up running into.

Basically, we ended up generating the entire wrapper into the config file, so that we run after that gets sourced

https://github.com/BirdeeHub/nix-wrapper-modules/blob/708dd619b5096f93c55ccef25af9cae5e57d4190/wrapperModules/z/zsh/zshenv.nix

---

I am not sure if any of that is relevant for fish, but that would be the first things to test.

Depending on when [this](https://github.com/NixOS/nixpkgs/blob/6201e203d09599479a3b3450ed24fa81537ebc4e/pkgs/by-name/fi/fish/package.nix#L76-L78) gets ran, we may or may not run into the same issues.

---

These links may or may not be useful as well, they are the relevant PRs for when the zsh wrapper was added and also for a fix that was needed for it.

There may be some useful information in there, if you do end up finding that there are similar issues here as well

https://github.com/BirdeeHub/nix-wrapper-modules/pull/347

https://github.com/BirdeeHub/nix-wrapper-modules/pull/355

https://github.com/BirdeeHub/nix-wrapper-modules/pull/356

ormoyo:
After some testing, it seems a change to an environment variable in the wrapper is kept even if it has an override in /etc/set-environment.

As for https://github.com/BirdeeHub/nix-wrapper-modules/discussions/295#discussioncomment-15814531, config.extraPackages works completly fine.

And for [that](https://github.com/NixOS/nixpkgs/blob/6201e203d09599479a3b3450ed24fa81537ebc4e/pkgs/by-name/fi/fish/package.nix#L76-L78), the fish arg [--no-config](https://man.archlinux.org/man/fish.1#N) is set, so only the wrapper's configration is used (I think the same applies to the home-manager stuff).

BirdeeHub:
Ok I will check it out sometime this week on my system to make sure everything is good :)

Somewhat hesitant about `--no-config`, worried that will make it so that some things are not provided by nixos or home manager

ormoyo:
After reading https://github.com/BirdeeHub/nix-wrapper-modules/discussions/326  I have harnessed the understanding of spec and dag/dal. And will implement it in the soon to come commits. You should add the example you gave in https://github.com/BirdeeHub/nix-wrapper-modules/discussions/326#discussioncomment-16045310 to the docs. 

I had a hard time understanding the spec type before I read that discussion.

BirdeeHub:
I still have a few days until I can try this out, Im glad you were able to figure out how the spec type works!

One thing to keep in mind with them is that they are slightly harder to override than a normal submodule option when it comes to overriding fields that are not the main one, (although you still can, hence `wlib.ignoreSpecField`), so use the DAG type specifically when the value definitely has a single defined main field where the normalization is useful enough, or needs to match another option which is a DAL.

The DAL type, however, can generally be freely used instead of a list, because you already couldn't override a list like that.

I agree documentation on them could be improved. But also docgen could be improved too, and there are still several core pieces of the project to write, (which I now have added to the readme actually).

I have trouble picking which thing to work on, and unfortunately, improving written documentation is often not the one I pick.

The reason it is taking me a while to get around to trying this one out, is because the specific thing I want to try out is using it as the default system shell, and that is somewhat of a bigger ordeal than just building and running a program is, so I need some time I can set aside for that rather than just poking at it while I am doing other things, so, thank you for being patient.

ormoyo:
Good thing you mentioned default system shell, I tried that myself, but I didn't add the attribute "shellPath" to the derivation

BirdeeHub:
```nix
{ pkgs, wlib, lib, ... }: {
  imports = [ wlib.wrapperModules.fish ];
  extraPackages = [ pkgs.hello ];
}
```

`fish`

hello found

`unset __NIXOS_SET_ENVIRONMENT_DONE && fish`

no hello

This is because nixpkgs adds something to fish config that runs after our wrapper script but before our added cli arguments which sources the nixos `config.system.build.setEnvironment` at `/etc/set-environment` and that SETS the path and a few other variables rather than extending them.

In circumstances where fish is the first shell started, it is likely that __NIXOS_SET_ENVIRONMENT_DONE is not yet set, which would then cause it to source that file, and remove our changes to variables that ran within the wrapper script itself which are also set by that script, namely, PATH.

This is a thing only done in nixpkgs for shells for the most part.

Ill get back to you soon with details on how to get around that, but basically this means we are going to have to do the thing we did for zsh where we generate the main wrapper content into the config file and not a wrapper script...

We also need to be using makeBinaryWrapper in some way because scripts cant be used as shebangs on mac.

So, all of that would be achieved the way we did in the zsh wrapper, there we generated it all into zshenv, which runs after the set-environment script.

https://github.com/BirdeeHub/nix-wrapper-modules/blob/fb62851ffc5f6a4d53ebc00b93743e29b41e6224/wrapperModules/z/zsh/zshenv.nix#L54-L210

And then, because we would only want to repeat that logic for the main fish binary, we would fix the wrapperVariants options

https://github.com/BirdeeHub/nix-wrapper-modules/blob/fb62851ffc5f6a4d53ebc00b93743e29b41e6224/wrapperModules/z/zsh/module.nix#L91-L97

https://github.com/BirdeeHub/nix-wrapper-modules/blob/fb62851ffc5f6a4d53ebc00b93743e29b41e6224/wrapperModules/z/zsh/variants.nix

Sorry for wrapping shells being a bit brutal, nix primarily does its work by modifying the environment of processes, and shells are a main way it does that.

---

I also have some suggestions for the design around `configFiles` and `functionFiles` options.

namely, why are they separate if they do the same thing? and also, neither of them can access placeholders by default.

People likely just want a way to pass it config content and source other files if they want to, maybe something simpler like this would be better? With the bonus that it also would allow use of placeholders in said files so that if they had multiple config files they could refer to each other arbitrarily?

```nix
options.configFile = mkOption {
  type = wlib.types.file pkgs;
  default.content = "";
  default.path = config.constructFiles.configFile.path;
  description = "Configuration file to be sourced by the shell";
};
config.constructFiles.configFile = {
  relPath = "??";
  content = config.configFile.content;
};
```

Apologies for wlib.types.file being misleading, I honestly don't like it when used with default options because pkgs.writeText makes the placeholders in it point to the drv created by that pkgs.writeText call and not the main wrapper derivation, but I could understand it being helpful for people quickly doing things in their config. Honestly it is just there because I started with lassulus wrappers as a starting point, and its one of the few things I didn't change much, and honestly, regretting my decision to not just remove it to begin with.

But, since it is there, if I do a new version, the new version will just be the same thing but without the default of pkgs.writeText, and then it wont need you to pass it a pkgs. And I don't think I will delete the old version, I will just tell people to prefer not to use it in wrappers submitted to the main repo. Or maybe I can find an even cleaner way to do it idk.

I also think what you did for completionFiles is also fine, those are so often provided by other sources, that it is fine to use wlib.types.file for those, they still are able to generate files for those with constructFiles and pass placeholders there, but at the very least the main config file should support them by default

ormoyo:
Ok, so I basically copied the zsh implementation for circumventing the nixpkgs overwrites and changed the zsh syntax to fish syntax.  

It keeps all values even with `unset __NIXOS_SET_ENVIRONMENT_DONE`. 
Feel free to point things out.

And as a matter of fact, the `functionFiles` option was supposed to use fish's [autoloading functions](https://fishshell.com/docs/current/language.html#autoloading-functions) (I just forgot) but honestly, it's good enough to just let the user define a function in `configFile`.

BirdeeHub:
It looks correct, but there is obviously a lot more that has been done in this version, so give me a little bit of time :)

Thank you for implementing this, it will be a great addition.

adrian-gierakowski:
Would be great to have this.
I came here after discovering that there is no fish module in https://github.com/Lassulus/wrappers :)

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+            functor,
+            multiple ? true,
+          }:
+          let
+            pluginLines =
+              if (builtins.isFunction dirList) then
+                map (plugin: map functor (dirList plugin)) plugins
+              else
+                map functor dirList;
+            toString = plugins: builtins.toString (map (p: p.src) plugins);
```
BirdeeHub:
```suggestion
            pluginsToString = plugins: toString (map (p: p.src) plugins);
```

Lets not shadow `toString`

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+          in
+          optionalString (plugins != [ ]) ''
+            set plugin${optionalString multiple "_list"} ${toString plugins}
```
BirdeeHub:
```suggestion
            set plugin${optionalString multiple "_list"} ${pluginsToString plugins}
```

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+
+  config.constructFiles.generatedConfig = {
+    relPath = "${config.binName}-config.fish";
+    builder =
+      let
+        startSection = ''
+          if set -q __wrapped_fish_sourced
+            return
+          end
+          set -gx __wrapped_fish_sourced 1
```
BirdeeHub, line 237:
```suggestion
          set -gx __wrapped_fish_sourced 1
          fish_add_path --path ${dirOf config.wrapperPaths.placeholder}
```

This solved some issues for me. I am not sure why, but it seems like a good idea, for when doing nix run for things.

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+      }
+    )
+  ];
+  options = {
+    configFile = mkOption {
+      type = wlib.types.file pkgs;
+      default = {
+        content = "";
+        path = config.constructFiles.generatedConfig.path;
+      };
```
BirdeeHub:
```suggestion
      type = wlib.types.file {
        path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
      };
```

We get to clean this up a little now. (If this doesn't work for you, rebase)

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+        default = cfg.pluginCompletionDirs;
+        description = "The directories which will be checked for config files";
+      };
+    };
+  };
+in
+{
+  imports = [
+    wlib.modules.symlinkScript
+    wlib.modules.constructFiles
```
BirdeeHub, line 111:
```suggestion
    wlib.modules.constructFiles
    ../../z/zsh/variants.nix
```

^ cursed, but in the interest of demonstration

The suggestion is just a demonstration, but please copy that file into this directory instead and include it.

But this would change the purpose of the wrapperVariants options from wrapping multiple values in the fish derivation, into options for wrapping extra packages in context of the fish derivation.

It also adds back the wrapperImplementation option to the wrapperVariants options.

Do you think this is a good idea? The variants are currently being handled by the normal wrapper function, and there is only fish we don't have to wrap other stuff.

BirdeeHub (commented):
Just a few fairly simple things left! (well, one isnt simple, but it is simple to do at least)

Otherwise, it seems to work. It even avoids duplication when I call it multiple times as far as I can tell, which is something we tripped over with zsh

wrapperModules/f/fish/module.nix
```diff
@@ -0,0 +1,538 @@
...
+        {
+          options.command = mkOption {
+            type = types.str;
+            default = name;
+            description = "The command to apply the completion for";
+          };
+        }
+      )
+    ];
+  };
```
BirdeeHub:
```suggestion
  completionType = wlib.types.file (
    { name, config, ... }:
    {
      config.path = lib.mkOptionDefault pkgs.writeText name config.content;
      options.command = mkOption {
        type = types.str;
        default = name;
        description = "The command to apply the completion for";
      };
    }
  );
```

We can do this more simply now. Some updates were done to `wlib.types.file`

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/408#issuecomment-4326218013

> Would be great to have this.
> I came here after discovering that there is no fish module in https://github.com/Lassulus/wrappers :)

I agree it will be a great addition.

There is a lot more here to check out in the meantime though!

Plenty of other wrapper modules of course, but the options for wrapping stuff yourself should be more powerful while still feeling similar to use, and the ability to use these wrapper modules directly in other module options as submodules is quite nice. We also have a flake-parts module you can import and then use other wrapper modules in that handles basically all the boilerplate of calling `evalModule` and exporting the package from `packages.*.*` for you.

wrapperModules/f/fish/module.nix
ormoyo: I guess better not to confuse the masses

wrapperModules/f/fish/module.nix
ormoyo: I don't think changing the convention you established for wrapperVarients is a good idea, it just increases the confusion. 

Maybe show me an example of the usefulness of this.

BirdeeHub:
```nix
config.wrapperVariants.wezterm = {
  flags."-e" = "${config.wrapperPaths.placeholder}"; #<- it can see the fish wrapper drv
  # package = pkgs.wezterm; # <- we don't need this because of the addition, it can grab based on name by default. we still could though, we could set it to an already wrapped wezterm if we want, the wezterm wrapper doesnt use -e and -e will win.
  wrapperImplementation = "binary"; # <- we added this back for the variants, since they are not fish and do not use our modified wrapper.
  # mirror = false; #<- things that are not also fish do not need --init-command added to their args by default
};
```

wrapperModules/f/fish/module.nix
ormoyo: Great, that also gives me the opportunity to convert completionType to completionModule. So it'll be the same convention as pluginModule and abbreviationModule.

wrapperModules/f/fish/module.nix
BirdeeHub: You should at least add wrapperImplementation back (for just the variants). The rest is optional. I should really have a better way of doing that (when I do, I will come update it here because I will do it in all the places it was done)

ormoyo: I think this will do. 
I guess this will be like a shell wrapper exclusive.

BirdeeHub: pretty much yeah.

shell wrappers are... interesting.

Basically, they NEED to use the binary implementation because otherwise people try to use them in shebangs on mac and are sad, but then also nixpkgs patches them, and that means you need to custom case the main wrapper, but can't really apply it as generically to any other program.

So the best solution I have found so far is to make the wrapperVariants options do something very slightly differently by default in those cases.

Soon, I should add a way to disable wrapperImplementation and such for the main level without disabling for wrapperVariants, so that we don't have to copy the whole definition still and I will come do that here when I do that.

BirdeeHub:
I think that is good?

Thank you for sticking with it.

Hopefully we don't find all the problems with it immediately despite all the testing and revision, but it seems pretty good to me!

ormoyo:
Very good, thank you for all the help and explainations along the way

BirdeeHub:
You're welcome! (Im at least partially responsible for the struggle, nixpkgs is the rest, may as well help haha)

PR #419, opened 2026-04-06
feat(wrapperModules.hyfetch): init

ricardomaps:


wrapperModules/h/hyfetch/module.nix
```diff
@@ -0,0 +1,39 @@
...
+      example = lib.literalExpression ''
+        {
+          preset = "rainbow";
+          mode = "rgb";
+          color_align = {
+            mode = "horizontal";
+          };
+        }
+      '';
+    };
```
BirdeeHub, line 27:
```suggestion
    };
    configFile = lib.mkOption {
      type = wlib.types.stringable;
      default = config.constructFiles.generatedConfig.path;
      description = ''
        The path to the config file. Can be anywhere.
        
         By default points to `config.constructFiles.generatedConfig.path`, which contains the generated result of `config.settings`
      '';
    };
```

wrapperModules/h/hyfetch/module.nix
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  config = {
+    package = lib.mkDefault pkgs.hyfetch;
+    constructFiles.generatedConfig = {
+      relPath = "${config.binName}-settings.json";
+      content = builtins.toJSON config.settings;
+    };
+    flags."--config-file" = config.constructFiles.generatedConfig.path;
```
BirdeeHub:
```suggestion
    flags."--config-file" = config.configFile;
```

BirdeeHub (commented):
It looks like hyfetch can write to its own config file.

We may want to give an explicit path option that allows them to point it out of the store at first if they wish.

We dont need to allow them to pass raw json file content, they can `builtins.fromJSON`, json doesnt append well anyway, so just the path is fine

wrapperModules/h/hyfetch/module.nix
ricardomaps: I pushed the changes. That's actually pretty cool, but it only works if i pass an absolute path as a string, is that right?

BirdeeHub: You could

```nix
configFile = "~/wherever";
```

Or if you wanted to take it a bit further:

```nix
flags."--config-file" = _: { esc-fn = x: x; };
```

And then

```nix
configFile = "\"$HOME/wherever\"";
```

But that only works when `wrapperImplementation = "nix";`, the default, in the other 2 backends, variables either expand at build time or are fully escaped, which would obviously make that not work.

PR #420, opened 2026-04-07
feat(wrapperModules.eww): init

clay53:
I assume there's going to need to be some changes, but I'm not familiar enough with the repository to know what. An example usage of this is [here](https://github.com/upenn-embedded/final-project-s26-t1/blob/c6d3e6b447c0b07ec3dfde9098c35db7466d27fd/packages/eww.nix).


wrapperModules/e/eww/module.nix
```diff
@@ -0,0 +1,45 @@
...
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    yuck = lib.mkOption {
+      type = lib.types.str;
```
BirdeeHub:
```suggestion
      type = lib.types.lines;
```

So you can add to it across modules

If you want to overwrite, you can do that with lib.mkForce

wrapperModules/e/eww/module.nix
```diff
@@ -0,0 +1,45 @@
...
+      default = "";
+    };
+    style = lib.mkOption {
+      type = lib.types.str;
```
BirdeeHub:
```suggestion
      type = lib.types.lines;
```

same here

BirdeeHub:
Also, we are creating a directory, but technically we haven't fully ensured that those files always stay together in that directory.

Generally we have been making an option for the path for the config in the drv, and the output it will be generated into

https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L64-L78

providing it to the appropriate option, which you did, but also export it via `passthru` via `config.wrapper`

https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L81-L86

and ensuring that things stay in the specified output and path

https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L101-L102

Doing it in this way allows for easier access to the generated paths from both inside and outside of the wrapper module, even if all you have is the package, and ensures that the directory stays intact.

But also, other than that, seems fine to me, I will have to try it out and/or check more carefully in the eww docs when I get a chance!

wrapperModules/e/eww/module.nix
BirdeeHub: There is unfortunately no `lib.generators.toLisp` for us to use to pass automagically translated values like we do for lua, so `lib.types.lines` is just fine.

(unless you want to write one and PR that too, maybe it can also work for fennel! But that would be way above and beyond what is required here for this)

clay53: > There is unfortunately no `lib.generators.toLisp` for us to use to pass automagically translated values like we do for lua, so `lib.types.lines` is just fine.
> 
> (unless you want to write one and PR that too, maybe it can also work for fennel! But that would be way above and beyond what is required here for this)

That's what I thought. I'm glad it's not required :laughing: . Thank you

clay53:
> Also, we are creating a directory, but technically we haven't fully ensured that those files always stay together in that directory.
> 
> Generally we have been making an option for the path for the config in the drv, and the output it will be generated into
> 
> https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L64-L78
> 
> providing it to the appropriate option, which you did, but also export it via `passthru` via `config.wrapper`
> 
> https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L81-L86
> 
> and ensuring that things stay in the specified output and path
> 
> https://github.com/BirdeeHub/nix-wrapper-modules/blob/5369f45fb04a729074e2c6fe43d9f577318cd853/wrapperModules/h/helix/module.nix#L101-L102
> 
> Doing it in this way allows for easier access to the generated paths from both inside and outside of the wrapper module, even if all you have is the package, and ensures that the directory stays intact.
> 
> But also, other than that, seems fine to me, I will have to try it out and/or check more carefully in the eww docs when I get a chance!

These should be the rest of the suggestions. Is the `lib.mkOverride` in the constructFiles.*.relPath and .output necessary? It seems to work without

clay53:
I'm just testing out eww myself!

BirdeeHub:
> These should be the rest of the suggestions. Is the lib.mkOverride in the constructFiles.*.relPath and .output necessary? It seems to work without

It would work without them. However, you would be able to mkForce over them and put them somewhere else? And we are trying to guarantee they are in the directory.

I suppose technically it is up to you if you want to allow that? It feels more sound to just make it block all attempts to modify it to me personally, which is why I have been doing that.

If we wanted to allow that later, allowing it would be a nonbreaking change, but blocking it later would be a breaking change.

So, in fear of making a decision that is annoying to change, I have been adding mkOverride 0 (that is the max priority), which, in combination with how lib.types.str works where definitions across files do not append, blocks all attempts at modification.

If you find out you really want to move them later in some individual way, totally fine we can remove the mkOverride

But adding it later because of annoyance from people changing it when trying to mapAttrs into constructFiles, and filing issues asking what they did, but then adding it breaking people using that for hacks and getting more issues filed sounds worse.

Generally I like to underspecify when I have no invariants, and almost overspecify when I have invariants, the invariant in this case being, the config directory they both have to go into.

meanwhile, the rest of the files you want, it doesnt really matter where they are, you can call them from lisp.

You can build them with constructFiles, you can pass them directly into the text of the yuck lines option from nixpkgs, idk.

---

I would need to look and see if they have anything else that actually has to be in that config directory, if not, then this general design is completely prepared for the task. If so, we will have to do a little more, or find another way to provide our config files individually, because currently we are likely blocking other use of the config directory at runtime. Which is usually fine, and if they are following xdg correctly should be, but sometimes is not.

Edit: according to their docs, they do not edit the config directory, and that is all that is in there! This is good. I would still prefer it had the mkOverride 0 added to relPath and output for them

Seems like what we are doing is the supported way to do it too

wrapperModules/e/eww/module.nix
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    generatedConfig.output = lib.mkOption {
+      type = lib.types.str;
+      default = config.outputName;
+      description = ''
+        The derivation output for the config generated by this wrapper module
+      '';
+    };
+    generatedConfig.placeholder = lib.mkOption {
```
BirdeeHub: I guess, alternatively, we could just tie all of this to config.yuck and delete both these options

```nix
    constructFiles.yuck = {
      content = config.yuck;
      relPath = "${config.binName}-config/eww.yuck";
    };

    constructFiles.style = {
      content = config.style;
      output = lib.mkOverride 0 config.constructFiles.yuck.output;
      relPath = lib.mkOverride 0 "${dirOf config.constructFiles.yuck.relPath}/eww.${config.styleFormat}";
    };

    flags."--config" = dirOf config.constructFiles.yuck.path;

    passthru.generatedConfig = dirOf config.constructFiles.yuck.outPath;
```

BirdeeHub: honestly now that I look at it, this is kinda a nice way to do this I might start doing this more. Undecided. Up to you for this one entirely.

This way, with mkOverride 0 added to the relPath and output of the current constructFiles values, or the way I just put above.

clay53: I think this is much nicer. I'll implement this quick

clay53:
Also, I agree with the lib.mkOverride 0. I didn't understand its purpose before, but that makes a lot of sense.

BirdeeHub:
Hey, sorry, so I guess I was a bit tired.

I realized 2 things:

Including css files by absolute path from other css files can be somewhat annoying.

Would you mind if we did

```nix
    style = lib.mkOption {
      description = ''
        The CSS or SCSS style file to use for MPV.
      '';
      default = { };
      type = lib.types.submodule {
        options = {
          path = lib.mkOption {
            type = lib.types.nullOr wlib.types.stringable;
            default = null;
            description = ''
              Path to an existing file.
              Takes precedence over `content` if both are set.
            '';
          };
          content = lib.mkOption {
            type = lib.types.nullOr lib.types.lines;
            default = null;
            description = ''
              Inline css/scss file content.
              Used when `path` is null.
            '';
          };
          format = lib.mkOption {
            type = lib.types.enum [
              "css"
              "scss"
            ];
            default = "scss";
            description = ''
              File type extension of the style file.
            '';
          };
        };
      };
    };
```

And then

```nix
    constructFiles.style = {
      content = if builtins.isString (config.style.content or null) then config.style.content else "";
      output = lib.mkOverride 0 config.constructFiles.yuck.output;
      relPath = lib.mkOverride 0 "${dirOf config.constructFiles.yuck.relPath}/eww.${config.style.format}";
      ${if config.style.path or null != null then "builder" else null} =
        ''mkdir -p "$(dirname "$2")" && ln -s "${config.style.path}" "$2"'';
    };
```

(also `config.yuck` is missing a description field, and I am not sure why the tests aren't yelling at us for that?)

clay53:
> ```nix
> ${if config.style.path or null != null then "builder" else null} =
>         ''mkdir -p "$(dirname "$2")" && ln -s "${config.style.path}" "$2"'';
> ```

I've added the suggestions. I'm wondering, since I imagine the pattern of wanting to input files string or by file is going to be pretty common, if there should be like a `wlib.types.*` for it. For example, why not have it for `yuck` as well? I don't know if I should do that for this PR, but maybe later on or smth.

BirdeeHub:
We don't need it for `config.yuck` because `include` works with an absolute path in a satisfactory way whereas we do need it for the css because css imports like relative paths or url urls which would be slightly annoying.

There is a `wlib.types.file` type, but in that type path defaults to pkgs.writeText for the content option, and using pkgs.writeText like that makes it so that you cannot use placeholders in them, so in practice even when using wlib.types.file you still have to specify the default for both the content and the path to something from config.constructFiles

If it were to be added anywhere, it would probably be best to add directly to config.constructFiles in some way to be honest, but there is at least a way to do it for now, and this way will continue to work.

BirdeeHub:
Also, thank you for your patience, it looks good to me now!

clay53:
Nice! Thank you!

PR #421, opened 2026-04-07
feat(wrapperModules.mpv): expand script and config options

BirdeeHub:
see https://github.com/BirdeeHub/nix-wrapper-modules/pull/403

changes `config.scripts` list into a `config.script.<name>.{ path ? null, content ? null, opts ? { }, enable ? true }` option.

Detects if path is a nixpkgs one or not by checking for `passthru.scriptName`, if path is provided, path beats content

Adds `config.configDir.<name>.{ path ? null, content ? null, enable ? true }`

@iynaix, thoughts?

In particular, should we deprecate `config.scripts` and have the detection for if `.path` is a nixpkgs one, or leave both `config.script` and `config.scripts` as things and rename one of them? Maybe we could just rename `config.scripts` to `config.nixpkgsScripts` and remove the detection thing? But the detection thing is kinda nice also

### DIFF
--- a/wrapperModules/m/mpv/check.nix
+++ b/wrapperModules/m/mpv/check.nix
@@ -4,30 +4,164 @@
 }:
 
 let
-  mpvWrapped =
-    (self.wrappers.mpv.apply {
-      inherit pkgs;
-      scripts = [
-        pkgs.mpvScripts.visualizer
-      ];
-      "mpv.conf".content = ''
-        ao=null
-        vo=null
+  mpvWithPkgs = self.wrappers.mpv.apply { inherit pkgs; };
+  mpvWithoutConfigDirectory = mpvWithPkgs.wrap {
+    script.visualizer.path = pkgs.mpvScripts.visualizer;
+    "mpv.conf".content = ''
+      ao=null
+      vo=null
+    '';
+  };
+  mpvWithConfigDirectory = mpvWithPkgs.wrap {
+    script.visualizer.path = pkgs.mpvScripts.visualizer;
+    configDir = {
+      "script-opts/visualizer.conf".content = ''
+        mode="force"
       '';
-    }).wrapper;
+    };
+    "mpv.conf".content = ''
+      ao=null
+      vo=null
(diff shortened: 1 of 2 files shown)


iynaix:
I think the detection thing is nice and should stay. It's not a big deal, but `config.script.<name>` feels kind of weird given that you could set multiple of them and it is singular.

I was going to suggest using `script-opts-append` in the mpv.conf instead of as flags, but looking at my own config, I have mpv profiles, which would complicate the ordering within mpv.conf, so I think flags is the way to go.

BirdeeHub:
`config.script.<name>`

I mean, yeah I totally agree

But `scripts` was taken, and I didnt want to make it super long?

You uhhhh

`config.script.<name>`... I mean... that individual expression technically only defines 1 script?

but yeah I agree, but `scriptFiles` I decided was too long for what it was?

> I was going to suggest using script-opts-append in the mpv.conf instead of as flags

Wait you can do that?

I guess that makes sense actually...

I mean....

I could do it that way?

But, I guess, this way they are more indepdendent, IDK. And it sounds like you think that this way might be at least as good? So, I guess, we can see how it goes like this then.

iynaix:
So what I meant was

```
# mpv config blah blah

script-opts-append=....
```

breaks if you use profiles like I do, e.g.

```
# mpv config blah blah

[anime]
# profile settings blah blah

script-opts-append=....
```

because that would only apply the script-opts to the anime profile; unfortunately order matters.

So the flags approach is better imo, even though it is uglier.

BirdeeHub:
ah ok ty for the explanation!

Hopefully this works well, if you find some annoyances do let me know again.

Sorry for taking the PR from where you started and running away with it myself, but after exploring it I got a vision of approximately what I wanted and rather than dictate it to you and make you write it, I figured I would give us a better starting point for the improvements first.

Or, at least, I didn't get a vision of what I wanted. Its more like, I think I saw what it seemed they wanted us to do, if that makes sense. Or at least close enough.

PR #427, opened 2026-04-10
feat(wrapperModules.glance): init

jtrrll:
Adds a wrapper module for the lightweight, portable dashboard [glance](https://github.com/glanceapp/glance).

### DIFF
--- a/wrapperModules/g/glance/module.nix
+++ b/wrapperModules/g/glance/module.nix
@@ -1,0 +1,31 @@
+{
+  config,
+  lib,
+  pkgs,
+  wlib,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    settings = lib.mkOption {
+      inherit (pkgs.formats.yaml { }) type;
+      default = { };
+      description = ''
+        Configuration for glance.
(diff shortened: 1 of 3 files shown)


wrapperModules/g/glance/module.nix
```diff
@@ -0,0 +1,31 @@
...
+      description = ''
+        Configuration for glance.
+        See <https://github.com/glanceapp/glance/blob/main/docs/configuration.md>
+        for available options.
+      '';
+    };
+  };
+  config = {
+    constructFiles.generatedConfig = {
+      content = lib.generators.toYAML { } config.settings;
```
BirdeeHub, line 23: you tested this and it works?

lib.generators.toYAML is just builtins.toJSON

https://github.com/NixOS/nixpkgs/blob/4eb3c802539f72fe7538cb049c65ae2cb27a45e6/lib/generators.nix#L923

If this works for this program, nice.

It looks like it passes their validate command, so that is a good sign that it does.

So, this is good.

Still need to look into the docs for glance and stuff, to make sure we aren't missing something important, but it looks good so far.

BirdeeHub:
Tested it, it works, it looks like its just the 1 config file, placeholders work in it because you used lib.generators.toYAML and constructFiles, LGTM thank you!

PR #428, opened 2026-04-10
feat(wrapperModules.wlr-which-key): init

nouritsu:
https://github.com/MaxVerevkin/wlr-which-key

### DIFF
--- a/wrapperModules/w/wlr-which-key/module.nix
+++ b/wrapperModules/w/wlr-which-key/module.nix
@@ -1,0 +1,53 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  yamlFmt = pkgs.formats.yaml { };
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    configPath = lib.mkOption {
+      type = wlib.types.stringable;
+      default = config.constructFiles.cfg.path;
+      description = "Path to YAML configuration file.";
+    };
+
+    settings = lib.mkOption {
+      type = yamlFmt.type;
+      default = { };
+      description = ''
+        Configuration for wlr-which-key.
+        See <https://github.com/MaxVerevkin/wlr-which-key>
+      '';
+    };
+
+    initialKeys = lib.mkOption {
+      type = lib.types.str;
+      default = "";
+      description = ''
+        Space separated key sequence to execute after launching menu.
+      '';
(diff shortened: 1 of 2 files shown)


wrapperModules/w/wlr-which-key/module.nix
```diff
@@ -0,0 +1,47 @@
...
+    };
+  };
+
+  config.package = lib.mkDefault pkgs.wlr-which-key;
+
+  config.flags."--initial-keys" = lib.mkIf (config.initialKeys != "") config.initialKeys;
+  config.addFlag = [ config.constructFiles.cfg.path ];
+
+  config.constructFiles.cfg = {
+    content = builtins.readFile (yamlFmt.generate "${config.binName}.yaml" config.settings);
```
BirdeeHub:
```suggestion
    content = lib.generators.toYAML { } config.settings;
```

Apparently they have a toYAML now? They have it because they just use toJSON because yaml is a superset of json.

If that doesnt work for some reason, you can use

```nix
    builder = ''mkdir -p "$(dirname "$2")" && ${pkgs.remarshal}/bin/json2yaml "$1" "$2"'';
```

Which is how `yamlFmt.generate` works still and should generate actual yaml instead of json

wrapperModules/w/wlr-which-key/module.nix
```diff
@@ -0,0 +1,47 @@
...
+        Configuration for wlr-which-key.
+        See <https://github.com/MaxVerevkin/wlr-which-key>
+      '';
+    };
+  };
+
+  config.package = lib.mkDefault pkgs.wlr-which-key;
+
+  config.flags."--initial-keys" = lib.mkIf (config.initialKeys != "") config.initialKeys;
+  config.addFlag = [ config.constructFiles.cfg.path ];
```
BirdeeHub: should we be doing some kind of `mkIf` on this `addFlag`, since it is slightly harder to override values in lists?

other than this and the toYAML thing though, based on what I can tell from the docs this is reasonable.

Maybe instead of `mkIf` we could make a `configPath` option that takes a `wlib.types.stringable`, which is by default `config.constructFiles.cfg.path`?

IDK, but one of those 2 things. `mkIf` is probably easier.

nouritsu: I would like to make a `configPath` option instead. I'll resolve this today.

PR #430, opened 2026-04-11
feat(wrapperModules.parallel): init

xavwe:
https://www.gnu.org/software/parallel/

### DIFF
--- a/maintainers/default.nix
+++ b/maintainers/default.nix
@@ -73,4 +73,10 @@
     github = "clay53";
     githubId = 16981283;
   };
+  xavwe = {
+    email = "git@xavwe.dev";
+    github = "xavwe";
+    githubId = 125409009;
+    name = "Xaver Wenhart";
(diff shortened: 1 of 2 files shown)


BirdeeHub:
https://www.gnu.org/software/parallel/parallel_design.html#citation-notice

^ this mentions `--citation` but not `--will-cite`

In addition, why only that 1 flag?

`inputs.nix-wrapper-modules.lib.wrapPackage { inherit pkgs; package = pkgs.parallel-full; flags."--will-cite" = true; }`

If there isn't anything else that would be advantageous to set via a general purpose wrapper script, do we need such a wrapper module pre-built for users? If everyone will wrap it in a custom way, I am not sure the benefit of having a standard wrapper module we have to maintain here for the 1 boolean argument.

xavwe:
Yeah, you’re right. I will close this for now.

PR #435, opened 2026-04-14
feat(wrapperModules.hyprlock): init

nouritsu:
https://wiki.hypr.land/Hypr-Ecosystem/hyprlock/

### DIFF
--- a/wrapperModules/h/hyprlock/module.nix
+++ b/wrapperModules/h/hyprlock/module.nix
@@ -139,7 +139,7 @@ in
     "hyprlock.conf" = lib.mkOption {
       type = wlib.types.file {
         path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
-        content = lib.mkOptionDefault (
+        content = (
           lib.optionalString (config.settings != { }) (toHyprconf {
             inherit (config) importantPrefixes;
             attrs = config.settings;
--- a/wrapperModules/h/hyprlock/module.nix
+++ b/wrapperModules/h/hyprlock/module.nix
@@ -1,0 +1,199 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  /*
+    from:
+    https://github.com/nix-community/home-manager/blob/8a423e444b17dde406097328604a64fc7429e34e/modules/lib/generators.nix
+  */
+  toHyprconf =
+    {
+      attrs,
+      indentLevel ? 0,
+      importantPrefixes ? [ "$" ],
+    }:
+    let
+      inherit (lib)
+        all
+        concatMapStringsSep
+        concatStrings
+        concatStringsSep
+        filterAttrs
+        foldl
+        generators
+        hasPrefix
+        isAttrs
+        isList
+        mapAttrsToList
+        replicate
+        attrNames
+        ;
+
+      initialIndent = concatStrings (replicate indentLevel "  ");
+
+      toHyprconf' =
+        indent: attrs:
+        let
+          isImportantField =
+            n: _: foldl (acc: prev: if hasPrefix prev n then true else acc) false importantPrefixes;
+          importantFields = filterAttrs isImportantField attrs;
+          withoutImportantFields = fields: removeAttrs fields (attrNames importantFields);
+
+          allSections = filterAttrs (_n: v: isAttrs v || isList v) attrs;
+          sections = withoutImportantFields allSections;
+
+          mkSection =
+            n: attrs:
+            if isList attrs then
+              let
+                separator = if all isAttrs attrs then "\n" else "";
+              in
+              (concatMapStringsSep separator (a: mkSection n a) attrs)
+            else if isAttrs attrs then
+              ''
+                ${indent}${n} {
+                ${toHyprconf' "  ${indent}" attrs}${indent}}
+              ''
+            else
+              toHyprconf' indent { ${n} = attrs; };
+
+          mkFields = generators.toKeyValue {
+            listsAsDuplicateKeys = true;
+            inherit indent;
+          };
+
+          allFields = filterAttrs (_n: v: !(isAttrs v || isList v)) attrs;
+          fields = withoutImportantFields allFields;
+        in
+        mkFields importantFields
+        + concatStringsSep "\n" (mapAttrsToList mkSection sections)
+        + mkFields fields;
+    in
+    toHyprconf' initialIndent attrs;
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    settings = lib.mkOption {
+      /*
+        from:
+        https://github.com/nix-community/home-manager/blob/8a423e444b17dde406097328604a64fc7429e34e/modules/programs/hyprlock.nix
+      */
+      type =
+        with lib.types;
+        let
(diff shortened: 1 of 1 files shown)


wrapperModules/h/hyprlock/module.nix
```diff
@@ -0,0 +1,125 @@
...
+              path
+              (attrsOf valueType)
+              (listOf valueType)
+            ])
+            // {
+              description = "Hyprlock configuration value";
+            };
+        in
+        valueType;
+      default = { };
```
BirdeeHub, line 105: https://github.com/nix-community/home-manager/blob/8a423e444b17dde406097328604a64fc7429e34e/modules/programs/hyprlock.nix#L59-L102

Home manager has an example thing, would you mind adding an example here as well?

wrapperModules/h/hyprlock/module.nix
```diff
@@ -0,0 +1,125 @@
...
+        See <https://wiki.hypr.land/Hypr-Ecosystem/hyprlock>
+      '';
+    };
+  };
+
+  config.package = lib.mkDefault pkgs.hyprlock;
+  config.flags."--config" = lib.mkIf (config.settings != { }) config.constructFiles.cfg.path;
+
+  config.constructFiles.cfg = {
+    content = toHyprconf { attrs = config.settings; };
```
BirdeeHub:
```suggestion
    content = toHyprconf { attrs = config.settings; importantPrefixes = config.importantPrefixes; };
```

Home manager has some kind of importantPrefixes option thing, and you have the code that handles it here already, but no option to provide it.

BirdeeHub (commented):
Other than needing an example of some kind for users, and a missing option, seems like the right approach, and there aren't any problems I see with the approach.

wrapperModules/h/hyprlock/module.nix
BirdeeHub: You also might want to add the extraConfig and sourceFirst things they have? https://github.com/nix-community/home-manager/blob/8a423e444b17dde406097328604a64fc7429e34e/modules/programs/hyprlock.nix#L110-L120

nouritsu: I'll make these changes. Would you advise using wlib.types.file for such cases (extraConfig could just append to content then)? It's not very clear to me where I should be using that and where I should just define my own options. It looks like the wlib file type gives you content and path which is more useful than what I did in the wlr-which-key contribution for example.

BirdeeHub: It is honestly not entirely clear to me when one should be using `wlib.types.file` either.

In its default form it uses `pkgs.writeText`, which I don't really like in the submitted modules because then placeholders don't refer to the final wrapper derivation in that file anymore. It works alright in your config if you don't need that, but in ones submitted here I generally would like to try to get people to set the default path for it themselves if they use it in something submitted here. The reason we have it, is because lassulus wrappers had it. And honestly, it is one of the few things that survived mostly unmodified from the rewrite. Basically just that and some portion of the tests, and the names of some of the options survived the rewrite.

In this case, I think the best design would be to make a path option

```nix
lib.mkOption {
  type = wlib.types.stringable;
  default = config.constructFiles.cfg.path;
};
```

And then the pre and post extraConfig options that home manager does.


But I would accept one that uses `wlib.types.file` somehow, you would need to set `default.path = config.constructFiles.cfg.path;`, and then either set the default content to the generated value, or take the value of content and use it in the generated config.

Personally I would prefer just a path option and then the pre and post extraConfig things home manager does, just `lib.types.lines` options that get placed before and after the generated content.

Technically, they could already do pre and post extraConfig

`config.constructFiles.cfg.content = "appended";`

`config.constructFiles.cfg.content = lib.mkBefore "prepended";`

all using `wlib.types.file` here would do, would make it more obvious that this content option exists to append or prepend to. Whereas giving explicit before and after is doing a little more of the work for them and feels like a better guarantee for the user that that thing is going to work in that way correctly, and gives them that same prepend or append control with mkBefore for both the prepended and the appended content.



Given that even with their toHyperconf and that just setting a home.files value you can technically append or prepend to already, kinda like our constructFiles, they deemed it necessary, we should probably do it too, and throw in a path option for overriding that as well, basically.

nouritsu:
Added example and `importantPrefixes` option. I am not adding `sourceFirst` to keep the module leaner (users can simply add "source" to `importantPrefixes`).

wrapperModules/h/hyprlock/module.nix
```diff
@@ -0,0 +1,196 @@
...
+
+    "hyprlock.conf" = lib.mkOption {
+      type = wlib.types.file pkgs;
+      default.path = config.constructFiles.generatedConfig.path;
+      default.content =
+        lib.optionalString (config.settings != { }) (toHyprconf {
+          inherit (config) importantPrefixes;
+          attrs = config.settings;
+        })
+        + lib.optionalString (config.extraConfig != "") config.extraConfig;
```
BirdeeHub:
```suggestion
      type = wlib.types.file {
        path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
        content =
          lib.optionalString (config.settings != { }) (toHyprconf {
            inherit (config) importantPrefixes;
            attrs = config.settings;
          })
          + lib.optionalString (config.extraConfig != "") config.extraConfig;
      };
```

I very recently updated `wlib.types.file`

That way you had it would work too, I made the change in a backwards compatible manner, but if they set content, it would remove path unless you do a workaround...

See https://github.com/BirdeeHub/nix-wrapper-modules/issues/447 from literally yesterday sorry about that

Otherwise it looks good though!

nouritsu: Alright, I'm also adding the `default = {}` and a `lib.mkOptionDefault` on content since WezTerm's wrapper seems to do the same.

nouritsu:
Done. Also rebased with synced `main`.

wrapperModules/h/hyprlock/module.nix
BirdeeHub: ty I forgot `default = { };`

wrapperModules/h/hyprlock/module.nix
```diff
@@ -0,0 +1,199 @@
...
+      description = ''
+        Configuration for Hyprlock.
+        See <https://wiki.hypr.land/Hypr-Ecosystem/hyprlock>
+      '';
+    };
+
+    "hyprlock.conf" = lib.mkOption {
+      type = wlib.types.file {
+        path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
+        content = lib.mkOptionDefault (
```
BirdeeHub:
```suggestion
        content = (
```

Sorry.. one last thing... Just a suggestion but, do you want assigning to this to extend the existing by default, or overwrite it?

As you have it currently, it would overwrite, without `lib.mkOptionDefault` it would append.

BirdeeHub: Just making sure that was the intention, rather than just copying the wezterm wrapper

The wezterm default ends in return so append by default would be bad, which is why i chose that for that particular wrapper.

It seems like this format appends well, so to me I would think not using `lib.mkOptionDefault` here would be the move, which is why I bring it up.

PR #431, opened 2026-04-12
chore(deps): update flake.lock files

BirdeeHub:
This PR updates the Nix flake.lock files automatically.

PR #439, opened 2026-04-15
feat(tests): add test utils (tlib)

zenoli:
Final version, combining the good stuff from this original draft (V1) and https://github.com/BirdeeHub/nix-wrapper-modules/pull/452 (V3).

A brief overview of what changed:

- set based, nix-native syntax. No more `runTests/runTest` or `describe/it`. (Version A from [this comment])(https://github.com/BirdeeHub/nix-wrapper-modules/pull/452#discussion_r3124277838)
- Arbitrarily nestable tests from V3 by generating rendering the test attr set into a complex bash command that logs nice error messages wit a full trace of the nesting hierarchy.
- Tests automatically respecting the `meta.platforms` of a wrapper module, if the wrapper name is passed to the test in the first argument (`settings.wrapper`).
- NEW: Debug flag to generate an executable version of the rendered script for inspection and debugging
- NEW: Support for multi-line assertions and error messages.
- NEW: Shorthand syntax for assertions: Can be provided as a string directly without an error message, in which case an error message, echoeing the failed condition string will be automatically generated.

@BirdeeHub the one thing I did not re-implement from your version is the `extend` feature. If you want this I will have to study it in more detail. I think it should be possible to add if it is really needed. Just let me know.

- [x] Remove direnv module
- [x] Add a nice way to run tests only for systems it is defined in
- [x] Update CONTRIBUTING.md

### DIFF
--- a/ci/docs/default.nix
+++ b/ci/docs/default.nix
@@ -208,6 +208,23 @@ in
           src = "${placeholder "generated"}/wrapper_docs/${n}.md";
         }) (builtins.attrNames config.drv.wrapper_docs);
       }
+      {
+        name = "Contributing";
+        data = "numbered";
+        path = "md/CONTRIBUTING.md";
+        src = ../../CONTRIBUTING.md;
+        subchapters = [
+          {
+            name = "tlib";
+            data = "numbered";
+            path = "tlib.md";
+            src = "${placeholder "out"}/wrappers-lib/tlib.md";
+            build = ''
+              ${pkgs.nixdoc}/bin/nixdoc --category "" --description 'Testing library `tlib` documentation' --file '${../test-lib.nix}' --prefix "tlib" >> $out/wrappers-lib/tlib.md
+            '';
+          }
+        ];
+      }
     ];
   };
 }
--- a/ci/flake.nix
+++ b/ci/flake.nix
@@ -20,6 +20,7 @@
             inherit system;
             config.allowUnfree = true;
           };
+          tlib = pkgs.callPackage ./test-lib.nix { inherit self; };
 
           # Load checks from ci/checks/ directory
           coreAndCiChecks = lib.pipe ./checks [
@@ -43,7 +44,7 @@
                     name = "${prefix}-${name}";
                     inherit value;
                   };
-                  result = pkgs.callPackage value { inherit self; };
+                  result = pkgs.callPackage value { inherit self tlib; };
                 in
                 if result == null then
                   [ ]
--- a/ci/test-lib.nix
+++ b/ci/test-lib.nix
@@ -1,0 +1,328 @@
+{
+  self,
+  lib,
+  runCommand,
+  writeShellScript,
+  stdenv,
+  pkgs,
+  ...
+}:
+let
+  wlib = self.lib;
+
+  errMsg = msg: "(echo ${renderMsg msg} >&2 && return 1)";
+
+  indentBlock =
+    str: num:
+    let
+      idnt = "  ";
+      lines = lib.splitString "\n" str;
+      indentLine = line: (wlib.repeatStr idnt num) + line;
+    in
+    lib.concatMapStringsSep "\n" indentLine lines;
+
+  # Use [ANSI-C Quoting](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html#ANSI_002dC-Quoting-1)
+  # to properly display multi-line messages without being indented if occurring in deeply nested tests.
+  renderMsg =
+    str:
+    let
+      lines = lib.splitString "\n" str;
+      result = "$'${lib.concatStringsSep "\\n" lines}'";
+    in
+    if (lib.length lines) == 1 then str else result;
+
+  toSanitizedJSON =
+    value:
+    if builtins.isAttrs value then
+      builtins.toJSON (
+        lib.mapAttrsRecursive (
+          path: v:
+          if builtins.isFunction v then
+            let
+              res = builtins.unsafeGetAttrPos (lib.last path) (
+                lib.getAttrFromPath (lib.sublist 0 (builtins.length path - 1) path) value
+              );
+            in
+            "<lambda${
+              if builtins.isAttrs res then
+                ":${res.file or ""}:${toString (res.line or "")}:${toString (res.column or "")}"
+              else
+                ""
+            }>"
+          else
+            v
+        ) value
+      )
+    else
+      builtins.toJSON value;
+
+  render =
+    node:
+    let
+      renderAssertion =
+        let
+          _renderAssertion =
+            { cond, msg }:
+            ''
+              (
+                ${cond}
+              ) || ${errMsg msg}'';
+        in
+        assertion:
+        if lib.isString assertion then
+          _renderAssertion {
+            cond = assertion;
+            msg = "Failed assertion: ${assertion}";
+          }
+        else
+          _renderAssertion assertion;
+
+      renderAssertions =
+        assertions: lib.concatMapStringsSep " &&\n" (a: "(${renderAssertion a})") assertions;
+
+      renderNode =
+        node:
+        let
+          block =
+            if !(lib.isAttrs node) then
+              renderAssertions (lib.toList node)
+            else
+              lib.concatMapAttrsStringSep " &&\n" (name: childNode: ''
+                ((
+                ${renderNode childNode}
+                ) || ${errMsg name})'') node;
+        in
+        indentBlock block 1;
+    in
+    renderNode node;
+
+  _test =
+    {
+      name,
+      debug ? false,
+    }:
+    testSet:
+    let
+      testScript = ''
+        run () {
+          ${render testSet};
+        }
+        run || (echo "Test '${name}' failed" >&2 && exit 1)
+      '';
+    in
+    if debug then
+      writeShellScript name ''
+        ${testScript}
+      ''
+    else
+      runCommand name { passthru.test = testScript; } ''
+        ${testScript}
+        touch $out
+      '';
+
+in
+{
+  /**
+    Takes tests provided as an attrs in `testSet`, renders it to a bash script,
+    and executes the rendered script using `pkgs.runCommand`.
+
+    If the command runs successfully, the test set passes. Otherwise, a detailed
+    error message describing the failed assertion is shown in the logs.
+
+    # Type
+    ```
+    test :: (String | AttrSet) -> TestSet -> Derivation
+    ```
+
+    where:
+
+    ```
+    TestSet :: attrsOf (TestSet | Test)
+
+    Test :: [ Assertion ]
+
+    Assertion :: String | { cond :: String; msg :: String; }
+    ```
+
+    # Arguments
+
+    settings
+    : Either a string or an attrs.
+
+      If it is a string, it will be taken as the name of the derivation built
+      by `runCommand`.
+
+      If it is a set, at least one of the keys `name` or `wrapper` need to be set
+      with a string value.
+
+      - If you are defining tests for wrappers you should set `settings.wrapper` to the name
+      of the wrapper. This way, the `wrapper.meta.platforms` is considered during test execution,
+      and the test will only be run if the system it is running on is supported.
+      - If `settings.wrapper` is not set, the test will always be run. This makes sense if you
+      are testing core options or library functions.
+      - If only `settings.wrapper` is set, the name will be derived from this value by suffixing it with `-test`.
+      - If `settings.name` is set, it will be taken as the name of the derivation.
+      - If both are set, `settings.name` takes precedence.
+
+      There is also an optional boolean option `settings.debug` (default = `false`).
+      If it is set to `true`, the generated bash script is built as an executable script that can be inspected
+      and run in `result` when running a specific test.
+
+      The test can be disabled by setting `settings.enable = false`.
+
+    testSet
+    : The set of tests to run.
+  */
+  test =
+    settings: testSet:
+    let
+      name =
+        if lib.isString settings then
+          settings
+        else if settings ? name then
+          settings.name
+        else if settings ? wrapper then
+          "${settings.wrapper}-test"
+        else
+          throw ''
+            Invalid argument for `test`.
+            The first argument must be either a string (the test name) or an attrs
+            with at least one of the keys 'name' or 'wrapper', but got:
+
+            ${lib.toSanitizedJSON settings}
+          '';
+      wrapperName = settings.wrapper or null;
+      platforms =
+        if wrapperName == null then
+          null
+        else if (wrapperName != null) && (lib.hasAttr wrapperName self.wrappers) then
+          self.wrappers.${wrapperName}.meta.platforms
+        else
+          throw ''
+            Invalid argument for `test`.
+            The provided wrapper '${wrapperName}' was not found.
+            Available wrappers are:
+
+            ${lib.toSanitizedJSON (lib.attrNames self.wrappers)}
+          '';
+      enabled =
+        (settings.enable or true)
+        && (platforms == null || (platforms != null && builtins.elem stdenv.hostPlatform.system platforms));
+    in
+    if enabled then
+      _test {
+        inherit name;
+        debug = settings.debug or false;
+      } testSet
+    else
+      null;
+
+  /**
+    Returns an `Assertion` that checks whether `path` is an existing directory.
+
+    # Type
+    ```
+    isDirectory :: String -> Assertion
+    ```
+
+    # Arguments
+    path
+    : The filesystem path to check.
+  */
+  isDirectory = path: {
+    cond = ''[[ -d "${path}" ]]'';
+    msg = "No such directory ${path}";
+  };
+
+  /**
+    Returns an `Assertion` that checks whether `path` is an existing regular file.
+
+    # Type
+    ```
+    isFile :: String -> Assertion
+    ```
+
+    # Arguments
+    path
+    : The filesystem path to check.
+  */
+  isFile = path: {
+    cond = ''[ -f "${path}" ]'';
+    msg = "No such file ${path}";
+  };
+
+  /**
+    Returns an `Assertion` that checks whether `path` does **not** exist as a regular file.
+
+    # Type
+    ```
+    notIsFile :: String -> Assertion
+    ```
+
+    # Arguments
+    path
+    : The filesystem path that should be absent.
+  */
+  notIsFile = path: {
+    cond = ''[ ! -f "${path}" ]'';
+    msg = "File ${path} should not exist";
+  };
+
+  /**
+    Returns an `Assertion` that checks whether `file` contains a line matching `pattern`.
+
+    The check is performed with `grep -q`, so `pattern` is treated as a basic regular expression.
+
+    # Type
+    ```
+    fileContains :: String -> String -> Assertion
+    ```
+
+    # Arguments
+    file
+    : Path to the file to search.
+
+    pattern
+    : Basic regular expression to search for.
+  */
+  fileContains = file: pattern: {
+    cond = ''grep -q '${pattern}' "${file}"'';
+    msg = "Pattern '${pattern}' not found in ${file}";
+  };
+
+  /**
+    Returns an `Assertion` that checks whether `expected` and `actual` are equal.
+
+    The comparison is performed in Nix at evaluation time. If the values differ,
+    the assertion message shows both values serialised as JSON.
+
+    # Type
+    ```
+    areEqual :: Any -> Any -> Assertion
+    ```
+
+    # Arguments
+    expected
+    : The expected value.
+
+    actual
+    : The value to compare against `expected`.
+  */
+  areEqual =
+    expected: actual:
+    let
+      equal = expected == actual;
+    in
+    {
+      cond = if equal then "true" else "false";
+      msg =
+        if !equal then
+          ''
+            Expected:
+              ${toSanitizedJSON expected}
+            but got:
+              ${toSanitizedJSON actual}''
+        else
+          "This should never happen.";
+    };
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Adding Modules!
+## Adding Modules!
 
 There are 2 kinds of modules in this repository. One kind which defines the `package` option, and one kind which does not.
 
@@ -121,15 +121,15 @@ Example:
 }
 ```
 
-# Formatting
+## Formatting
 
 `nix fmt`
 
-# Tests
+## Tests
 
 `nix flake check -Lv ./ci`
 
-# Run Site Generator Locally
+## Run Site Generator Locally
 
 `nix run ./ci`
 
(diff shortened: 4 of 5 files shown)


ci/flake.nix
```diff
@@ -43,9 +44,9 @@
                     name = "${prefix}-${name}";
                     inherit value;
                   };
-                  result = pkgs.callPackage value { inherit self; };
+                  result = pkgs.callPackage value { inherit self tlib; };
                 in
-                if isNull result then
+                if result == null then
```
zenoli: Apparently this is deprecated:
https://nix.dev/manual/nix/2.18/language/builtins#builtins-isNull

BirdeeHub: oh wow I had no idea. I now have changed both `isNull` in the codebase into `== null`.

BirdeeHub:
Yes that would be where to pass them

I didnt think about it needing all the `...` but it would yeah... `pkgs.callPackage` only provides stuff you ask for, but for `self` and now `tlib` which are passed explicitly, those always get passed even if you don't ask for them, so it either needs both `self` and `tlib` arguments, or the `...`, or both.

zenoli:
> Yes that would be where to pass them
> 
> I didnt think about it needing all the `...` but it would yeah... `pkgs.callPackage` only provides stuff you ask for, but for `self` and now `tlib` which are passed explicitly, those always get passed even if you don't ask for them, so it either needs both `self` and `tlib` arguments, or the `...`, or both.

Yes, I sighed when I saw that, but it was less annoying than anticipated using quickfix list and a small macro :-)

Cool, I will finalize the PR soon then.

BirdeeHub:
oh actually.

Maybe we could just export tlib from the ci flake? It will warn about that not following the schema, I kinda don't care, but also maybe we could just update { inherit tlib; } into the lib output in the ci flake and then that solves that part too?

Then ppl could just grab self.tlib (or self.lib.tlib) like they currently just grab self.lib for wlib.

Honestly, Im not sure if that is better though. It is just another idea, fully up to you, it doesn't matter to me. Not really sure what the best way to pass them in is, but, that is where they are called, so, whatever way it gets passed would have to go through that same place in some way.

I am completely fine with the way currently shown in the PR for passing them in, so, yeah IDK, just throwing out more ideas.

ci/test-lib.nix
```diff
@@ -0,0 +1,66 @@
...
+    file: pattern:
+    createAssertion {
+      cond = ''grep -q '${pattern}' "${file}"'';
+      message = "Pattern '${pattern}' not found in ${file}";
+    };
+
+  runTests =
+    wrapperModule: scripts:
+    let
+      wrapper = wrapperModule.apply { inherit pkgs; };
```
zenoli: Passing the module here would be enough for the `platforms` guard, but I would also like to grab `wrapper.binName` for the testname. Hence, I need to apply `pkgs` here. Do you have an issue with that @BirdeeHub ?

And now since I basically have a default wrapper available, I was thinking of passing it to `runTest` such that they can use and extend this instead of each `runTest` having to define a wrapper from scratch using `let ... in` as I currently do in `direnv/check.nix`.

Thoughts @BirdeeHub ?

BirdeeHub: As far as passing pkgs goes, sure, that is fine. They get the same pkgs given to them.

You should create the test drv with the initial cases/scripts and module they pass and stuff. I like the idea of not passing the module a bunch of times and only passing it once per test drv.

You should allow the helper to also recieve a package if you can, the stuff is exposed under passthru.configuration if you get a package.

You should try to use [`wlib.makeCustomizable`](https://birdeehub.github.io/nix-wrapper-modules/lib/wlib.html#function-library-wlib.makeCustomizable) in order to allow them to `test.addTest` or `test.addScript` or `test.addCase` or whatever you wanna call it afterwards too! (Not necessary though. But it would be cool right?) It is a generalization of `lib.mkOverrideable` that allows that function to be whatever you want.

zenoli:
I implemented the wrapper-passing-solution on a separate branch, because I am not sure whether I am happy with it:
https://github.com/zenoli/nix-wrapper-modules/pull/2

The idea is as follows:

```nix
runTests self.wrappers.direnv [  # <-- 1. specify the wrapper module here
  (runTest "if nix-direnv is enabled then lib/nix-direnv.sh should exists"
    { nix-direnv.enable = true; }  # <-- 2. specify the test-specific wrapper config
    (wrapper: [  #  <-- 3. receive the wrapper package with the config from (2.) already applied
      (isDirectory (getDotdir wrapper))   # <-- 4. define assertions
      (isFile "${getDotdir wrapper}/lib/nix-direnv.sh")
    ])
  )
  ...
```

The example above, everything seems pretty clean imo.
Where I feel it gets a bit clunky is if you want to define variables that need to be used in the wrapper config as well as in the assertions:

```nix
  (
    let
      libScriptContent = "echo foo";  # needs to be outside of `runTest` as the scope needs to cover the config and assertions
    in
    (runTest "if a lib-script is set then it should be generated" { lib."foo.sh" = libScriptContent; } (
      wrapper:
      let
        libScriptFile = "${getDotdir wrapper}/lib/foo.sh";
      in
      [
        (isDirectory (getDotdir wrapper))
        (isFile libScriptFile)
        (fileContains libScriptFile libScriptContent)

      ]
    ))
  )
```

Maybe I should scratch the idea of passing the config and simply receive the wrapper (with only `pkgs` applied) and leave extending the wrapper do the user.

My issue with this is, that I have to invent a new wrapper name:

```nix
  (runTest "if nix-direnv is enabled then lib/nix-direnv.sh should exists"  # <-- no more config param here
    (wrapper: 
      let
        wrapper2 = wrapper.wrap { nix-direnv.enable = true; };  # <-- here I need a new name
      in 
        [
          (isDirectory (getDotdir wrapper))
          (isFile "${getDotdir wrapper}/lib/nix-direnv.sh")
        ])
  )

```

There are of course better ways to than just picking "wrapper2" but it is still odd.

BirdeeHub:
Ah I see. So, because the whole test owns a copy of the module, but you may want variations of it for each case.

Im not sure. Maybe them owning a copy of it is the wrong abstraction?

Or, you could have each case be allowed to provide its own module if they want? Then the question with that would be, should the next test get the updated state of that, or should it receive the original copy? The user should probably be able to specify that?

Also, you meant this, right?

```nix
  (runTest "if nix-direnv is enabled then lib/nix-direnv.sh should exists"  # <-- no more config param here
    (wrapper: 
      let
        wrapper2 = wrapper.wrap { nix-direnv.enable = true; };  # <-- here I need a new name
      in 
        [
          (isDirectory (getDotdir wrapper2))
          (isFile "${getDotdir wrapper2}/lib/nix-direnv.sh")
        ])
  )
```

zenoli:
> Or, you could have each case be allowed to provide its own module if they want? 

Yes, this is the most flexible solution (and basically what I had in my first draft):

```nix
(runTest "if nix-direnv is disabled then lib/nix-direnv.sh should not exist" (
    let
      wrapper = self.wrappers.direnv.wrap {
        inherit pkgs;
        nix-direnv.enable = false;
      };
    in
    [
      (isDirectory (getDotdir wrapper))
      (notIsFile "${getDotdir wrapper}/lib/nix-direnv.sh")
    ]
  ))
```

Maybe I will make `runTest` able to recieve both a function `wrapper -> <listOf Assertion>` (where wrapper is pre-configured as in my 2nd draft) but also simply a `<listOf Assertions>` where they can build their wrapper from scratch using `let .. in` (as above).

> Then the question with that would be, should the next test get the updated state of that, or should it receive the original copy? The user should probably be able to specify that?

I'm not sure if I understand what you mean. I think each `runTest` should - in some way - provide its own wrapper variant s.t. they stay independent and modular. I wouldn't want a solution where wrappers get successively extended for `runTest` later in the list, if this is what you meant? 
What could make sense if if tests could be arbitrarily nested, and all children inheriting the base wrapper of their parent, but I think this is overkill.

> Also, you meant this, right?

Yes, exactly. Forgot to update `wrapper2` in the assertions.

All in all, I think the initial solution was the most [KISS](https://en.wikipedia.org/wiki/KISS_principle) aproach.
It is flexible, and the syntax is clean and readable (which I think is important for incentivising maintainers to adopt the framework.
The ability with declaratively specifying the config per `runTest` is nice syntactic sugar.

I will see if I find a solution that keeps the KISS approach by default but allows to opt-in to declarative config solution.

BirdeeHub:
```nix
(runTest "if nix-direnv is disabled then lib/nix-direnv.sh should not exist" (
    let
      wrapper = self.wrappers.direnv.wrap {
        inherit pkgs;
        nix-direnv.enable = false;
      };
    in
    [
      (isDirectory (getDotdir wrapper))
      (notIsFile "${getDotdir wrapper}/lib/nix-direnv.sh")
    ]
  ))
```

^ I mean, they could already do that with literally any of these solutions, this isn't really what I meant.

This is what you were doing with the wrapper2 thing too.

I more meant, should we be allowing someone to pass an EXTRA module per each individual test case. As in, they would pass a module that extends the base one and we handle that somehow.

That being said, I am starting to think that would be too much.

> Maybe I will make runTest able to recieve both a function wrapper -> <listOf Assertion> (where wrapper is pre-configured as in my 2nd draft) but also simply a <listOf Assertions> where they can build their wrapper from scratch using let .. in (as above).

This is not the worst idea I have ever heard, but its still basically just doing `wrapper2`, and if someone only uses `wrapper2` they kinda start to ask "why `wrapper1` then"?

If we only need it for name and `meta.platforms`, should we just have them specify a name for the test, and a list of platforms?

> > Then the question with that would be, should the next test get the updated state of that, or should it receive the original copy? The user should probably be able to specify that?

^ This part of the statement doesn't make sense without the first part I was trying to say.

That being said, I am not opposed to simpler.

I still need to try this out, I haven't tried it yet. I don't always notice what needs to be done unless I try to use it for something myself.

zenoli:
I have the opt-in solution almost ready. I will push it soon. 

> I still need to try this out, I haven't tried it yet. I don't always notice what needs to be done unless I try to use it for something myself.

Yes of course. This would be very valuable. Try it once its ready. We can still strip out the opt-in part if turns out to be unnecessary.

BirdeeHub:
What if instead of passing in a wrapper module to runTest at all, we just do the runCommand templating you have for runTest, and then expose a `mkIfPlatform` kinda function that takes the test (as its second parameter) plus any of: (a module), (evaluated module), (config set of evaluated module) (package) or (list of platform strings) as its first parameter, and then returns the (test drv or null) based on the `meta.platforms` value it has, that they can use at the end to disable it instead of writing out the elem system systems function themselves?

Then they can be in charge of what happens with the module and stuff and we can just focus on making helpers for turning those into tests.

```nix
mkTestIfPlatform = cond: test: let
  platforms = if builtins.isList then cond
    else cond.passthru.configuration.meta.platforms or cond.config.meta.platforms or cond.meta.platforms or (self.lib.evalModule cond).config.meta.platforms;
in
if builtins.elem pkgs.stdenv.hostPlatform.system platforms then test else null;
```

^? something like that?

And then we don't need `runTest` to accept a module at all really? They can just use the module plus the helpers to construct tests however they want, and then pass whatever they have plus their test to that and return the result?

zenoli:
I created a separate PR with the changes:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/446

I've also written a section in CONTRIBUTING.md on its usage:

https://github.com/zenoli/nix-wrapper-modules/blob/test-lib-module-passing-rework/CONTRIBUTING.md#writing-tests

This is more or less my proposal as of now. 
Maybe give me feedback over there once you get around to testing it?

About your inputs in your latest comment:

> What if instead of passing in a wrapper module to runTest at all, we just do the runCommand templating you have for runTest, and then expose a `mkIfPlatform` kinda function that takes the test (as its second parameter) plus any of: (a module), (evaluated module), (config set of evaluated module) (package) or (list of platform strings) as its first parameter, and then returns the (test drv or null) based on the `meta.platforms` value it has, that they can use at the end to disable it instead of writing out the elem system systems function themselves?
> 
> Then they can be in charge of what happens with the module and stuff and we can just focus on making helpers for turning those into tests.
> 
> ```nix
> mkTestIfPlatform = cond: test: let
>   platforms = if builtins.isList then cond
>     else cond.passthru.configuration.meta.platforms or cond.config.meta.platforms or cond.meta.platforms or (self.lib.evalModule cond).config.meta.platforms;
> in
> if builtins.elem pkgs.stdenv.hostPlatform.system platforms then test else null;
> ```
> 
> ^? something like that?
> 
> And then we don't need `runTest` to accept a module at all really? They can just use the module plus the helpers to construct tests however they want, and then pass whatever they have plus their test to that and return the result?

Can you give me a "sketch"  of your ideas in the form of a review comment somewhere here:
https://github.com/zenoli/nix-wrapper-modules/blob/test-lib-module-passing-rework/CONTRIBUTING.md#writing-tests

Just like how it would look like to specify the tests with those ideas - to have a better basis for discussion?

BirdeeHub:
So, I will hopefully get an answer back to you about this today or tomorrow.

I have a slight addition to v2

basically just this for expected instead of regular toJSON.

```nix
  toSanitizedJSON =
    value:
    if builtins.isAttrs value then
      builtins.toJSON (
        lib.mapAttrsRecursive (
          path: v:
          if builtins.isFunction v then
            let
              id = toString (
                builtins.unsafeGetAttrPos (lib.last path) (
                  lib.getAttrFromPath (lib.sublist 0 (builtins.length path - 1) path) value
                )
              );
            in
            "<function${if id != "" then " ${id}" else ""}>"
          else
            v
        ) value
      )
    else
      builtins.toJSON value;
```

---

And I have another direction we could take it much like the first version which didn't accept the wrapper module as an argument, but with some slight differences. But I have been busy the last few days and haven't found the time to give a well thought-through reply.

I hope to get you an example implementation and usage of what I was talking about for that other direction sometime today though. We will see.

I will need to fix https://github.com/BirdeeHub/nix-wrapper-modules/issues/447 and https://github.com/BirdeeHub/nix-wrapper-modules/issues/449 first though.

zenoli:
Thanks for the sanitize fn. I think I know what it does as I yesterday noticed it cannot transform arbitrary nix sets.

zenoli:
> So, I will hopefully get an answer back to you about this today or tomorrow.

You don't need to hurry. I again generalized the v2 approach a lot yesterday to also make it suitable for non-wrapper tests but general ones (like core/types/etc).

Better wait until I pushed this and I will ping you once it ready. 

@BirdeeHub (pinging you now too, to make sure you see it and not waste time reviewing outdated code :-))

That said, the changes are really just a generalization of the current v2: It generalizes the conditional execution for platforms to generic conditions, and the passing of config (which gets transformed to a configured wrapper) to passing generic contexts

ci/test-lib.nix
```diff
@@ -0,0 +1,242 @@
...
+    msg = "File ${path} should not exist";
+  };
+
+  fileContains = file: pattern: {
+    cond = ''grep -q '${pattern}' "${file}"'';
+    msg = "Pattern '${pattern}' not found in ${file}";
+  };
+
+  areEqual = expected: actual: {
+    msg = ''
```
BirdeeHub: Since we test this in nix, we should only write this if it failed. Just in case expected and actual contain a bunch of stuff we have to build.

ci/test-lib.nix
```diff
@@ -0,0 +1,242 @@
...
+      writeShellScript name ''
+        ${testScript}
+      ''
+    else
+      runCommand name { passthru.test = testScript; } ''
+        ${testScript}
+        touch $out
+      '';
+
+  /*
```
BirdeeHub:
```suggestion
  /**
```

^ doc comment syntax (we should generate docs for this, and put contributing docs on the website soon)

We don't need to do that as part of this PR, (unless you want to) but we should fix up the comments in preparation of that.

ci/test-lib.nix
```diff
@@ -0,0 +1,242 @@
...
+        debug = settings.debug or false;
+      } testSet
+    else
+      null;
+
+in
+{
+  inherit test;
+
+  isDirectory = path: {
```
BirdeeHub, line 232: On the note of generating docs for this, can you add `/** */` comments for these too?

ci/test-lib.nix
BirdeeHub: https://github.com/BirdeeHub/nix-wrapper-modules/blob/fd6d1dc5fdfb1aa39c355f4da36773a972a35721/ci/docs/default.nix#L124-L138

^ we should be able to add one of these for CONTRIBUTING.md + the test lib? Maybe stick it on the end after the wrapper modules?

This part can be done in a later PR, although you can do it in this one if you want to, but we should at least prepare the comments in this file for such a thing

BirdeeHub (commented):
I am much happier with this design.

ci/test-lib.nix
```diff
@@ -0,0 +1,242 @@
...
+      _test {
+        inherit name;
+        debug = settings.debug or false;
+      } testSet
+    else
+      null;
+
+in
+{
+  inherit test;
```
BirdeeHub:
```suggestion
  /**
    doc comment here
  */
  test = test;
```

^ nixdoc is dum

BirdeeHub:
> the one thing I did not re-implement from your version is the extend feature. If you want this I will have to study it in more detail. I think it should be possible to add if it is really needed. Just let me know.

^ that's fine. It was kinda overkill and of limited utility. But, `wlib.makeCustomizable` is pretty cool, you should check it out still XD Don't bother figuring out how it works completely, trust me I only just barely managed to understand it when I wrote it.

It calls the thing, and adds the function back to the result of the thing so you can call it again basically, and it merges the args for the calls.

It works like `lib.mkOverrideable` with 2 extra args and it lets you customize how the args get merged from one call to the next, thats all you really need to know to use it XD

well, that, and if you add more on top, you should set `patches` in the settings (second arg) so the previous ones don't remove the new ones. `lib.mkOverrideable` doesnt have a way to solve that, so that is a new concept thing to that function

It could always be added later if we do want it.

ci/test-lib.nix
zenoli: You mean like this?

```nix
  areEqual =
    expected: actual:
    let
      equal = expected == actual;
    in
    {
      cond = if equal then "true" else "false";
      msg =
        if !equal then
          ''
            Expected:
              ${toSanitizedJSON expected}
            but got:
              ${toSanitizedJSON actual}''
        else
          "This should never happen.";
    };
```

They will be evaluated either way (and I guess they would still build stuff) but it would avoid having huge JSONs rendered in the generated script for  a code path that is unreachable.

ci/test-lib.nix
zenoli: Done

zenoli:
> ^ that's fine. It was kinda overkill and of limited utility. But, wlib.makeCustomizable is pretty cool, you should check it out still XD Don't bother figuring out how it works completely, trust me I only just barely managed to understand it when I wrote it.

Oh I did check it out already and yes it is pretty cool. I honestly was just to lazy (and exhausted) to re-implement it because the cleanup turned out to be more work than anticipated and wanted to call it a day. I think we should still be able to add in in a non-breaking way, should we ever have a need for it.

ci/test-lib.nix
zenoli: I added a chapter "Contributing" at the end displaying CONTRIBUTING.md 
I also added the generated docs for tlib under this chapter.

I'm not familiar with nixdoc nor mdbook so I'm not sure if I named all the arguments etc. correctly

I also changed the heading-levels in Contributing.md to make it consistent with the rest (I changed the h1 --> h2 and adjusted the hierarchy)

zenoli:
@BirdeeHub thanks for the review. I implemented you're ideas. Can you check again?

ci/test-lib.nix
BirdeeHub: Yeah thats perfect.

And it won't necessarily build stuff (on success, anyway)

It just checks that the sets are equal.

So, if 2 things got the same exact pkgs set, then it would see that they have the same pkgs set from the same place, and presumably it would only have to evaluate that far?

Whereas if it is printing the whole thing, `pkgs` is an option, and someone could theoretically have something containing it that they want to check...

On failure, whatever, you will see that, and know approximately what is happening. We can maybe improve that behavior eventually, i.e. have it detect nixpkgs and exclude it as well, like we do with functions, but like, at the very least we want success to be fast, and fast does NOT include printing all of nixpkgs into a file.

Another possibly useful thing would be to unsafeDiscardStringContext on the "sanitized" json, maybe we should do that. All the sanitizer does rn is remove functions. If we discarded the context, then it would further ensure we dont build anything spurious.

This is fine for now tho. If we need to we can sanitize more later.

ci/test-lib.nix
BirdeeHub: > I'm not familiar with nixdoc nor mdbook so I'm not sure if I named all the arguments etc. correctly

Fair enough

In the generated docs rn its giving me ```Testing library `tlib documentation``` as the title, so something is wrong with the nixdoc args, but it should be an easy fix, will look into it. All I did was run `nix run ./ci#docs && python3 -m http.server 1337` and look at it so far.

ci/docs/default.nix
```diff
@@ -208,6 +208,23 @@ in
...
+        path = "md/CONTRIBUTING.md";
+        src = ../../CONTRIBUTING.md;
+        subchapters = [
+          {
+            name = "tlib";
+            data = "numbered";
+            path = "tlib.md";
+            src = "${placeholder "out"}/wrappers-lib/tlib.md";
+            build = ''
+              ${pkgs.nixdoc}/bin/nixdoc --category "makeWrapper" --description 'Testing library `tlib documentation' --file '${../test-lib.nix}' --prefix "tlib" >> $out/wrappers-lib/tlib.md
```
BirdeeHub:
```suggestion
              ${pkgs.nixdoc}/bin/nixdoc --category "" --description 'Testing library `tlib` documentation' --file '${../test-lib.nix}' --prefix "tlib" >> $out/wrappers-lib/tlib.md
```

nixdoc arg fixes (again, nixdoc is a bit silly here, but, this looks the best in the generated result)

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+
+      If it is a string, it will be taken as the name of the derivation built
+      by `runCommand`.
+
+      If it is a set, at least one of the keys `name` or `wrapper` need to be set
+      with a string value.
+
+      If you are defining tests for wrappers you should set `settings.wrapper` to the name
+      of the wrapper. This way, the `wrapper.meta.platforms` is considered during test execution,
+      and the test will only be run if the system it is running on is supported.
```
BirdeeHub, line 160:
```suggestion
      and the test will only be run if the system it is running on is supported.

```

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+      If `settings.wrapper` is not set, the test will always be run. This makes sense if you
+      are testing core options or library functions.
```
BirdeeHub, line 162:
```suggestion
      are testing core options or library functions.

```

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+      If only `settings.wrapper` is set, the name will be derived from this value by suffixing it with `-test`.
```
BirdeeHub:
```suggestion
      If only `settings.wrapper` is set, the name will be derived from this value by suffixing it with `-test`.

```

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+      If `settings.name` is set, it will be taken as the name of the derivation.
```
BirdeeHub:
```suggestion
      If `settings.name` is set, it will be taken as the name of the derivation.

```

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+      If both are set, `settings.name` takes precedence.
+
+      There is also an optional boolean option `settings.debug` (default = `false`).
```
BirdeeHub, line 167:
```suggestion
      There is also an optional boolean option `settings.debug` (default = `false`).

```

And maybe an extra newline here too?

ci/test-lib.nix
BirdeeHub: They get all scrunched up when rendered. It follows markdown rules.

We should either add an extra newline between the cases, or make them bullet points (although that might mess with it being "in the list" thing it has going on, so maybe just newlines)

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+    str:
+    let
+      lines = lib.splitString "\n" str;
+      result = "$'${lib.concatStringsSep "\\n" lines}'";
+    in
+    if (lib.length lines) == 1 then str else result;
+
+  toSanitizedJSON =
+    value:
+    if builtins.isAttrs value then
```
BirdeeHub, line 36:
```suggestion
    builtins.unsafeDiscardStringContext (if builtins.isAttrs value then
```

We might want to do this, just in case? IDK, it probably doesn't matter, and we can always add this later (dont forget to add the closing `)`, I didnt put it in the suggestion because its like 20 lines down)

BirdeeHub: Lets only do this if it turns out that it is an actual problem actually.

BirdeeHub:
Just fixes to the nixdoc args and some (minor) formatting nitpicks left I think!

`nix run ./ci#docs -- ./_site && cd _site && python3 -m http.server 1337` it and go to `https://localhost:1337` to check it out, it is pretty fast to build tbh.

The mdbook module is pretty nice :smile_cat:

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+in
+{
+  /**
+    Takes tests provided as an attrs in `testSet`, renders it to a bash script,
+    and executes the rendered script using `pkgs.runCommand`.
+
+    If the command runs successfully, the test set passes. Otherwise, a detailed
+    error message describing the failed assertion is shown in the logs.
+
+    # Type
```
BirdeeHub, line 132:
```suggestion
    ## Type
```

Everywhere you see 1 `#` in this file needs `##`. `nixdoc` will add only 1 more `#` to them, so 1 `#` means it is rendered to `##` but we want them to be rendered to `###`

Otherwise we spam the sidebar with Type and Argument shortcuts which are totally unnecessary. Its fine it just looks messy.

I know, the other ones already spam the sidebar with all the options for all the modules in each wrapper, I hope to eventually make that better too.

BirdeeHub: Ok, so, I tried it out, actually, nevermind, leave them like this, I can't seem to get the sidebar to shut up about them even with the extra `#` and idk why

Ill handle that when I get around to docs next

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+        else if (wrapperName != null) && (lib.hasAttr wrapperName self.wrappers) then
+          self.wrappers.${wrapperName}.meta.platforms
+        else
+          throw ''
+            Invalid argument for `test`.
+            The provided wrapper '${wrapperName}' was not found.
+            Available wrappers are:
+
+            ${lib.toSanitizedJSON (lib.attrNames self.wrappers)}
+          '';
```
BirdeeHub, line 207:
```suggestion
          '';
          enabled = (settings.enable or true) && (platforms == null || (platforms != null && builtins.elem stdenv.hostPlatform.system platforms))
```

Last actual addition, I want an enable option (defaulting to true) so I can `enable = false` so I can more easily disable stuff to figure out what the heck has a build failure this time on update (because sometimes I have a good guess or 2)

ci/test-lib.nix
```diff
@@ -0,0 +1,325 @@
...
+    in
+    if
+      platforms == null || (platforms != null && builtins.elem stdenv.hostPlatform.system platforms)
```
BirdeeHub:
```suggestion
      enabled
```

https://github.com/BirdeeHub/nix-wrapper-modules/pull/439/changes#r3153941309

BirdeeHub:
wait what changed to make it fail?

BirdeeHub:
oh the fish module got added and the checks for it need `...`

BirdeeHub:
I added ... to the fish tests now :) You might have to rebase on top of that somehow

BirdeeHub:
Also I apparently forgot a semicolon in my suggestion for enabled (which I did not test btw)

zenoli:
@BirdeeHub don't worry, I'm still updating stuff :-)

zenoli:
About the enable option:

If I run a wrapper test on an unsupported platform, the test should not be run - independent of the enable flag, right?

BirdeeHub:
> About the enable option:
> 
> If I run a wrapper test on an unsupported platform, the test should not be run - independent of the enable flag, right?

If I do `enable = false` I want it to be disabled. I don't care what platforms. Just disabled.

Otherwise, it goes based on platform, or is enabled if not a wrapper test.

So, in this case, enable does nothing unless it is false, in which case, it disables the test for sure.

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/431/changes

Also ^ this is why the enable option needs to disable and why I want it, for context. (basically, claude-code...)

ci/test-lib.nix
zenoli: I went with bullet points, it renders fine imo: 

Image

There is an extra newline after the bullet points, idk why the paragraph after the list renders so close to the list but I guess that is a markdown render thing.

zenoli:
Ok, I think I'm done with the changes.

Rebased & squashed + removed direnv module.

Anything else @BirdeeHub ?

BirdeeHub:
No I think that is it! I think you did a good job with these!

Will wait for the mac tests to run.

Github is being crazy today, it has been very hard for me to use it I have had to pull every single thing locally.

zenoli:
Thanks! And thanks for the various inputs. Took more work than initially anticipated, but learned a lot about nix along the way :-)

Mac tests are through.

zenoli:
> Github is being crazy today, it has been very hard for me to use it I have had to pull every single thing locally.

FYI: https://www.githubstatus.com/incidents/x69zbgdyfzg0

PR #441, opened 2026-04-16
feat(wrapperModules.tmux): improve plugins option

BirdeeHub:
now uses a spec type rather than `lib.types.oneOf`. Main field still is `plugin`, to avoid breaking existing configurations. This also improves support for tmux plugins from sources outside of nixpkgs.

The reason it did not already use a spec type for this, is that this module was added before the spec type was created.


BirdeeHub:
Also, yes I am still going to fix up yazi, this was easy and I just noticed it

PR #444, opened 2026-04-16
fix(wrapperModules.yazi): provide some way of installing plugins

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/issues/395 ?

This is greatly improved from its previous state of having no way to install plugins other than doing this yourself.

Im not super happy about this. I am finding out they didn't bother to do an RTP, and they didn't allow you to change the dir that ya uses. If they used an RTP for plugins rather than just a single path, and then let us set where the ya download and lockfile locations are, it would be fantastic but that is not available. I looked really hard but there is nothing like that.

So, unless we do what we did for noctalia-shell, where we also provide options for copying it out of the store at runtime, we can only provide plugins via nix now.

That being said, theres no build steps or anything and nix is good at fetching stuff, and there are already yazi plugins in nixpkgs, so it isn't that big of a deal. So I am thinking for now this might be fine, and maybe those other options could be added someday if necessary.

It also is much better than not having any way to provide them at all without doing this yourself, which is the current situation.

We really need services options and/or bubblewrap for solving these "requires writable config for some features" problems better.

### DIFF
--- a/wrapperModules/y/yazi/module.nix
+++ b/wrapperModules/y/yazi/module.nix
@@ -447,6 +447,49 @@ in
     '';
   };
 
+  options.plugins = lib.mkOption {
+    type = lib.types.attrsOf (lib.types.nullOr wlib.types.stringable);
+    default = { };
+    description = ''
+      An attribute set of plugin names and their paths
+    '';
+    example = lib.literalMD ''
+      ```nix
+      with pkgs.yaziPlugins; {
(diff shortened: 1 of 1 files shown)


BirdeeHub:
So, if ya ends up important enough that we have to run the copy thing we did for noctalia-shell for this, it should be doable in a fairly non-breaking manner.

So, on that note, I think this is probably fine?

PR #452, opened 2026-04-21
WIP feat(tests): add test utils v3

BirdeeHub:
@zenoli

Thoughts on this instead?

Im worried about trying to manage the wrapper module for the user that it would lead to people making a million drvs, and also, it adds complexity points to handle that that could be better spent making them nest somewhat arbitrarily?

Here, we only accept it to check platform

I left the direnv wrapper module in as an example for you to review at least 1 usage of it

I will remove the direnv wrapper module from this PR before merging it if we go with this design.

It also takes an arbitrarily nested set of lists of conditions too

And you can call .extend and add more tests to it.

I still need to add docs for it and stuff. But what do you think? Flexible enough?

### DIFF
--- a/ci/test-lib.nix
+++ b/ci/test-lib.nix
@@ -1,0 +1,193 @@
+{
+  self,
+  lib,
+  runCommand,
+  pkgs,
+  ...
+}:
+let
+  wlib = self.lib;
+  inherit (lib) isList;
+  toSanitizedJSON =
+    value:
+    if builtins.isAttrs value then
+      builtins.toJSON (
+        lib.mapAttrsRecursive (
+          path: v:
+          if builtins.isFunction v then
+            let
+              res = builtins.unsafeGetAttrPos (lib.last path) (
+                lib.getAttrFromPath (lib.sublist 0 (builtins.length path - 1) path) value
+              );
+            in
+            "<lambda${
+              if builtins.isAttrs res then
+                ":${res.file or ""}:${toString (res.line or "")}:${toString (res.column or "")}"
+              else
+                ""
+            }>"
+          else
+            v
+        ) value
+      )
+    else
+      builtins.toJSON value;
+
+  recursiveUpdateWithMerging = wlib.recursiveMergeUntil {
+    until =
+      path: lh: rh:
+      !(wlib.isNonDrvAttrs lh) || !(wlib.isNonDrvAttrs rh);
+    merge =
+      path: left: right:
+      if isList left && isList right then left ++ right else right;
+  };
+
+  enableForSystem =
+    system: value: test:
+    let
+      platforms =
+        if builtins.isList value then
+          value
+        else
+          value.passthru.configuration.meta.platforms or value.config.meta.platforms or value.meta.platforms
+            or (wlib.evalModule value).config.meta.platforms;
+    in
+    if builtins.elem system platforms then test else null;
+
+  test = message: condition: { inherit message condition; };
+in
+{
+  inherit enableForSystem toSanitizedJSON test;
+  enableBySystem = enableForSystem pkgs.stdenv.hostPlatform.system;
+
+  isDirectory = path: test "No such directory ${path}" ''[ -d "${path}" ]'';
+
+  isFile = path: test "No such file ${path}" ''[ -f "${path}" ]'';
+
+  notIsFile = path: test "File ${path} should not exist" ''[ ! -f "${path}" ]'';
+
+  fileContains =
+    file: pattern: test "Pattern '${pattern}' not found in ${file}" ''grep -q '${pattern}' "${file}"'';
+
+  areEqual =
+    expected: actual:
+    test (
+      if expected == actual then
+        "areEqual"
+      else
+        lib.escapeShellArg "Expected:\n${toSanitizedJSON expected}\nbut got:\n${toSanitizedJSON actual}"
+    ) (if expected == actual then "true" else "false");
+
+  /**
+    first argument is a string.
+    It is the name of the test derivation,
+    and default warning message name.
+
+    Second argument is of type `cond` as specified below
+
+    ```nix
+    cond = let
+      assertion = { condition :: cond, message :: str };
+    in str | assertion | [ (str | assertion) ] | (attrsOf cond);
+    ```
+
+    TODO: show usage
+
+    you can call .extend on the result to add more items
+    It is the same, but without the first name argument
+  */
+  mkTestDrv =
+    name:
+    let
+      idnt = len: "\n" + wlib.repeatStr "  " len;
+      createAssertion =
+        prefix:
+        { condition, message }:
+        let
+          loc = prefix ++ [ message ];
+          lvl = builtins.length loc;
+          renderCond =
+            first: c:
+            if lib.isStringLike c then
+              c
+            else if builtins.isList c then
+              let
+                conds = builtins.filter (v: v.cond or "" != [ ] && v.cond or v != "") c;
+              in
+              if conds == [ ] then
+                ""
+              else
+                "( " + lib.concatMapStringsSep " ) && ( " (renderCond false) conds + " )"
+            else if !first && builtins.isAttrs c then
+              createAssertion loc c
+            else if first && builtins.isAttrs c then
+              helper loc c
+            else
+              throw "Invalid condition type in assertion, received ${toSanitizedJSON c}";
+        in
+        if condition == [ ] then
+          ""
+        else
+          "(${idnt (lvl + 1)}${renderCond true condition}${idnt lvl}) || (echo 'failing test at:' ${lib.escapeShellArg "${lib.options.showOption prefix}:"} ${message} >&2; return 1)";
+      helper =
+        prefix: assertions:
+        let
+          lvl = builtins.length prefix;
+        in
+        if
+          builtins.isAttrs assertions
+          && (
+            lib.removeAttrs assertions [
+              "condition"
+              "message"
+            ] != { }
+          )
+        then
+          lib.concatMapAttrsStringSep "${idnt lvl}" (
+            n: v:
+            if v == [ ] then
+              ""
+            else
+              let
+                loc = prefix ++ [ n ];
+                text = helper loc v;
+              in
+              "(${idnt (lvl + 1)}${text}${idnt lvl}) || (echo 'failing test at:' ${lib.escapeShellArg "${lib.options.showOption prefix}:"} ${n} >&2 && ${
+                if builtins.length loc == 1 then "exit" else "return"
+              } 1)"
+          ) assertions
+        else
+          "(${idnt (lvl + 1)}"
+          + lib.concatMapStringsSep "${idnt lvl}) && (${idnt (lvl + 1)}" (
+            a: if lib.isStringLike a then "${a}" else createAssertion prefix a
+          ) (lib.toList assertions)
+          + "${idnt lvl})";
+      normArgs =
+        args:
+        if builtins.isList args then
+          { ${name} = args; }
+        else if lib.isStringLike args then
+          { ${name} = [ args ]; }
+        else
+          args;
+    in
+    wlib.makeCustomizable "extend"
+      {
+        mergeArgs =
+          og: new:
+          if lib.isFunction new then
+            new (normArgs og)
+          else
+            recursiveUpdateWithMerging (normArgs og) (normArgs new);
+      }
+      (
+        assertions:
+        let
+          finalText = helper [ ] (normArgs assertions);
+        in
+        runCommand name { passthru.test = finalText; } ''
+          ${finalText}
+          mkdir -p $out
+        ''
+      );
+}
--- a/lib/lib.nix
+++ b/lib/lib.nix
@@ -607,6 +607,103 @@ in
       (builtins.attrNames attrs);
 
   /**
+    checks for an attribute set that is not string-like and returns true if it is
+  */
+  isNonDrvAttrs = v: builtins.isAttrs v && !lib.isStringLike v;
+
+  /**
+    A generalized recursive attribute set merge with user-defined stopping and
+    conflict resolution behavior.
+
+    This function is similar to `lib.recursiveUpdateUntil`, but extends it by
+    allowing a custom `merge` function to decide how values are merged when
+    recursion stops.
+
+    The merge walks both attribute sets (`lhs` and `rhs`) in parallel:
+
+    - If only one side defines a value at a given path, that value is used.
+    - If both sides define a value:
+      - `until path lhs rhs` is evaluated.
+        - If it returns `true`, recursion stops at this path and
+          `merge path lhs rhs` determines the result.
+        - If it returns `false`, recursion continues into the values (which are
+          expected to be attribute sets).
+
+    Default behavior:
+    - Recursion continues only while both values are attribute sets
+      (`builtins.isAttrs`).
+    - When recursion stops, the right-hand side (`rhs`) replaces the left-hand side.
+
+    Parameters:
+      - `until` :: `[string]` -> `any` -> `any` -> `bool`:
+        - Predicate that determines when to stop descending into values.
+        - Arguments:
+          - `path`: attribute path from the root (e.g. `[ "a" "b" "c" ]`)
+          - `lhs`: value from the left-hand side
+          - `rhs`: value from the right-hand side
+
+      - `merge` :: `[string]` -> `any` -> `any` -> `bool`:
+        - Function used to resolve a conflict when `until` returns true.
+          Receives the same arguments as `until`.
+
+    Arguments:
+      - `lhs` :: `attrset`:
+        Base attribute set.
+
+      - `rhs` :: `attrset`:
+        Overlay attribute set (takes precedence by default).
+
+    Returns:
+      - `attrset`:
+        The recursively merged result.
+
+    Notes:
+    - Merge is right-biased unless `merge` overrides that behavior.
+    - Non-attribute values are treated as leaves and resolved via `merge`.
+    - The `path` argument allows context-aware merging strategies.
+
+    Example (concatenate lists instead of replacing them):
+    ```nix
+      recursiveMergeUntil {
+        merge = path: l: r:
+          if builtins.isList l && builtins.isList r then l ++ r else r;
+      } { a = [1]; } { a = [2]; }
+
+      => { a = [1 2]; }
+    ```
+  */
+  recursiveMergeUntil =
+    # source: https://github.com/BirdeeHub/nixCats-nvim/blob/da76c45b33d589836946bb566bd91df4cd3cfb09/utils/lib.nix#L57
+    {
+      until ? (
+        path: lh: rh:
+        !(builtins.isAttrs lh) || !(builtins.isAttrs rh)
+      ),
+      merge ? (
+        path: l: r:
+        r
+      ),
+    }:
+    lhs: rhs:
+    let
+      f =
+        attrPath:
+        builtins.zipAttrsWith (
+          n: values:
+          let
+            here = attrPath ++ [ n ];
+          in
+          if builtins.length values == 1 then
+            builtins.head values
+          else if until here (builtins.elemAt values 1) (builtins.head values) then
+            merge here (builtins.elemAt values 1) (builtins.head values)
+          else
+            f here values
+        );
+    in
+    f [ ] [ rhs lhs ];
+
+  /**
     `genStr` :: `string` -> `int` -> `string`
 
     or
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,11 +155,7 @@ Sometimes it is not easily possible to run the program within a derivation, in t
 Example:
 
 ```nix
-{
-  pkgs,
-  runCommand,
-  self,
-}:
+{ pkgs, self, tlib, ... }:
 let
   gitWrapped = self.wrappers.git.wrap {
     inherit pkgs;
@@ -170,12 +166,18 @@ let
       };
     };
   };
+  inherit (tlib) it fileContains test enableBySystem ;
 in
-runCommand "git-test" { } ''
-  "${gitWrapped}/bin/git" config user.name | grep -q "Test User"
-  "${gitWrapped}/bin/git" config user.email | grep -q "test@example.com"
-  touch $out
-''
+# git uses meta.platforms = lib.platforms.all, so this will be enabled on all platforms
+enableBySystem self.wrappers.git (
+  # define a test derivation using the testing library!
+  test "git-test" [
+    # define tests for your test derivation using `it` and the other helpers
+    (it "has test user" ''"${gitWrapped}/bin/git" config user.name | grep -q "Test User"'')
+    (it "has test email" ''"${gitWrapped}/bin/git" config user.email | grep -q "test@example.com"'')
+    (fileContains (gitWrapped.configuration.constructFiles.gitconfig.outPath) "test@example\.com")
+  ]
+)
 ```
 
 If your module declares a list of valid platforms via its `meta.platforms` option, you should disable your test on the relevant platforms like so:
--- a/ci/flake.nix
+++ b/ci/flake.nix
@@ -20,6 +20,7 @@
             inherit system;
             config.allowUnfree = true;
           };
+          tlib = pkgs.callPackage ./test-lib.nix { inherit self; };
 
           # Load checks from ci/checks/ directory
           coreAndCiChecks = lib.pipe ./checks [
@@ -43,7 +44,7 @@
                     name = "${prefix}-${name}";
                     inherit value;
                   };
-                  result = pkgs.callPackage value { inherit self; };
+                  result = pkgs.callPackage value { inherit self tlib; };
                 in
                 if result == null then
                   [ ]
--- a/wrapperModules/d/direnv/check.nix
+++ b/wrapperModules/d/direnv/check.nix
@@ -1,0 +1,57 @@
+{
+  lib,
+  callPackage,
+  tlib,
+  self,
+  ...
+}@args:
+let
+  inherit (tlib) test enableBySystem areEqual;
+in
+builtins.mapAttrs (_: v: enableBySystem self.wrappers.direnv v) (
+  lib.fix (self: {
+    "1" = callPackage ./check1.nix args;
+    "2" = callPackage ./check2.nix args;
+    "3" = callPackage ./check3.nix args;
+    "4" = self."1".extend {
+      moretests = {
+        "another test" = "echo hello";
+        "another test2" = "echo world";
+      };
+      outer = {
+        inner = test "should have echoed 'hi'" "echo hi";
+        inner2 = test "should have echoed 'hi again'" { something = "echo 'hi again'"; };
+        inner3 = test "should have echoed 'woot'" (test "should have echoed 'woot'" "echo 'woot woot'");
+      };
+      zzz = [
+        (test "another test" {
+          like.this.too = "echo wtf1";
+        })
+      ];
+    };
+    "5" = self."4".extend [
+      (test "well... so... it" {
+        can.be.nested.like.this.too = [
+          (areEqual 2 2)
+          (areEqual
+            {
+              a = 1;
+              b = 2;
+              c = 3;
+              inherit enableBySystem;
+            }
+            {
+              a = 1;
+              b = 2;
+              c = 3;
+              inherit enableBySystem;
+            }
+          )
+        ];
+      })
+      (test "is" {
+        a.bit.strange = "echo 'but it completed the loop nicely'";
+      })
+    ];
+  })
+)
--- a/wrapperModules/d/direnv/check1.nix
+++ b/wrapperModules/d/direnv/check1.nix
@@ -1,0 +1,157 @@
+{
+  pkgs,
+  self,
+  tlib,
+  ...
+}:
+
+let
+  inherit (tlib)
+    fileContains
+    isDirectory
+    isFile
+    notIsFile
+    ;
+  getDotdir = wrapper: "${wrapper}/${wrapper.configuration.configDirname}";
+in
+tlib.mkTestDrv "direnv-tests" [
+  {
+    message = "wrapper should output correct version";
+    condition =
+      let
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+        };
+      in
+      ''
+        "${wrapper}/bin/direnv" --version | grep -q "${wrapper.version}"
+      '';
+  }
+  {
+    message = "if nix-direnv is enabled then lib/nix-direnv.sh should exists";
+    condition =
+      let
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          nix-direnv.enable = true;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
+        (isFile "${getDotdir wrapper}/lib/nix-direnv.sh")
+      ];
+  }
+  {
+    message = "if nix-direnv is disabled then lib/nix-direnv.sh should not exist";
+    condition =
+      let
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          nix-direnv.enable = false;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
+        (notIsFile "${getDotdir wrapper}/lib/nix-direnv.sh")
+      ];
+  }
+  {
+    message = "if mise is enabled then lib/mise.sh should exists";
+    condition =
+      let
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          mise.enable = true;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
+        (isFile "${getDotdir wrapper}/lib/mise.sh")
+      ];
+  }
+  {
+    message = "if mise is disabled then lib/mise.sh should not exist";
+    condition =
+      let
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          mise.enable = false;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
+        (notIsFile "${getDotdir wrapper}/lib/mise.sh")
+      ];
+  }
+  {
+    message = "if a lib-script is set then it should be generated";
+    condition =
+      let
+        libScriptFile = "${getDotdir wrapper}/lib/foo.sh";
+        libScriptContent = "echo foo";
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          lib."foo.sh" = libScriptContent;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
+        (isFile libScriptFile)
+        (fileContains libScriptFile libScriptContent)
+      ];
+  }
+  {
+    message = "if silent mode is enabled then log settings should be set";
+    condition =
+      let
+        direnvTomlFile = "${getDotdir wrapper}/direnv.toml";
+        wrapper = self.wrappers.direnv.wrap {
+          inherit pkgs;
+          silent = true;
+        };
+      in
+      [
+        (isDirectory (getDotdir wrapper))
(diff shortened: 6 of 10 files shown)


zenoli:
I will first finish up my vision for v2 to get at least some kind of "feeling of achievement" because I am like 99 percent there.

It is probably overengineered and I have a slight fix to v1 that I also planned to try. Once I have that out of the way I can take a step back and compare all the versions side by side.

What I quite liked about about my approach was the "dsl" like feel of the syntax by using explicit `runTests/runTest` blocks (as opposed to your approach of making this implicit by using plain sets as the tests).

Maybe my approach is non-standard and yours is more the "nix way" of doing it. It is probably very subjective but of course I am biased to the`runTest/runTests` syntax.

Nesting conditions is definitely nice. 

---

From the discussion so far, I would make sense to define some concepts:

In my variants (V1, V2) I have these concepts:

- "test suite": Collection of related tests. Marked by `runTests`
- "test": Member of a test suite, marked by `runTest`. Has a name and a list of assertions.
- "assertions": list of conditions (bash commands) with error messages that are defined in a test.

In your approach (V3):

- "test-suite": is marked by tlib.define
- "test": the recursive/nested structure inside you pass to tlib.define. you use "msg" here for the "name", and "cond" for the "list of assertions", I use the same keys but for assertions and not tests, which might cause confusion when we discuss the different versions, so I wanted to point that out :-)
- "assertions": Same as V1, V2.

BirdeeHub:
You can define `runTest = msg: cond: { inherit msg cond; }` and it should feel the same within `define`, no?

We can name it `it` and then you can `define "sometests" [ (it "does something" "echo hello") ]` and have [busted](https://lunarmodules.github.io/busted/) in nix haha

Edit: I added `it` example to `direnv/check3.nix`

BirdeeHub:
Basically, I tried your v2 one, and I found myself wanting the wrapper to not be wrapped into it.

So I looked at your v1 one, and realized that there was actually a lot to go on in terms of design that was good, but that it could use nesting, and a few other things that could be cool.

zenoli:
"it" syntax looks promising.

Will get to have a close look at everythin in about 10h.

BirdeeHub:
I had to steal a function from nixCats to get the extend working properly with the nested lists.

Honestly it is actually a pretty useful function I will add it to wlib

BirdeeHub:
```
nix-repl> :lf ./ci
nix-repl> :p checks.x86_64-linux.wrapperModule-direnv-1.test
nix-repl> :p checks.x86_64-linux.wrapperModule-direnv-2.test
```

^ You can look at the generated test easily like this.

lib/lib.nix
```diff
@@ -606,6 +606,103 @@ in
...
+
+    Parameters:
+      - `until` :: `[string]` -> `any` -> `any` -> `bool`:
+        - Predicate that determines when to stop descending into values.
+        - Arguments:
+          - `path`: attribute path from the root (e.g. `[ "a" "b" "c" ]`)
+          - `lhs`: value from the left-hand side
+          - `rhs`: value from the right-hand side
+
+      - `pick` :: `[string]` -> `any` -> `any` -> `bool`:
```
zenoli:
```suggestion
      - `pick` :: `[string]` -> `any` -> `any` -> `any`:
```

lib/lib.nix
```diff
@@ -606,6 +606,103 @@ in
...
+    ```
+  */
+  recursiveMergeUntil =
+    # source: https://github.com/BirdeeHub/nixCats-nvim/blob/da76c45b33d589836946bb566bd91df4cd3cfb09/utils/lib.nix#L57
+    {
+      until ? (
+        path: lh: rh:
+        !(builtins.isAttrs lh) || !(builtins.isAttrs rh)
+      ),
+      pick ? (
```
zenoli: Wouldn't `merge` or `mergeFn` be more adequate here?
```suggestion
      mergeFn ? (
```

lib/lib.nix
zenoli: Out of scope but `repeatStr` would be more accurate:

```suggestion
  repeatStr = str: num: builtins.concatStringsSep "" (builtins.genList (_: str) num);
```

zenoli:
Should this be supported too?
```nix
tlib.define "my-testsuite" {
  outer = {
    inner = { cond = "return 1"; msg = "should have echoed 'hi'"; };
  };
}
```

Currently I need to wrap the assertions in a list for it to work:
```nix
tlib.define "my-testsuite" {
  outer = {
    inner = [{ cond = "return 1"; msg = "should have echoed 'hi'"; }];
  };
}
```

Running the first snippet gives me a confusing output:

```
error: Cannot build '/nix/store/5gb064hadfnys6yb06dlywxqaba4bir0-my-testsuite.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/vbygjf2f2390al48mx0fp2pqzk6rrwx4-my-testsuite
       Last 5 log lines:
       > test `outer.inner.cond` failed
       > /build/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90: line 10: should: command not found  # <-- msg is interpreted as `cond`
       > test `outer.inner.msg` failed
       > test `outer.inner` failed
       > test `outer` failed
       For full logs, run:
         nix log /nix/store/5gb064hadfnys6yb06dlywxqaba4bir0-my-testsuite.drv
```

BirdeeHub:
I wanted to but it got waaaaaay too confusing to figure out what is a set and what is an assertion anymore and I couldn't get it to work well enough.

It would also make it so that you can't call something `somename.cond.idk = "something";`

I also had to figure out how to merge the differing types on extend

I decided simpler would be better.

Maybe I can give it a better error tho? I'll try.

Personally I would either just do the list, or I would:

```nix
tlib.define "my-testsuite" {
  outer.inner."should have echoed 'hi'" = "return 1";
}
```

The difference is that one says the full path to where it is and the other just prints the message, but in terms of "what failed and why" the utility is the same

Testing only needs so much nesting anyway, past 2-3 levels the usefulness drops off hard. Especially because we still can't filter tests within a single drv (or, well, we can with the function form of `.extend` in the repl or an eval command technically, but we would be building a different drv)



I expect most to choose either a set of lists, or a list of assertions with strings or lists of assertions or strings in them, like one of the examples in the direnv module

Also, to be fair, it is pretty clear that both cond and msg are being interpreted as tests here

```
> test `outer.inner.cond` failed
> /build/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90: line 10: should: command not found
> test `outer.inner.msg` failed
> test `outer.inner` failed
> test `outer` failed
```

Also, that error means it makes it all the way to the drv and runs. addErrorContext I don't think can help me there?

ci/test-lib.nix
```diff
@@ -0,0 +1,164 @@
...
+  fileContains = file: pattern: {
+    cond = ''grep -q '${pattern}' "${file}"'';
+    msg = "Pattern '${pattern}' not found in ${file}";
+  };
+
+  # accepts an arbitrarily nested set of strings or lists of assertions
+  # lists of assertions are lists of { cond, msg } sets.
+  # cond may be a string, or a list of strings or assertions, (or the set form all over again).
+  # you can call .extend on the result to add more items
+  # It is the same, but without the first name argument
```
zenoli: Even after reading this multiple times, I'm still confused as to what is allowed and what not.

> lists of assertions are lists of { cond, msg } sets.

So... `assertion := { cond, msg }`?

> cond may be a string, or a list of strings or assertions, 

So `cond := str | [str] | [assertion]`

> (or the set form all over again)

So even: `cond := str | [str] | [assertion] | attrsOf cond`

Is this the full grammar?

If yes, maybe define it formally to make it more explicit.

zenoli (commented):
I tested it and it works.

The whole `define/helper` recursion thing is hard to follow.
Don't know if they need to be broken up a bit or I am just too dumb.
Will try again tomorrow.

The "it" syntax makes it cool to read, but seeing what it really is, makes it also a bit silly to introduce indirection just for this. 

If we decide to use this "it" indirection, I would somehow want to force maintainers to use it, just to have consistently looking tests throughout the the repo. If not, then we should make it very transparent how the grammar for composing conditions/assertions works. It enables different flavors of how to structure tests and people are free to form them as they like - with the disadvantage of making them less consistent to read (which was what I was working very hard towards it in my versions, so still a bit sad to see them vaporize).

ci/test-lib.nix
BirdeeHub: Yes that would make it clearer

And I think it should be this, right?

`cond = str | [ (str | assertion) ] | attrsOf cond`

`assertion = { cond :: cond, msg :: str, ... }`

> with the disadvantage of making them less consistent to read (which was what I was working very hard towards it in my versions, so still a bit sad to see them vaporize)

I actually didn't notice that was what you were going for but it makes more sense now that I look at it again. Sorry about that. Ill take a look again and see if there is somewhere in the middle

My major motivation on changing it was just that I didn't like that wrapper was getting "wrapped" into it. (heh)

What I had assumed was the major motivator for proposing the library was that it was hard to consistently define multiple case tests within 1 drv in a way that had a printout that tells you which one failed precisely in an easy to locate manner, and possibly tell you why.

> If we decide to use this "it" indirection, I would somehow want to force maintainers to use it, just to have consistently looking tests throughout the the repo.

Hmmmmmm

I think I know how

`{ condition, message }`

instead of

`{ cond, msg, ... }`

That would enforce that when people do the list form they wanna make and use a verb/phrase for it (to match the other ones like `fileContains` and `notIsFile` etc...)

zenoli: > cond = str | [ (str | assertion) ] | attrsOf cond

This would imply that lists can be heterogeneous? If that's the case the yes. I all elements have to be either `str` or `assertion`, then it would have to be as I wrote it.

zenoli: > Hmmmmmm
> 
> I think I know how
> 
> `{ condition, message }`
> 
> instead of
> 
> `{ cond, msg, ... }`
> 
> That would enforce that when people do the list form they wanna make and use a verb/phrase for it (to match the other ones like `fileContains` and `notIsFile` etc...)

I didn't get that. Are you still referring to the `it = cond: msg ...` function?

Also I don't think the `busted` analogy works here. In `busted` "it" is only used in the "leaves" where the final assertions happen. They use "describe" to group tests into behavior.

```
# "it" syntax would probably look like this...
  (it "should to this" [
    (it "should also do this" [ 
      (it "should also do this" [ assertions ])
    ])
    (it "should also do this")
  ])

vs this is how busted groups them

  (describe "behavior" [
    (describe "sub behavior" [ 
      (it "should also do this" [ assertions ])
    ])
  ])
```

BirdeeHub: Right

But we wanted to inforce they use `it`.

So we do NOT want them doing

```
[
  { cond = ""; msg = ""; }
]
```

So, we make it more cumbersome to do so

```
[
  { condition = ""; message = ""; }
]
```

^ nobody will use this form without a helper (or at least much fewer) when compared to `[ (it "" "") ]`

That cuts down on some of the variation, and when people do still do it this way, it is more obvious and thus easier to read.

> This would imply that lists can be heterogeneous? If that's the case the yes. I all elements have to be either str or assertion, then it would have to be as I wrote it.

yes they are heterogeneous. They may contain strings or assertions.

```nix
cond = let
    assertion = { condition :: cond, message :: str };
in (str | [ (str | assertion) ] | (attrsOf cond))
```

lib/lib.nix
BirdeeHub: `action`, or `merge`?

BirdeeHub: ehh yeah merge is fine

ci/test-lib.nix
BirdeeHub: > Also I don't think the busted analogy works here. In busted "it" is only used in the "leaves" where the final assertions happen. They use "describe" to group tests into behavior.

Yeah I would agree I just chose the same names.

We could do `test` and `ok`, but that felt even worse.

`describe` and `it` were so non-descriptive that I chose them.

Are there names which either mean less, or mean what we need them to mean, that are similarly short? `it` is super nice.

Maybe `test` and `it` ? Just maximum vague? or `mkTests` and `test` I guess is like ok, but I don't want to go longer than 4 letters on that second one

BirdeeHub: one thing I could do somewhat easily to make it less confusing would be

```nix
cond = let
    assertion = { condition :: str | [ (str | assertion) ], message :: str };
in (str | [ (str | assertion) ] | (attrsOf cond))
```

But that wouldnt solve allowing `{ a.b = it "" ""; }` instead of just `{ a.b = [ (it "" "") ]; }` either.

It would just stop you from doing the nesting kind from within an assertion for uniformity's sake

It just looked like an obvious loop to finish, but I can't see many people using that particular feature of it either.

Do you think this would be helpful to limit for readability? Or is the problem that we allow both a set and a list at all?

> But that wouldnt solve allowing `{ a.b = it "" ""; }` instead of just `{ a.b = [ (it "" "") ]; }` either.

While it would not be that hard to check when doing the nested form that this one has only `condition` and `message` in it, and treat it like an assertion. However, we also are merging them, and this means we have to extend the merge thing. And then it also means `somename.condition.something` is not legal (wait....does it? it actually might not?? We might be able to do this yes, but only if we don't remove the further nesting from condition.)

BirdeeHub:
Also how do I give you co authorship of the commit if we go with this one. Im going to look that up

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/452#issuecomment-4291353156

^ this is now supported

The type is now

```nix
    cond = let
      assertion = { condition :: cond, message :: str };
    in str | assertion | [ (str | assertion) ] | (attrsOf cond);
```

So now we need to figure out what to name `define` `it` and `drv.extend` `it`

We could always just to `test` `it` and `drv.test` `it`

Or `test` `it` and `drv.extend` `it`

BirdeeHub:
I kept `it`

I changed `define` to `test` to lessen busted analogies.

lib/lib.nix
BirdeeHub: I could make it optionally accept a function instead of a string and then it is well named again? because then it would work like `builtins.genList`

BirdeeHub:
```nix
genStr = str: num: builtins.concatStringsSep "" (builtins.genList (lib.toFunction str) num);
```

BirdeeHub:
```nix
  /**
    `genStr` :: `string` -> `int` -> `string`

    `genStr` :: (`int`: `string`) -> `int` -> `string`

    Generates a string by repeating the input string the specified number of times,
    or by calling the provided function with the index the specified number of times.
  */
  genStr = str: num: builtins.concatStringsSep "" (builtins.genList (lib.toFunction str) num);
```

^ is this a good idea? or should we not add another check to a hot path function.

Or should we make a separate repeatStr that just repeats after we change this one to do this?

ci/test-lib.nix
```diff
@@ -0,0 +1,193 @@
...
+    in
+    if builtins.elem system platforms then test else null;
+
+  it = message: condition: { inherit message condition; };
+in
+{
+  inherit enableForSystem toSanitizedJSON it;
+  enableBySystem = enableForSystem pkgs.stdenv.hostPlatform.system;
+
+  isDirectory = path: it "No such directory ${path}" ''[ -d "${path}" ]'';
```
zenoli: This feels wrong. 

The "it" syntax implies that the name/message argument should describe the intent of the behavior to be tested:

`it "should ensure file exists" *assertions testing the behavior*`

But now we have:

`it "descriptive error message"`

Just by looking at this line in isolation, it should read:

```nix
isDirectory = path: mkAssert "No such directory ${path}" ''[ -d "${path}" ]'';
```

Actually, even this is unconventional. it should even be:

```nix
isDirectory = path: mkAssert ''[ -d "${path}" ]'' "No such directory ${path}" ;
```

(`assert` is already taken)

Maybe this would work also:

```nix
isDirectory = path: test ''[ -d "${path}" ]'' "No such directory ${path}" ;
```

But honestly, it feels we are trying to make a compromise between your and my vision that is worse than going all-in in either of the to visions.

Let me try to outline "my vision". 
It consists of three components:

- `test-suite := [test | test-suite]` (restrict it to lists and not sets, because every test is required to specify a name)
- `test := name: [assertion]`
- `assertion := { cond, errMsg }` (`errMsg` was currently named `msg` but it is actually an `errMsg`)

My solutions (V1, V2) roughly had:

- `runTests` as `test-suite`. They lacked the recursive property though. Only a single top-level `test-suite` was possible.
- `runTest` as the test
- `assertion`: I was generating strings directly, but conceptually they played the same role

Basically, if I were to rename `runTests --> describe` and `runTest --> it` (+ implemented recursion) I would have the busted syntax.

But I do not care so much about having busted syntax, but I think these abstract components which I outlined are pretty much an established convention in many programming languages.

In Java/Python you would use classes for test-suite (in some you can then nest the classes) and the functions within those classes are the tests.

If we have something structurally equivalent to this convention, people will immediately feel familiar writing and reading them.

Your current solution, is more powerful in the sense that you can do even more fancy nesting, but my main concern is, that it is unconventional. 

And I'd rather emphasize that it is "different" by not misleading the reader to believe they are working with something  familiar by using established conventions such as "it".

BirdeeHub: > I do not care so much about having busted syntax

I don't either I just thought it was fun. I do care about having something I am going to type over and over be short.

`do` `check` `run` `test` `it`

My concern with my approach is that the messages double as names and vice versa.

I will need to figure out a way to deal with that properly.

BirdeeHub: I kinda like the nesting, but the nesting is what makes it unconventional.

If every test is required to have a name, isn't making it sets better? Because then you can put the name as the name in the set?

What if we removed the assertion abstraction from the main function of this so it doesnt handle the set form for the assertions anymore, kept the ability to define nested tests, but those tests are just strings or lists of strings.

Then the helpers would work like you had them and build to strings, and it would keep some of the crazy nesting behavior under control, while still allowing you to break them up a bit.

BirdeeHub: also I changed the names to `mkTestDrv` and `test`

zenoli: > If every test is required to have a name, isn't making it sets better? Because then you can put the name as the name in the set?

Yes, this would work too. Actually this make it obvious to me what we want.

It's gonna take a while to explain, so please bare with me. 
Here we go:

There are two ways to model a mapping from `name --> entity`:

1. By using an attrsOf: `attrsOf test`. The names are the keys of the `attrs`
2. By using a list: `[test]` where `test = name: ...`. Here the function is providing the name.

Another point is, how to model assertions. There are also two ways:

1. Specify assertions as `{ cond, errMsg }` which will be rendered to a bash command string by the framework.
2. Specify bash commands as strings directly.

So we have two points, and two ways of modelling each, which makes in total 2x2 combinations which are all valid solution designs.  I will explicilty outline all of these now:


### Version A: Using `attrsOf` and `{cond, errMsg}`

#### Grammar:
- `test-suite := attrsOf (test-suite | test )` 
- `test := attrsOf [assertion]`
- `assertion := { cond, errMsg }`

#### Example:

```nix
{
  "direnv-tests" = {
    "it should have the right version" = [
      { cond = "some bash cond"; errMsg = "Wrong version."; }
    ];
    "it should do something else" = [
      (isDirectory "some directory")  # <-- syntactic sugar for { cond, errMsg}
    ];
    "nix-direnv-tests" = { # <-- new test suite indicated by attrs
      "it should test something for nix-direnv" = [
        { cond = "???"; errMsg = "???"; } # <-- maybe we could also convert single assertiosn toList but I don't think it is necessary:
                                          # The issue is that we already provide test-suites as attrs and we need to check the keys.
                                          # Also, tests named "cond" would not be allowed as this has a different meaning.
      ];
    };
  };
}
```

---

### Version B: Using `attrsOf` and `bash cmd string`

#### Grammar:
- `test-suite := attrsOf (test-suite | test )` 
- `test := attrsOf [assertion]`
- `assertion := str`

#### Example:

```nix
{
  "direnv-tests" = {
    "it should have the right version" = [
      (mkAssert { cond = "some bash cond"; errMsg = "Wrong version."; })  # <-- mkAssert generate a bash string
    ];
    "it should do something else" = [
      (isDirectory "some directory")  # <-- wrapper around mkAssert
    ];
    "nix-direnv-tests" = { # <-- new test suite indicated by attrs
      "it should test something for nix-direnv" = mkAssert { cond = "???"; errMsg = "???"; }; # <-- Now we can easily allow this, as it is unambiguous
                                                                                             # When we provide a string directly, simply call `toList`
                                                                                             # and we have a valid [assertion]
      ];
    };
  };
}
```

---

### Version C: Using `[test]` and `{cond, errMsg}`

#### Grammar:
- `test-suite := name: [test-suite | test]`
- `test := name: [assertion]`
- `assertion := { cond, errMsg}`


#### Example:

```nix
describe "direnv-tests" [
  (it "should have the right version" [
    { cond = "some bash cond"; errMsg = "Wrong version."; }  # <-- mkAssert generate a bash string
  ])
  (it "should do someting else" = [
    (isDirectory "some directory")  # <-- wrapper around mkAssert
  ])
  (describe "nix-direnv-tests" [ # nested test-suite
    (it "should test something for nix-direnv" [{ cond = "???"; errMsg = "???"; }])
  ])
]

```

---


### Version D: Using `[test]` and `bash cmd string`

#### Grammar:
- `test-suite := name: [test-suite | test]`
- `test := name: [assertion]`
- `assertion := str`


#### Example:

```nix
describe "direnv-tests" [
  (it "should have the right version" [
    (mkAssert { cond = "some bash cond"; errMsg = "Wrong version."; })  # <-- mkAssert generate a bash string
  ])
  (it "should do someting else" = [
    (isDirectory "some directory")  # <-- wrapper around mkAssert
  ])
  (describe "nix-direnv-tests" [ # nested test-suite
    (it "should test something for nix-direnv" (mkAssert { cond = "???"; errMsg = "???"; }))
  ])
]

```

---

All four options are more than enough in terms of flexibility: They allowing arbitrary nesting of tests using test-suites, are fully conventional, easy to read and easy to use. I don't care which one, and in case of options C & D what words `test/define/describe/it` whatever we use (allthough now I got a bit used to the busted syntax, lol)

Can we PLEASE, PRETTY PLEASE agree on one of these four and implement it as shown here?

If you want to do it its fine, I am very glad to it myself. Just let's not do something independent in parallel again :-)

zenoli: Actually, (C, D) can be easily converted to (A, B) using:

```nix
it = name: assertions:
  { inherit name; value = assertions; };

describe = name: children:
    { inherit name; value: lib.listToAttrs children }
```

So, all we need is a function that takes an attrs like in A or B and creates the fancy bash command using indents, "(", "&&", ")" as you did and we should have a nice solution.

zenoli: Ok I tested it and it works.
And I think the solution is very clean and concise:

https://github.com/zenoli/nix-wrapper-modules/blob/test-utils/ci/test-lib.nix#L93-L122

(the branch is still a horrible mess because I haven't cleaned it up yet. Just look at the render function)

You can now define arbitrarily nested test suites like this:

```nix
describe "collection a" [
    (it "test a" [
      {
        cond = "[[ 1 -eq 1 ]]";
        message = "msg11";
      }
      {
        cond = "[[ 1 -eq 1 ]]";
        message = "msg11";
      }
    ])
    (it "test b" [
      {
        cond = "[[ 1 -eq 1 ]]";
        message = "msg21";
      }
      {
        cond = "[[ 1 -eq 1 ]]";
        message = "msg21";
      }
    ])
    (describe "collection b" [  # <-- nested
      (it "test c" [
        {
          cond = "[[ 1 -eq 1 ]]";
          message = "msg31";
        }
        {
          cond = "[[ 1 -eq 1 ]]";
          message = "msg31";
        }
      ])
    ])
  ];
```

I took your approach of generating the nicely formatted bash command:

```bash
run() {
  ((
    ((
      ((
        (([[ 1 -eq 1 ]]) || (echo "msg31" >&2 && return 1)) &&
        (([[ 1 -eq 1 ]]) || (echo "msg31" >&2 && return 1))
      ) || (echo 'test c' >&2 && return 1))
    ) || (echo 'collection b' >&2 && return 1)) &&
    ((
      (([[ 1 -eq 1 ]]) || (echo "msg11" >&2 && return 1)) &&
      (([[ 1 -eq 1 ]]) || (echo "msg11" >&2 && return 1))
    ) || (echo 'test a' >&2 && return 1)) &&
    ((
      (([[ 1 -eq 1 ]]) || (echo "msg21" >&2 && return 1)) &&
      (([[ 1 -eq 1 ]]) || (echo "msg21" >&2 && return 1))
    ) || (echo 'test b' >&2 && return 1))
  ) || (echo 'collection a' >&2 && return 1))
}

run || (echo 'Test failed' >&2 && exit 1)

```

As outlined above the `describe/it` functions are just syntactic sugar. Without them you would provide `attrs` like this:

```nix
  testTree = {
    "outer-test-suite" = {
      "test1" = [
        {
          cond = "[[ 1 -eq 1 ]]";
          message = "msg11";
        }
        {
          cond = "[[ 1 -eq 1 ]]";
          message = "msg12";
        }
      ];
      "test2" = [
        {
          cond = "[[ 1 -eq 1 ]]";
          message = "msg21";
        }
      ];
      a.b.c.d.e.f.g = {
        "test3" = [
          {
            cond = "[[ 1 -eq 1 ]]";
            message = "msg31";
          }
        ];
      };
    };
  };
```
which would generate:

```bash
run() {
  ((
    ((
      ((
        ((
          ((
            ((
              ((
                ((
                  ((
                    (([[ 1 -eq 0 ]]) || (echo "msg31" >&2 && return 1))
                  ) || (echo 'test3' >&2 && return 1))
                ) || (echo 'g' >&2 && return 1))
              ) || (echo 'f' >&2 && return 1))
            ) || (echo 'e' >&2 && return 1))
          ) || (echo 'd' >&2 && return 1))
        ) || (echo 'c' >&2 && return 1))
      ) || (echo 'b' >&2 && return 1))
    ) || (echo 'a' >&2 && return 1)) &&
    ((
      (([[ 1 -eq 1 ]]) || (echo "msg11" >&2 && return 1)) &&
      (([[ 1 -eq 1 ]]) || (echo "msg12" >&2 && return 1))
    ) || (echo 'test1' >&2 && return 1)) &&
    ((
      (([[ 1 -eq 1 ]]) || (echo "msg21" >&2 && return 1))
    ) || (echo 'test2' >&2 && return 1))
  ) || (echo 'outer-test-suite' >&2 && return 1))
}

run || (echo 'Test failed' >&2 && exit 1)
```

zenoli: I would clean it up tomorrow, steal your `enableBySystem` function and everything you added which is currently missing in my branch and update the docs.

I really think this now the best version that unifies the ideas from both "visions" without it being a "compromise".

What do you think @BirdeeHub ?

BirdeeHub: > Just let's not do something independent in parallel again :-)

You asked me for an example, and I wasn't quite sure lol

IDK I just figured I would do a proper proposal of an idea XD

I will take a look at it properly soon and pick one.

It looks better though.

Just to be clear, there are only 2 to pick from now, and the one in the PR is option A, vs option B which is taking strings instead of cond and errmsg, correct?

Is there a reason we can't take both string AND `{ cond, errmsg }`? Or just consistency?

zenoli: > You asked me for an example, and I wasn't quite sure lol
>
> IDK I just figured I would do a proper proposal of an idea XD

Yeah, I was really just looking for some nix-pseudocode and did not expect you to commit to a full solution :-)

You're putting out nix code at a pace I simply cannot keep up with, and I just felt that my idea was becoming obsolete before I had a real chance to finalize it. I'm sorry for my frustration leaking through there a bit.

I am now actually thankful that you proposed an alternative because I think it really pushed towards a solution that is
... in equal parts your and my contribution...
... and is clearly superior to what I wrote initially.

> I will take a look at it properly soon and pick one.

Just please wait until I finished everything up, otherwise you will think it sucks for the wrong reasons^^
Will be another 10-12h until I get to it.

zenoli: > Just to be clear, there are only 2 to pick from now, and the one in the PR is option A, vs option B which is taking strings instead of cond and errmsg, correct?

Yes, only two, but it's A vs C (the two with `{cond, errMsg}`).
Option C, when all the `it/describe` functions are evaluated will essentially result in an attrs like in option A. (with a simple deviation at top level that is fixed with a one-liner, which also would give us a nice way to "force" the user to use `it/describe`, which to be honsest is my favourite at the moment (Option C, that is).

The reason why I prefer `{cond, errMsg}` is, that `render` transforms an attrs to a bash command string.
I felt it was more consistent, to also render assertions from `{cond, errMsg}` to a bash command string in `render`. It will make more sense once I cleaned it up.

zenoli: Hmm... I'm having an issue atm:

The generated script does not find the nix store paths during `runCommand` (even though the paths exist). 
This never was an issue so far, I need to figure out what causes it (the generated script works if I copy & paste it and execute it. So I assume have a race condition or sth?

```
trace:
run() {
  ((
    ((
      ((true) || (echo "I should never see this" >&2 && return 1)) &&
      (([[ -d "/nix/store/59768waqif1nasz1vagrn3m2x99sf8x4-direnv-2.37.1/direnv-dot-dir" ]]) || (echo "No such directory /nix/store/59768waqif1nasz1vagrn3m2x99sf8x4-direnv-2.37.1/direnv-dot-dir" >&2 && return 1)) &&
      (([ -f "/nix/store/59768waqif1nasz1vagrn3m2x99sf8x4-direnv-2.37.1/direnv-dot-dir/lib/nix-direnv.sh" ]) || (echo "No such file /nix/store/59768waqif1nasz1vagrn3m2x99sf8x4-direnv-2.37.1/direnv-dot-dir/lib/nix-direnv.sh" >&2 && return 1))
    ) || (echo 'if nix-direnv is enabled then lib/nix-direnv.sh should exists' >&2 && return 1))
  ) || (echo 'direnv-test' >&2 && return 1))
}

run || (echo 'Test failed' >&2 && exit 1);

touch $out

error: Cannot build '/nix/store/qaaay2xv50afw2siwk3pi3qxh5fp48kv-direnv-test.drv'.
       Reason: builder failed with exit code 1.
       Output paths:
         /nix/store/0jpy5zdxxbmc0a27nnqz8rgsgvqp8z10-direnv-test
       Last 4 log lines:
       > No such directory /nix/store/59768waqif1nasz1vagrn3m2x99sf8x4-direnv-2.37.1/direnv-dot-dir
       > if nix-direnv is enabled then lib/nix-direnv.sh should exists
       > direnv-test
       > Test failed
       For full logs, run:
         nix log /nix/store/qaaay2xv50afw2siwk3pi3qxh5fp48kv-direnv-test.drv
```

BirdeeHub: I think I prefer A?

Maybe? I think nesting with sets is much more natural in nix, and since we have separate names and error messages, this makes more sense.

Honestly, less helper functions to have to remember is generally better.

If we go with A though we should probably keep something like `it` still. Can A still be created with describe/it too if someone wants to?

Also

You have a render function, it is separate. It turns the tree format into the busted format.

I would the render function be built in, or would they still need to call render separately?

I still feel like it should be able to take either a string, or `{ cond, errMsg }`, and have it just report the name if you choose the string version?

https://github.com/BirdeeHub/nix-wrapper-modules/pull/452#discussion_r3132259640

^ I have no idea.

I don't think there should be anything removing the context explicitly?

Im not sure how one ends up with a race condition in nix. There is no concurrency. But it may be an ordering thing.

zenoli: > I think I prefer A?
> 
> Maybe? I think nesting with sets is much more natural in nix, and since we have separate names and error messages, this makes more sense.
> 
> Honestly, less helper functions to have to remember is generally better.
> 
> If we go with A though we should probably keep something like `it` still. Can A still be created with describe/it too if someone wants to?

Yes, when when the describe/it functions are evaluated, we have a set that basically is A (with a tiny transform missing at top-level which is fixed in a one-liner). So C really is just a fancy way to write sets for A.
But to be honest, I'd rather not implement describe/it at all if we go for A. It just makes it confusing for new users.

> You have a render function, it is separate. It turns the tree format into the busted format.

No, the render function only accepts the set representation as in A. It turns the set into a bash command string.

> I would the render function be built in, or would they still need to call render separately?

It will be built-in. The user just defines the tests and `render` gets called under-the-hood. Maybe this is not the case in the snippet I linked previously, but will be once I'm done with it.

> I still feel like it should be able to take either a string, or { cond, errMsg }, and have it just report the name if you choose the string version?

Yes, we can also do this. I also had a similar idea which I wanted to propose. But I first need to fix this bug with it not finding the store paths^^

BirdeeHub: Yeah it's pretty mysterious tbh.

It's probably some simple to fix (but impossible to spot) quirk manifesting in a weird way tho. Like toString without calling builtins.path first or something silly like that.

Normally, we use wlib.types.stringable for options that take paths, because it handles that for you and accepts more values. It normally means we don't have to think about that.

Maybe that is what is happening here somehow.

zenoli: > Yeah it's pretty mysterious tbh.
> 
> It's probably some simple to fix (but impossible to spot) quirk manifesting in a weird way tho. Like toString without calling builtins.path first or something silly like that.

I'm debugging this right now. I am rubber ducking a bit because I feel that if I describe what I found out, you immediately know what is going on:

AFAIK `runCommand` is a wrapper around `mkDerivation`. During building, I think I remember from Nix Pills, that the script is executed in some kind of "sandbox" with $HOME set to "homeless-shelter" or sth to make builds not rely on arbitrary system paths. I think there is some kind of magic going on, where all the store paths interpolated into the script using `${...}` are somehow "whitelisted" because they are build inputs and we need to rely on them by reading them.

Because I have a render function that renders an attr to a bash command string, I think I am obfuscating this. `mkDerivation` can no longer autmatically derive these build inputs, because it receives a plain string where interpolation already was evaluated.

For example, my tests want to use this derivation:
```nix
      wrapper = self.wrappers.direnv.wrap {
        inherit pkgs;
        nix-direnv.enable = true;
      };
 ```
 
but it currently does not work. If I "hack" this drv into the `runTest` fn (which is the test-lib function that renders the tests and executes `runCommand`) by explicilty inerpolating it here again, it works:


```nix
  runTest =
    test: # <-- the test attrs in "almost A"-form
    let
      testScript = render (lib.listToAttrs (lib.toList test));    # <-- The "one-line" transform to make it the "fully A" form

      wrapper = self.wrappers.direnv.wrap {  # <-- this should not be here. this is the "hack"
        inherit pkgs;
        nix-direnv.enable = true;
      };
    in
    lib.trace testScript (
      runCommand test.name { passthru.test = test; } ''
        ${render (lib.listToAttrs (lib.toList test))};
        echo ${wrapper} # <-- here I interpolate it.
        touch $out
      ''
    );
```

If in my direnv/check.nix I make assertion like `(isDirectory wrapper)` they only work because of this hack.

This explanation somewhat makes sense, I just don't get why it then also worked in my first version, or in your V3. I think we always had some form of "render" step. :shrug:

BirdeeHub: I think its that you are passing them to trim? Like 3 times or so times? Maybe? And then it is losing the string context?

If it isn't that, I have no clue

zenoli: Omg... yes that was it. You're a genius!

Thank you so much! I need to leave now, but will finish the cleanup later

zenoli: The first time I've even heard of string contexts, was in your "install" PR.
I wanted to know what it is about, stumbled on [this](https://nix.dev/manual/nix/2.26/language/string-context#string-context) and decided "yeah, I will probably never care about this" and closed it. Maybe I should look at it again^^

BirdeeHub: I do some fancy stuff with discarding contexts via structured derivation for the docs

It is good to know.

You likely won't use it directly. It will help you fix bugs like this one though. And know when you actually should discard

BirdeeHub:
@zenoli I suppose I should close this now that the example has served its purpose?

But, if you are curious about uses of `wlib.makeCustomizeable` I guess come back and check this out haha

Ill leave the branch on the repo at least until we finish with the thing, but I think it should still be viewable via this PR even after the branch is removed.

PR #453, opened 2026-04-22
feat(wlib.genStr): now takes function as well

BirdeeHub:
also added wlib.repeatStr which is the old genStr. The new genStr accepts both forms of argument

https://github.com/BirdeeHub/nix-wrapper-modules/pull/452#discussion_r3120872218

This is backward compatible, you do not need to change usage.

The motivation is that `builtins.genList` and `lib.genAttrs` both can receive functions as their first argument. This technically made `wlib.genStr` misnamed because it could not. Now it can!

### DIFF
--- a/lib/lib.nix
+++ b/lib/lib.nix
@@ -607,11 +607,23 @@ in
       (builtins.attrNames attrs);
 
   /**
-    genStr :: string -> int -> string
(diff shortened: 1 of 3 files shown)


BirdeeHub:
wow why is it still running

PR #460, opened 2026-04-24
fix(wrapperModules.niri): change extraConfig to lib.types.lines

iynaix:
Use `lib.types.lines` for `extraConfig` so additional text can be appended when using .apply or .wrap.


BirdeeHub:
TY must have missed that one. Usually we use lines for this sort of thing, so not sure what happened there!

PR #465, opened 2026-04-27
feat(wrapperModules.mangowc): use new official mango configuration syntax

pengolord:
Mango recently added a nix -> mango converter to their home-manager module, figured it'd be nice to copy here

This _should_ make it so that anyone that uses the home-manager module can just copy-paste their config and it will still work
It also shouldn't impact anyone using the nix wrapper currently (aside from the renaming of `extraContent` to `extraConfig`)

### DIFF
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -245,7 +245,7 @@
       relPath = "autostart_sh";
       content = config.autostart_sh;
       builder = ''
-        mkdir -p "$(dirname "$2")" && printf '%s\n' '#!/bin/bash' > "$2" && cat "$1" >> "$2" && chmod +x "$2"
+        mkdir -p "$(dirname "$2")" && printf '%s\n' ${lib.escapeShellArg "#!${pkgs.bash}${pkgs.bash.shellPath}"} > "$2" && cat "$1" >> "$2" && chmod +x "$2"
       '';
     };
 
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -234,7 +234,15 @@
           + "\n"
           + extraConfig
           + "\n"
-          + lib.optionalString (config.autostart_sh != "") "\nexec-once=~/.config/mango/autostart.sh\n";
+          + lib.optionalString (
+            config.autostart_sh != ""
+          ) "\nexec-once=${config.constructFiles.autostart_sh.path}\n";
+    };
+
+    constructFiles.autostart_sh = {
+      relPath = "autostart_sh";
+      content = config.autostart_sh;
+      builder = ''mkdir -p "$(dirname "$2")" && cp "$1" "$2" && chmod +x "$2"'';
     };
 
     flags."-c" = config.configFile.path;
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -192,6 +192,7 @@
           Deprecated, please use 'extraConfig' instead
         '';
         default = "";
+        internal = true;
       };
     };
 
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -244,7 +244,9 @@
     constructFiles.autostart_sh = {
       relPath = "autostart_sh";
       content = config.autostart_sh;
-      builder = ''mkdir -p "$(dirname "$2")" && cp "$1" "$2" && chmod +x "$2"'';
+      builder = ''
+        mkdir -p "$(dirname "$2")" && printf '%s\n' '#!/bin/bash' > "$2" && cat "$1" >> "$2" && chmod +x "$2"
+      '';
     };
 
     flags."-c" = config.configFile.path;
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -10,7 +10,12 @@
 
   options =
     let
-      inherit (lib) mkOption types;
+      inherit (lib)
+        literalExpression
+        mkOption
+        mkOptionDefault
+        types
+        ;
     in
     {
       settings = mkOption {
@@ -38,16 +43,14 @@
           should be written as lists. Variables and colors names should be
           quoted. See <https://mangowc.vercel.app/docs> for more examples.
 
-          ::: {.note}
-          This option uses a structured format that is converted to Mango's
+          Note: This option uses a structured format that is converted to Mango's
           configuration syntax. Nested attributes are flattened with underscore separators.
           For example: `animation.duration_open = 400` becomes `animation_duration_open = 400`
 
           Keymodes (submaps) are supported via the special `keymode` attribute. Each keymode
           is a nested attribute set under `keymode` that contains its own bindings.
-          :::
         '';
-        example = lib.literalExpression ''
+        example = literalExpression ''
           {
             # Window effects
             blur = 1;
@@ -102,7 +105,7 @@
         type = types.lines;
         default = "";
         description = ''
-          Extra configuration lines to add to `~/.config/mango/config.conf`.
+          Extra configuration lines to add to the end of the generated config file.
           This is useful for advanced configurations that don't fit the structured
           settings format, or for options that aren't yet supported by the module.
         '';
@@ -113,7 +116,7 @@
       };
 
       topPrefixes = mkOption {
-        type = with lib.types; listOf str;
+        type = with types; listOf str;
         default = [ ];
         description = ''
           List of prefixes for attributes that should appear at the top of the config file.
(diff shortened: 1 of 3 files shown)


wrapperModules/m/mangowc/module.nix
```diff
@@ -8,61 +8,194 @@
...
+          }
+        '';
+      };
+
+      extraContent = mkOption {
+        type = types.lines;
+        description = ''
+          Deprecated, please use 'extraConfig' instead
+        '';
+        default = "";
```
BirdeeHub, line 193:
```suggestion
        default = "";
        internal = true;
```

^ remove it from the docs (thanks for adding a deprecation warning)

BirdeeHub:
Can you send me a link to the home manager module?

I am only seeing one for [pkgs.mangohud](https://github.com/nix-community/home-manager/blob/master/modules/programs/mangohud.nix)? Is this the same program?

It would be nice to review them side by side but I am unable to find it, which is weird.

But the options do look good, will look more carefully soon.

pengolord:
I completely copied the lib.nix from mango's flake
Here's the link
https://github.com/mangowm/mango/blob/main/nix/lib.nix
and the home-manager module
https://github.com/mangowm/mango/blob/main/nix/hm-modules.nix

pengolord:
oops, just realized that autostart_sh isn't applied at all

pengolord:
is there a way to make an executable construct file?

BirdeeHub:
> is there a way to make an executable construct file?

yes. You will have to set `builder = ''mkdir -p "$(dirname "$2")" && cp "$1" "$2" && chmod +x "$2"'';` for that constructed file.

Maybe we can add more options for this some day, but those options would just set the default for this value anyway, or something to that effect.

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/465#issuecomment-4324487583

^ TY I didnt find it because I was looking in home manager for it.

BirdeeHub:
> oops, just realized that autostart_sh isn't applied at all

and yes, your instinct is correct, you would want to create this via the wrapper derivation, and then use its placeholder path directly.

wrapperModules/m/mangowc/module.nix
```diff
@@ -8,61 +8,192 @@
...
+        example = ''
+          [
+            ./config.conf
+            ./binds.conf
+            ./theme.conf
+          ]
+        '';
+      };
+
+      configFile = mkOption {
```
BirdeeHub, line 169: Do we actually want 2 options here for overwriting the whole module?

Having a config.configFile.path is probably good, but they can still source impure files via sourcefiles option, so, I am not sure how necessary this option is?

If we do want both the path option, and something able to set the entire final content, maybe we could set

```nix
        type = wlib.types.file {
          content = the_generated_config; # <- note the lack of lib.mkOptionDefault means that when they set it, it will append by default.
          # We could then add autostart.sh via lib.mkAfter via setting `config.configFile.content = lib.mkAfter the_autostart;` to make sure it is after the stuff they set still, and just use that instead of extraConfig? Or maybe in addition to extraConfig?
        };
```

And then we would just use that content value as is in `config.constructFiles.generatedConfig.content`

BirdeeHub: ok, so, looking at what we had before, I see why we have it the way it is currently.

But, I guess, did you want to update how that works since the generated thing does more now?

If we are going to leave it like this, its fine, but we should make the warning on this option description more obvious, and include mentioning `config.settings` in it. The description also still says `extraContent`

So, yeah either we should revise this option slightly, or at least its description

But otherwise, looks good!

pengolord: I like the idea of removing the option - it seems obsolete - but I'm a little worried about breaking people's config
And changing it to append `content` to the end instead of using `extraConfig` would also be nice, but then the problem is that it wouldn't be a drop-in replacement for mango's home-manager module, or we would have two options that both add content to the end

wrapperModules/m/mangowc/module.nix
```diff
@@ -8,61 +8,192 @@
...
+        type = wlib.types.file {
+          path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
+        };
+        default = { };
+        description = ''
+          Config file that mango will set as its config file.
+
+          Note: If configFile.path or configFile.content is set, it will overwrite the effects of the `sourcedFiles` and `extraContent` options.
```
BirdeeHub:
```suggestion
          Note: If `configFile.path` is set, it will be used INSTEAD of the generated configuration. The generated file will still be created, however, and you can source it.
          
          If `configFile.content` is set, it will replace the contents of the generated config file entirely.
```

We need to update this message, and if we are leaving it like this, we should make this warning more obvious than it was.

wrapperModules/m/mangowc/module.nix
BirdeeHub: I am totally fine with 2 options that append.

`configFile.content` can also be used to prepend if we did that, via `lib.mkBefore`, so it still might be useful to have. And if they want to replace it, they can `lib.mkForce`. Personally, I would prefer we set content in this option tbh.

As far as breaking people's stuff, removing it entirely would. I would prefer to remove it, except I value not breaking stuff more.

But just setting it, because it is `lib.types.lines`, their existing stuff would just append, and it was previously mutually exclusive, so it would be appending to probably nothing. It is probably fine to just do that.

So, I think we should go with setting content in the `configFile.content` option, and then using that value in `constructFiles.generatedConfig.content`

pengolord:
okay, one last thing, the original uses `pkgs.writeShellScript`, which comes with some nice things like the shebang being included. Does construct files have an equivalent or should I use `pkgs.writeShellScript` anyways, since people could add their own construct files if needed?

wrapperModules/m/mangowc/module.nix
```diff
@@ -8,61 +8,194 @@
...
+      configFile = mkOption {
+        type = wlib.types.file {
+          path = mkOptionDefault config.constructFiles.generatedConfig.path;
+        };
+        default = { };
+        description = ''
+          The config file that mango will set as its primary config file.
+          By default, this file will be generated from whatever the other options are set to.
+
+          Note: If `configFile.path` is set, it will be used INSTEAD of the generated configuration. The generated file will still be created, however, and you can source it. Add your path to the `sourcedFiles` option if you want the generated config to still apply.
```
BirdeeHub:
```suggestion
          Note: If `configFile.path` is set, it will be used INSTEAD of the generated configuration. The generated file will still be created, however, and you can source it.
```

It won't. `sourcedFiles` goes into that file. They would have to source it by interpolating it via a placeholder path in nix, or something like `$(which mangowc)/../path/to/it` or some other way to get it into that other file at runtime

BirdeeHub:
> okay, one last thing, the original uses `pkgs.writeShellScript`, which comes with some nice things like the shebang being included. Does construct files have an equivalent or should I use `pkgs.writeShellScript` anyways, since people could add their own construct files if needed?

echo it in in the builder.

Don't use writeShellScript because it will make an intermediate derivation, causing placeholder paths to not work in that file, so it won't be able to refer to other stuff in the wrapper derivation.

wrapperModules/m/mangowc/module.nix
pengolord: Is this more clear?

Note: If `configFile.path` is set, it will be used INSTEAD of the generated configuration. The generated file will still be created, however, and you can source it. Add the path you would be setting `configFile.path` to to the `sourcedFiles` option instead if you want the generated config to still apply.

or still incorrect?

BirdeeHub: to the sourcedFiles option

to the sourcedFiles option where?
Because if you mean config.sourcedFiles, then that will not work, because the values in config.sourcedFiles go into the generated file.

Sorry, I should have explained what I meant a bit more

You would need to `source=${config.constructFiles.generatedConfig.path}` (or something equivalent) in the file you replace it with.

pengolord: Sorry I'm really bad at explaining this.
Note: If `configFile.path` is set, it will be used INSTEAD of the generated configuration. The generated file will still be created, however, and you can source it. If you don't want the generated configuration to get overwritten, add the path you want to use to the `sourcedFiles` option and do not set `configFile.path`.

BirdeeHub: Yes that is much better. I was indeed completely misunderstanding it.

wrapperModules/m/mangowc/module.nix
```diff
@@ -77,18 +210,47 @@
...
+            config.autostart_sh != ""
+          ) "\nexec-once=${config.constructFiles.autostart_sh.path}\n";
     };
 
-    flags."-c" = config.configFile.path;
+    constructFiles.autostart_sh = {
+      relPath = "autostart_sh";
+      content = config.autostart_sh;
+      builder = ''
+        mkdir -p "$(dirname "$2")" && printf '%s\n' '#!/bin/bash' > "$2" && cat "$1" >> "$2" && chmod +x "$2"
```
BirdeeHub:
```suggestion
        mkdir -p "$(dirname "$2")" && printf '%s\n' ${lib.escapeShellArg "#!${pkgs.bash}${pkgs.bash.shellPath}"} > "$2" && cat "$1" >> "$2" && chmod +x "$2"
```

PR #470, opened 2026-04-28
feat(scheduled deprecations): wrapperModules output now has modules in it

BirdeeHub:
April 30th deprecations

outputs.wrapperModules now has the importable form of the wrapper module, just like the flake-parts module produces.

Some other deprecations that were added at either the same time or significantly before have also been finally removed


BirdeeHub:
Im not entirely understanding the point of waiting until tomorrow to merge this, so I am going to merge it today, and then merge the install modules because I am tired of waiting for those as well.

So, april 30th deprecations are now april 29th deprecations I guess.

PR #473, opened 2026-04-28
feat(wrapperModules.niri): option for auto translated repeated values

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/discussions/472#discussioncomment-16740966

@patwid thoughts?

### DIFF
--- a/wrapperModules/n/niri/module.nix
+++ b/wrapperModules/n/niri/module.nix
@@ -151,38 +151,46 @@ in
             type = lib.types.attrs;
             description = "Bindings of niri";
             apply = convertAndWarn;
-            example = {
-              "Mod+T".spawn-sh = "alacritty";
-              "Mod+J".focus-column-or-monitor-left = _: { };
-              "Mod+N".spawn = [
-                "alacritty"
-                "msg"
-                "create-windown"
-              ];
-              "Mod+0".focus-workspace = 0;
-              "Mod+Escape" = _: {
-                props.allow-inhibiting = false;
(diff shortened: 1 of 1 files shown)


patwid:
I think the include option is similar to spawn-at-startup et al. Maybe we could model include as a dedicated option as well?

Usage could look something like this maybe:

```
include = [
  "required-path" 
  { _: { props = { optional = true; }; content = "optional-path"; } 
];
```

BirdeeHub:
We can certainly still add `includes` like we did `binds` and stuff like that. We probably should.

However, we would have to keep doing that every time there was an option like that that gets added.

I would like to have both if possible? But if I had to pick 1, generic lets people easily map to that themselves? Whereas in extraConfig, they might end up calling toKdl themselves anyway, or ignoring it and doing some crazy string templating that could have been much simpler like this.

BirdeeHub:
Because with extraSettings it is

```nix
config.extraSettings = [
  { include = ./idk; }
  { include = [ { optional = true; } "~/impure/path" ]; }
];
```

Which, is honestly fairly close to the original with some extra punctuation.

BirdeeHub:
Should it be `settings.extra`?

BirdeeHub:
Im going to just go with it. I was thinking about putting it inside settings as settings.extra, but what it actually is, is a list of multiple OF the settings set.

So, it belongs at the same level, and it is literally "extra `config.settings`"

I also improved the doc examples. It was sanitizing functions.

PR #477, opened 2026-04-28
fix(tlib): handle singleton assertions in set form

zenoli:
We did not handle of passing singleton assertions like this:

```nix
test "my-test" {
  "test description" = { cond = "some condition"; msg = "some message"; };
}
```

because they were indistinguishable from nested tests.

This is not ideal, as technically, we cannot create a test-group containing exactly two tests named "cond" and "msg" but I think it is fine.

### DIFF
--- a/ci/test-lib.nix
+++ b/ci/test-lib.nix
@@ -80,11 +80,23 @@ let
       renderAssertions =
         assertions: lib.concatMapStringsSep " &&\n" (a: "(${renderAssertion a})") assertions;
 
+      isAssertion =
+        x:
+        lib.isString x
+        || (
+          lib.isAttrs x
+          &&
+            lib.attrNames x == [
+              "cond"
+              "msg"
+            ]
+        );
+
       renderNode =
         node:
         let
           block =
-            if !(lib.isAttrs node) then
+            if lib.isList node || isAssertion node then
               renderAssertions (lib.toList node)
             else
               lib.concatMapAttrsStringSep " &&\n" (name: childNode: ''
@@ -137,12 +149,11 @@ in
     where:
 
     ```
-    TestSet :: attrsOf (TestSet | Test)
+    TestSet :: Test | attrsOf (TestSet | Test)
 
-    Test :: [ Assertion ]
+    Test :: Assertion | [ Assertion ]
 
-    Assertion :: String | { cond :: String; msg :: String; }
-    ```
+    Assertion :: String | { cond :: String; msg :: String; }    ```
 
     # Arguments
 


BirdeeHub:
Yeah thats basically what I ended up doing in my v3 version, but I had a mutually recursive list-set instead of just lists in sets, which was definitely more confusing.

Would you rather just leave it as list and fix the docs? Or do you want to ensure that works? Because string works like this, which is a nice shortcut, and needing to put assertions in a list is not too bad. But, they are often function calls so `[ (isDirectory ./idk) ]`... I guess that is annoying enough to warrant it.

zenoli:
> Yeah thats basically what I ended up doing in my v3 version,

Yes, it felt familiar^^

> Would you rather just leave it as list and fix the docs? Or do you want to ensure that works? Because string works like this, which is a nice shortcut, and needing to put assertions in a list is not too bad. But, they are often functions so [ (isDirectory ./idk) ]... I guess that is annoying enough to warrant it.

Either singleton works for both or for none. I think it is nice syntactic sugar to allow it for both. I cannot imagine the potential issue I outlined become reality. And if so, I am willing to remove this sugar and adjust all the tests.

I'm confident that it won't be an issue

BirdeeHub:
Also

https://birdeehub.github.io/nix-wrapper-modules/tlib.html#function-library-tlib.test

^ this type might be wrong?

BirdeeHub:
Ok. Well, review that type in the doc comment again because I think it is incorrect now? Or maybe before and it is right now? Unsure. But once we verify that is correct we can update this. I will check again in the next day or 2 for sure

zenoli:
You mean this part, right?

```
Test :: [ Assertion ]
```
It should be 

```
Test :: Assertion | [ Assertion ]
```

Also I noticed here that we can directly pass `[Assertion]` to the main `test` function:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/478/changes

So it should even be 

```
TestSet :: Test | attrsOf (TestSet | Test)

Test :: Assertion | [ Assertion ]

Assertion :: String | { cond :: String; msg :: String; }
```

Because in that PR I did not see a need to introduce another name for a group containing a single test, so I just passed `Assertion` directly which workes too but I did never consider before.

zenoli:
I updated the docs accordingly

BirdeeHub:
Yeah I think that is at the very least more correct.

BirdeeHub:
the tests are so unbelievably slow right now

zenoli:
> the tests are so unbelievably slow right now

It's the github issues right? Not a potential performance regression introduced by tlib?

BirdeeHub:
Oh, yeah, whoops. I meant github

github has been like this all day

tlib is being used for 1 test, and it has 1 list of assertions

ANd the ubuntu check is 12 minutes behind schedule and counting

If you managed to screw it up that bad you probably just proved P!=NP

BirdeeHub:
Ok, awake again, the tests passed obviously.

Ran them again and they took their usual minute or 2 to run

PR #478, opened 2026-04-28
test(makeCustomizable): refactor tests with tlib

zenoli:
I just wanted to check how well it works to use tlib for non-wrapper tests.

If I have time I can also do the other ones. If you don't want to risk touching all of the checks, or don't want to review all of it we can also leave them as they are and just use tlib going forward.

### DIFF
--- a/ci/checks/makeCustomizable.nix
+++ b/ci/checks/makeCustomizable.nix
@@ -1,9 +1,14 @@
 {
   pkgs,
   self,
+  tlib,
   ...
 }:
 let
+  inherit (tlib)
+    areEqual
+    test
+    ;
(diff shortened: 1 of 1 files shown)


BirdeeHub:
I was going to start with that one myself. I just rewrote it anyway XD

zenoli:
> I was going to start with that one myself. I just rewrote it anyway XD

Oh no. Feel free to take over, I don't mind :-)

BirdeeHub:
I wanted to get rid of luaEnv from it because they were changing where the value I was looking for was... But I had just done that, so there is nothing sacred about that test whatsoever (or any of them, for that matter, but that one in particular)

PR #485, opened 2026-04-30
feat(devShell): init

zenoli:
Mostly written by claude. I found it annoying to always remember these, so this made things a bit more convenient for me. Maybe it is for others too.

If you think this is bloat I have no hard feelings. You can better judge whether those things will change often, in which case I'd rather not want to maintain those :-)

It adds a shell.nix providing two helper scripts:

- `docs` for running the docs generator and serving it at localhost:1337
- `checks` for either running all checks, or specific wrapperModule/wlib/module checks if specified.

It uses direnv to automatically load/unload these scripts.
Direnv is of course not required and it can still be used with `nix develop` or `nix-shell` (we can also remove .envrc entirely)

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,89 @@
+{
+  pkgs ? import <nixpkgs> { },
+}:
+
+let
+  system = pkgs.stdenv.hostPlatform.system;
+  docs = pkgs.writeShellApplication {
+    name = "docs";
+    runtimeInputs = [ pkgs.python3 ];
+    text = ''
+      usage() {
+        echo "Usage: docs [OPTION]"
+        echo ""
+        echo "Builds and serves the documentation."
+        echo ""
+        echo "Options:"
+        echo "  -p, --port PORT  Port to serve on (default: 1337)"
+        echo "  -h, --help       Show this help"
+      }
+
+      port=1337
+
+      while [ $# -gt 0 ]; do
+        case "$1" in
+          -h|--help)
+            usage
+            exit 0
+            ;;
+          -p|--port)
+            port="$2"
+            shift 2
+            ;;
+          *)
+            echo "Unknown option: $1"
+            usage
+            exit 1
+            ;;
+        esac
+      done
+
+      nix run ./ci#docs -- ./_site && cd _site && python3 -m http.server "$port"
+    '';
+  };
+
+  test = pkgs.writeShellScriptBin "check" ''
+    usage() {
+      echo "Usage: check [OPTION] [NAME]"
+      echo ""
+      echo "With no arguments, runs all checks."
+      echo ""
+      echo "Options:"
+      echo "  -w, --wrapperModule NAME  Run check wrapperModule-NAME"
+      echo "  -l, --wlib NAME           Run check wlib-NAME"
+      echo "  -m, --module NAME         Run check module-NAME"
+      echo "  NAME                      Run check NAME"
+      echo "  -h, --help                Show this help"
+    }
+
+    if [ $# -eq 0 ]; then
+      nix flake check -Lv ./ci
+      exit 0
+    fi
+
+    case "$1" in
+      -h|--help)
+        usage
+        exit 0
+        ;;
+      -w|--wrapperModule)
+        nix build "./ci#checks.${system}.wrapperModule-$2"
+        ;;
+      -l|--wlib)
+        nix build "./ci#checks.${system}.wlib-$2"
+        ;;
+      -m|--module)
+        nix build "./ci#checks.${system}.module-$2"
+        ;;
+      *)
+        nix build "./ci#checks.${system}.$1"
+        ;;
+    esac && echo "Test passed!"
+  '';
+in
+pkgs.mkShell {
+  packages = [
+    docs
+    test
+  ];
+}
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 _site
 **/result-*
 **/repl-result-*
+.direnv
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,9 +125,6 @@ Example:
 
 `nix fmt`
 
-## Tests
-
-`nix flake check -Lv ./ci`
 
 ## Run Site Generator Locally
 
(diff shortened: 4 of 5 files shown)


shell.nix
```diff
@@ -0,0 +1,53 @@
+{ pkgs ? import <nixpkgs> { } }:
+
+let
+  system = pkgs.stdenv.hostPlatform.system;
+  docs = pkgs.writeShellApplication {
+    name = "docs";
+    runtimeInputs = [ pkgs.python3 ];
+    text = ''
+      nix run ./ci#docs -- ./_site && cd _site && python3 -m http.server 1337
```
BirdeeHub:
```suggestion
      nix run ./ci#docs -- ./_site && cd _site && python3 -m http.server ''${1:-1337}
```

should let them choose the port if they want to

shell.nix
```diff
@@ -0,0 +1,53 @@
...
+    '';
+  };
+
+  test = pkgs.writeShellScriptBin "check" ''
+    usage() {
+      echo "Usage: check [OPTION] [NAME]"
+      echo ""
+      echo "With no arguments, runs the docs."
```
BirdeeHub:
```suggestion
      echo "With no arguments, builds the docs to `./_site`."
```

It builds mdbook with a wrapped config, and then runs it, to build the docs. But we don't need to explain all that, but this is slightly more descriptive

shell.nix
```diff
@@ -0,0 +1,83 @@
...
+
+      nix run ./ci#docs -- ./_site && cd _site && python3 -m http.server "$port"
+    '';
+  };
+
+  test = pkgs.writeShellScriptBin "check" ''
+    usage() {
+      echo "Usage: check [OPTION] [NAME]"
+      echo ""
+      echo "With no arguments, runs the docs."
```
BirdeeHub: Although, second guessing...

Lets have no arguments run `nix flake check -Lv ./ci` instead (i.e. run ALL the checks)

Docs should be for docs, this should be for checks

zenoli: Ahem.. yeah that comment was completely off. The check script does exactly that. Docs are not generated in this script at all. I'm fixing it

zenoli: Yes, it is wrong the way it is rn.

zenoli: It should be fixed now

shell.nix
zenoli: As said below, this description should not be here at all. 
But on L14 it is described as  "Builds and serves the documentation."

shell.nix
```diff
@@ -0,0 +1,88 @@
...
+      nix flake check -Lv ./ci
+      exit 0
+    fi
+
+    case "$1" in
+      -h|--help)
+        usage
+        ;;
+      -w|--wrapperModule)
+        nix build "./ci#checks.${system}.wrapperModule-$2"
```
BirdeeHub, line 70:
```suggestion
        nix build -v "./ci#checks.${system}.wrapperModule-$2"
```

^ These give close to no output when successful. They do build something to ./result, but if they had built something already with a nix command, it might not be immediately obvious that it did anything or not.

Should we add -v to these?

BirdeeHub: nvm -v doesnt give much more. -vv would but then it is too much and mostly irrelevant. It is fine like this.

zenoli: I just tried this, but I don't see a difference with -v?

zenoli: I can of course echo "SUCCESS" or sth if the command is successful?

BirdeeHub: Hmmm

I guess we could do that?

Might be nice. Lets do that. Should be easy to add and will improve the experience.

The `-Lv` command does not need it. But the others, it would be nice.

BirdeeHub: but yeah otherwise, seems nice. Makes it so we lean less on documentation to describe this, which is nice.

zenoli: > Hmmm
> 
> I guess we could do that?
> 
> Might be nice. Lets do that. Should be easy to add and will improve the experience.
> 
> The `-Lv` command does not need it. But the others, it would be nice.

I did exactly this. If an individual test is successful it will print `Test passed!`

zenoli:
I have some improvements for CONTRIBUTING.md to do still. Pushing soon

BirdeeHub:
I like it

But building it had me wondering something, because my network cut out for a moment...

Should we be setting `preferLocalBuild` by default on wrapper modules?

Regardless, letting the mac test run and then I will merge this.

zenoli:
Improved the docs a bit:

image

zenoli:
> Should we be setting preferLocalBuild by default on wrapper modules?

I have no experience with it and its implication. You would set this for the wrapper tests only or in general?

shell.nix
```diff
@@ -0,0 +1,89 @@
...
+        echo "Usage: docs [OPTION]"
+        echo ""
+        echo "Builds and serves the documentation."
+        echo ""
+        echo "Options:"
+        echo "  -p, --port PORT  Port to serve on (default: 1377)"
+        echo "  -h, --help       Show this help"
+      }
+
+      port=1377
```
BirdeeHub:
```suggestion
      port=1337
```

shell.nix
```diff
@@ -0,0 +1,89 @@
...
+    name = "docs";
+    runtimeInputs = [ pkgs.python3 ];
+    text = ''
+      usage() {
+        echo "Usage: docs [OPTION]"
+        echo ""
+        echo "Builds and serves the documentation."
+        echo ""
+        echo "Options:"
+        echo "  -p, --port PORT  Port to serve on (default: 1377)"
```
BirdeeHub:
```suggestion
        echo "  -p, --port PORT  Port to serve on (default: 1337)"
```

BirdeeHub: its so close to 1337 that every single person is gonna misread that 🤣 

Either 1337 or 42069 or go boring with 8080 or something

shell.nix
zenoli: An important nit ^^

shell.nix
zenoli: So true :-)

zenoli:
I squashed and renamed the commit. I hope the format is fine this time?

BirdeeHub:
> > Should we be setting preferLocalBuild by default on wrapper modules?
> 
> I have no experience with it and its implication. You would set this for the wrapper tests only or in general?

It was unrelated to this PR, but, I was talking about putting it in lib/core.nix where it defines the default drv args

Building these kinds of things is usually so fast that it would be slower to ask the internet.

BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/485#issuecomment-4356134899

oo nice I didnt know if mdbook would handle that or not

zenoli:
> [#485 (comment)](https://github.com/BirdeeHub/nix-wrapper-modules/pull/485#issuecomment-4356134899)
> 
> oo nice I didnt know if mdbook would handle that or not

I think I read somewhere in their docs that you used to need a special preprocessor for this but is now natively supported.

shell.nix
zenoli: Forgot to push the change 😅 .

It's fixed now

BirdeeHub:
LGTM!

PR #488, opened 2026-05-01
feat(wrapperModules.starship): init

zenoli:
Wrapper Module for the Starship Prompt

### DIFF
--- a/wrapperModules/s/starship/module.nix
+++ b/wrapperModules/s/starship/module.nix
@@ -1,0 +1,117 @@
+{
+  wlib,
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  tomlFmt = pkgs.formats.toml { };
+
+  presetKey = "preset";
+  settingsKey = "settings";
+
+  defaultOrder = [
+    presetKey
+    settingsKey
+  ];
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    settings = lib.mkOption {
+      inherit (tomlFmt) type;
+      default = { };
+      description = ''
+        Pure nix configuration of starship.toml.
+        See <https://starship.rs/config/>
+      '';
+      example = {
+        directory.format = "[ $path ]($style)";
+      };
+    };
+    preset = lib.mkOption {
+      type = with lib.types; either str (listOf str);
+      default = [ ];
+      apply = lib.toList;
+      description = ''
+        One or more built-in starship presets to use as configuration.
+        When a list is provided, presets later in the list take precedence.
+        See <https://starship.rs/presets/>.
+      '';
+      example = [
+        "nerd-font-symbols"
+        "tokyo-night"
+      ];
+    };
+    order = lib.mkOption {
+      type = with lib.types; wlib.types.fixedList 2 (enum defaultOrder);
+      default = defaultOrder;
+      description = ''
+        The order in which the specified settings are merged.
+        Values later in the list will take precedence.
+
+        The allowed keys are:
+
+        - "${presetKey}": Settings from the the specified preset (`config.preset`)
+        - "${settingsKey}": Settings specified as a nix attrs (`config.settings`)
+      '';
+    };
+  };
+  config = {
+    package = lib.mkDefault pkgs.starship;
+    constructFiles."starship.toml" = {
+      content = builtins.toJSON config.settings;
+      relPath = "starship.toml";
+      builder =
+        let
+          # Holds the bash variable name to be used inside the builder script
+          nixSettingsTomlPathVar = "nix_settings_toml";
+          orderedSettings =
+            let
+              tomlSettingsMap = {
+                ${presetKey} = map (
+                  p: lib.escapeShellArg "${config.package}/share/starship/presets/${p}.toml"
+                ) config.preset;
+                ${settingsKey} = lib.optional (config.settings != { }) "\$${nixSettingsTomlPathVar}";
+              };
+            in
+            lib.concatMap (key: tomlSettingsMap.${key}) config.order;
+        in
+        # Builds the nix settings TOML file and stores the path in the
+        # bash variable named `${nixSettingsTomlPathVar}`
+        lib.optionalString (config.settings != { }) ''
+          ${nixSettingsTomlPathVar}="$(mktemp)"
+          mkdir -p "$(dirname ''$${nixSettingsTomlPathVar})"
+          ${pkgs.remarshal}/bin/json2toml "$1" "''$${nixSettingsTomlPathVar}"
+        ''
+        # Merges all specified presets TOML files and the TOML file generated
+        # from nix settings using tomlq
+        + ''
+          mkdir -p "$(dirname "$2")"
+          ${pkgs.yq}/bin/tomlq -s -t 'reduce .[] as $item ({}; . * $item)' \
+          ${lib.concatStringsSep " " orderedSettings} > "$2"
+        '';
+    };
+    env.STARSHIP_CONFIG = config.constructFiles."starship.toml".path;
+    meta = {
+      maintainers = with wlib.maintainers; [
+        birdee
+        zenoli
+      ];
+      description = ''
+        Wrapper Module for the [Starship Prompt](https://starship.rs/).
+
+        Starship is configured via a [TOML file](https://starship.rs/config/).
+        This module provides two ways to do this:
+
+        - By specifying one (or many) of the built-in preset configurations.
+        - By using pure Nix to write an attribute set that gets converted to TOML.
+
+        These two options are not mutually exclusive. If both are defined,
+        they will be merged according to the order specified in `config.order`.
+      '';
+    };
+  };
+}
--- a/wrapperModules/s/starship/check.nix
+++ b/wrapperModules/s/starship/check.nix
@@ -1,0 +1,95 @@
+{
+  pkgs,
+  self,
+  tlib,
+  ...
+}:
+
+let
+  inherit (tlib)
+    fileContains
+    isFile
+    test
+    ;
+  wm = self.wrappers.starship;
+in
+test { wrapper = "starship"; } {
+
+  "wrapper should output correct version" =
+    let
+      wrapper = wm.wrap {
+        inherit pkgs;
+      };
+    in
+    ''
+      "${wrapper}/bin/starship" --version |
+      grep -q "${wrapper.version}"
+    '';
+
+  "starship.toml is properly configured" =
+    let
+      wrapper = wm.wrap {
+        inherit pkgs;
+      };
+      configFile = "${wrapper}/starship.toml";
+    in
+    [
+      (isFile configFile)
+      (fileContains "${wrapper}/bin/starship" "STARSHIP_CONFIG.*${configFile}")
+    ];
+
+  "default settings order is respected" =
+    let
+      # See here for preset values: https://github.com/starship/starship/blob/main/docs/public/presets/toml/tokyo-night.toml#L22
+      wrapper = wm.wrap {
+        inherit pkgs;
+        preset = "tokyo-night";
+        settings.directory.style = "foo";
+      };
+      configFile = "${wrapper}/starship.toml";
+    in
+    [
+      (isFile configFile)
+      (fileContains configFile ''truncation_symbol = "…/"'')
+      (fileContains configFile ''style = "foo"'')
+    ];
+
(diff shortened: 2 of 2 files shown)


wrapperModules/s/starship/module.nix
```diff
@@ -0,0 +1,119 @@
...
+  wlib,
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  presetRoot = config.package.src + "/docs/public/presets/toml";
+  presets = lib.pipe presetRoot [
+    lib.readDir
```
BirdeeHub: I dont think we should be doing this I think that readDir in this case is an IFD, because we are calling it on the `config.package.src`.

Is there any way we can do this preset merge within the drv builder itself, rather than in nix code?

As cool as it is, the cost may be too high if we can only do it via IFD

wrapperModules/s/starship/module.nix
```diff
@@ -0,0 +1,119 @@
...
+  orderedSettings = lib.map (key: settingsMap.${key}) config.order;
+  mergedSettings = deepMerge orderedSettings;
+  mergedToml = tomlFmt.generate "starship.toml" mergedSettings;
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    "starship.toml" = lib.mkOption {
+      type = with lib.types; nullOr path;
```
BirdeeHub: I am having trouble figuring out what the difference is between this, and the settings option. This cannot be an impure path because we have to importTOML from it, so, I am unsure that this is any different than people calling importTOML themselves?

wrapperModules/s/starship/module.nix
zenoli: You are right, it is an IFD. 
I also have another IFD [here](https://github.com/BirdeeHub/nix-wrapper-modules/pull/488/changes#diff-499ddb10660fc62e2ea1c8e11bdc594cf7ca4fcd73b29b69155830ec0e25ab15R98) which I could fix by doing it like this: https://github.com/BirdeeHub/nix-wrapper-modules/commit/39ea2c7e212a4e8125615605f4d45eeffb3ae3d8#diff-f90d3590aadf353d8c24c9acf19f842b2aa952701fb6a831a015c0440dae9b47R27

But it seems that currently as soon as I am trying to touch a preset I get an IFD (I am currently testing it with: `nix build .#starship --option allow-import-from-derivation false`)

I don't want to give up on the preset feature right now. I noticed the nixos-module is using it as well (even a generalized version that allows for merging multiple presets): https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/starship.nix#L26

I'm still not 100% confident when an IFD occurs. Does the nixos module introduce an IFD or is the approach using `tomlq` somehow avoiding it?

wrapperModules/s/starship/module.nix
zenoli: So you suggest dropping this option and let users do `importTOML ./starship.toml`?

Yeah this would be simpler and better.

zenoli: I removed "starship.toml".

wrapperModules/s/starship/module.nix
zenoli: Ok, I removed the dynamic presets and fixed the preset loading by mimicking the approach from nixpkgs. 
It also supports merging a list of presets as in nixpkgs now.

wrapperModules/s/starship/module.nix
zenoli: I removed "starship.toml". There is only `config.settings` and `config.preset` now

wrapperModules/s/starship/module.nix
zenoli: If we still want an enum listing the presets we could hard code the list and add a test, verifying if the hardcoded list is consistent with the bundled presets using `readDir`. 

Of course this shifts the IFD to the test. I assume you don't want this too?
Also, if a preset changes upstream, the test would spontaneously fail.

If we don't want to allow IFD's at all, should we add them to the `check` script in shell.nix to enforce it?

BirdeeHub: > If we still want an enum listing the presets

NGL I don't really want to list out all the possible presets, that sounds like a nightmare to keep up to date since we can't IFD

Why not just test it with 1 or 2 of them? That way, we test our logic and not "which presets they have" unless I suppose they remove the presets we were using in the tests.

This is looking much better btw, but will have to look at it again soon.

BirdeeHub:
Will test this out in a bit, but it looks much better now that it merges them at build time rather than via IFD

wrapperModules/s/starship/module.nix
zenoli: Yeah prob not worth it to maintain them. I gave an example in de description and there is a link to the presets in the starship website, so people should be able to infer what strings are valid here. Having a dynamically derived enum would have been the best UX but I trust you that the performance penalty is not worth having it.

Are you against IFD in checks too? If so, should we disallow them in the `check` script?

BirdeeHub: It might be worth it to disable it in the check script? and maybe also the github workflow? In a separate PR though?

Basically, with an IFD, what happens is, it needs to evaluate the whole drv being IFD'd from during evaluation before it can continue. Which is single threaded and blocking...

And you are planning on using starship inside at least 2 other wrapper modules yourself.

So, that's why we want to avoid that

But in the checks themselves, they are all disconnected top level exports. I am not sure it really matters as much there

So, I am not sure. Probably? Regardless, I don't want to check their list, it only matters that when the user selects one that exists, it works, and preferably if it does not exist it tells them something to that effect. (this one throws with a message from tomlq if you do that, so it does, so that is good)

wrapperModules/s/starship/module.nix
```diff
@@ -0,0 +1,103 @@
...
+    settingsKey
+  ];
+
+  nixSettingsToml = tomlFmt.generate "nix-settings.toml" config.settings;
+
+  tomlSettingsMap = {
+    ${presetKey} = map (
+      p: lib.escapeShellArg "${config.package}/share/starship/presets/${p}.toml"
+    ) config.preset;
+    ${settingsKey} = [ (lib.escapeShellArg "${nixSettingsToml}") ];
```
BirdeeHub:
```suggestion
    ${settingsKey} = [ "\${nixSettingsTomlPath:-}" ];
```

^ will expand to nothing if unset

wrapperModules/s/starship/module.nix
```diff
@@ -0,0 +1,103 @@
...
+    };
+  };
+  config = {
+    package = lib.mkDefault pkgs.starship;
+    constructFiles."starship.toml" = {
+      content = "";
+      relPath = "starship.toml";
+      # Merge all applicable TOML files at build time using tomlq (avoids IFD).
+      # Store paths for preset and nix-settings are embedded directly in the builder string.
+      builder = ''
```
BirdeeHub:
```suggestion
      builder = lib.optionalString (config.settings != { }) ''
        nixSettingsTomlPath="$(mktemp)"
        mkdir -p "$(dirname $nixSettingsTomlPath)"
        ${pkgs.remarshal}/bin/json2toml "$1" "$nixSettingsTomlPath"
      '' + ''
```

wrapperModules/s/starship/module.nix
```diff
@@ -0,0 +1,103 @@
...
+
+        - "${presetKey}": Settings from the the specified preset (`config.preset`)
+        - "${settingsKey}": Settings specified as a nix attrs (`config.settings`)
+      '';
+    };
+  };
+  config = {
+    package = lib.mkDefault pkgs.starship;
+    constructFiles."starship.toml" = {
+      content = "";
```
BirdeeHub:
```suggestion
      content = builtins.toJSON config.settings;
```

BirdeeHub (commented):
lets use `.content` instead of `tomlFmt.generate`

Otherwise I think it's good (and cool, and I have a custom wrapper for starship in my config that I was able to improve by turning it into just an extension of this one so that is also cool).

wrapperModules/s/starship/module.nix
zenoli: `"$1"` will be the coming from `constructfiles."starship.toml".content, right?

I have to say this is very hard to read.

L76: `content = builtins.toJSON config.settings;` is confusing because it reads like we're passing JSON to a setting that ultimatively will be TOML

L81: We introduce an env variable `nixSettingsTomlPath` which is also referred to in L25 

L83: We convert the JSON nix settings to the TOML only here.

I guess this is more performant? It is not avoiding an IFD I am unaware of, right? Because I thought this was already solved.

If we do this, would you be ok if I added you to the maintainer list? I'm pretty sure I will have a hard time debugging this if this ever comes up in like a year or two and I forgot everything :-)

BirdeeHub: This one wasn't about IFD this one was about placeholders. Using tomlFmt.generate creates an intermediate drv, meaning you can't use placeholders in it. Which would maybe be useful, because you can add custom commands to the starship config, and that capability may benefit from having placeholders work in the file.

I would be OK with you adding me to the list.

Although, passing json to something that is going to be toml is how we usually do toml.

https://github.com/BirdeeHub/nix-wrapper-modules/blob/6316917effdaae39e3b483f798d20f1ddba52e5a/wrapperModules/a/alacritty/module.nix#L25-L27

The confusing part, is that, we make an intermediate file first (`$nixSettingsTomlPath`), and then combine it with the preset ones. Rather than just converting content directly to output.

`$1` is the input path created by the drv containing the content passed to the `constructFiles.<name>.content` option. `$2` is the output path `constructFiles.<name>.builder` is to output to.

BirdeeHub:
awww sorry... formatter...

Should have thought about that with the github suggestion thing.

zenoli:
> awww sorry... formatter...
> 
> Should have thought about that with the github suggestion thing.

No worries. It's fixed now.

I moved the whole TOML merging and ordering logic in a `let ... in` right above the builder script to make it easier to read.
I also refactored the nixSettingsTomlPath bash variable into a nix variable, to have a "single source of truth". But maybe it made things more unreadable now due to the "double interpolation" of nix string into bash variable names into bash evaluation :shrug:

BirdeeHub:
> I also refactored the nixSettingsTomlPath bash variable into a nix variable, to have a "single source of truth". But maybe it made things more unreadable now due to the "double interpolation" of nix string into bash variable names into bash evaluation 🤷

Its fine you also added a useful comment, which counteracts it XD

The escaping is a bit weird though yeah. (especially `'' ... ''$${`), But it is correct, and there are tests, so, its fine.

What I want to find is another way to emit toml using tomlq instead of remarshal.

But at the very least, tomlFmt.generate was going to pull remarshal anyway so we are no worse.

So, I think I am happy with this! Ill see if I can find a way to remove remarshal later.

zenoli:
> What I want to find is another way to emit toml using tomlq instead of remarshal.

I did a quick check in `yq` but could not find an easy way to convert JSON to TOML there. But I have not spent a lot of time tbh :-)

BirdeeHub:
Yeah I didnt either... regardless, it was going to pull remarshal anyway, at least maybe we can eventually do something about that this way.

wrapperModules/s/starship/module.nix
zenoli: Ok. I might add the flag in a separate PR soon. I would still make it opt-in, as there might be a case where a test-only IFD is needed.

zenoli:
Cool, I'm done with any changes from my side.

BirdeeHub:
I had 1 last thing to check sorry. Just wanted to check again with no args and with an invalid preset. It all works as expected! :+1:

BirdeeHub:
https://github.com/BirdeeHub/birdeeSystems/blob/e1358c38aef3a5cadd7f4941677c85907964eb22/common/wrappers/starship/default.nix

^ I have this layer on top to make a sourceable version in my config but it is kinda messy and doesn't work for nushell yet XD But I can stick it right on top with no issues!

zenoli:
Pretty cool!

I am mainly using it for zsh atm. I also added powerlevel10k (which is installed  as a zsh plugin under the hood) next to starship (which is added as an "integration" under the hood) and I also plan to add oh-my-posh and then add a nice way to easily switch between all of them.

All the annoying setup (such as with p10ks "instant prompt") will be managed by nix. 

Here is where I'm at atm (still work in progress)
https://github.com/zenoli/zsh-flake/blob/main/config/default.nix

PR #489, opened 2026-05-01
feat(core.install): add hjem support

netadr:
Created an entry for the `hjem` class under `install.modules` and added an `hjemModules` convenience flake output.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,6 +32,7 @@
         name: value: self.lib.getInstallModule { inherit name value; }
       ) self.lib.wrapperModules;
       homeModules = self.nixosModules;
+      hjemModules = self.nixosModules;
       devShells = forAllSystems (system: {
         default = import ./shell.nix { pkgs = getPkgs system; };
       });
--- a/lib/core.nix
+++ b/lib/core.nix
@@ -184,6 +184,15 @@ in
           in
           lib.mkIf cfg.enable [ cfg.wrapper ];
       };
+    hjem =
+      { config, ... }:
+      {
+        config.packages =
+          let
+            cfg = args.config.install.getWrapperConfig config;
+          in
+          lib.mkIf cfg.enable [ cfg.wrapper ];
+      };
     nixos =
       { config, ... }:
       {


BirdeeHub:
Seems helpful!

One question, I don't use hjem personally so I don't know

Is it a common pattern to import modules directly into `hjem.users.<name>`?

e.g.

```nix
{...}: {
  hjem.users.myuser = {
    imports = [ theinstallmodule ];
  };
}
```

It seems that is where it is intended that users import these install modules, is that correct?

If they are not used to doing this, would it be preferable to use type merging to set the value for all users instead? 

Or would that cause us to run into other problems and/or be generally annoying to not be able to as easily choose who it is imported for? If so, would it be a nice idea to have it be a top level thing, but then also add an option for which users to install it for via mkWrapperExtension?

Personally, I feel this is probably a fine design choice to have them import it directly for the user like that, but I wanted to ensure I definitely understand how to use it, and would like to hear your opinion on that point if I have understood it correctly.


---

Related question, not a blocker for this PR, just a question I wanted to hear your opinion on.

Should we be doing `(lib.hiPrio cfg.wrapper)` on these? To make sure it wins?

netadr:
> One question, I don't use hjem personally so I don't know
> 
> Is it a common pattern to import modules directly into `hjem.users.<name>`?

https://hjem.feel-co.org/options.html#option-hjem.extraModules

> Related question, not a blocker for this PR, just a question I wanted to hear your opinion on.
> 
> Should we be doing `(lib.hiPrio cfg.wrapper)` on these? To make sure it wins?

That I'm honestly not sure about. I would assume that if someone enables a wrapper that they would want it to take precedence in their environment, so probably yes?

BirdeeHub:
https://hjem.feel-co.org/options.html#option-hjem.extraModules

Ah very nice.

So they go in there then. Fair enough. Cool stuff good to know!

---

Thanks for the vote on the other question. I am thinking I should probably do that as well...

BirdeeHub:
Well, in light of extraModules, I think all my questions about this PR have been answered?

LGTM?

Unsure how much we actually need a new modules output for hjem from the main flake? But, it isn't hurting anything anyway its just an alias.

PR #491, opened 2026-05-02
feat(wrapperModules.atuin): init

appleptree:
Wrapper Module for the Atuin CLI.

### DIFF
--- a/wrapperModules/a/atuin/module.nix
+++ b/wrapperModules/a/atuin/module.nix
@@ -31,7 +31,9 @@ in
   config = {
     package = lib.mkDefault pkgs.atuin;
     env.ATUIN_CONFIG_DIR = dirOf config.constructFiles.generatedConfig.path;
-    passthru.ATUIN_CONFIG_DIR = "${config.wrapper.out}/${config.binName}-config";
+    passthru.ATUIN_CONFIG_DIR = "${
+      config.wrapper.${config.constructFiles.generatedConfig.output}
+    }/${config.binName}-config";
     constructFiles = {
       generatedConfig = {
         content = builtins.toJSON config.settings;
--- a/wrapperModules/a/atuin/module.nix
+++ b/wrapperModules/a/atuin/module.nix
@@ -31,6 +31,7 @@ in
   config = {
     package = lib.mkDefault pkgs.atuin;
     env.ATUIN_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config";
+    passthru.ATUIN_CONFIG_DIR = "${config.wrapper.out}/${config.binName}-config";
     constructFiles = {
       generatedConfig = {
         content = builtins.toJSON config.settings;
--- a/wrapperModules/a/atuin/module.nix
+++ b/wrapperModules/a/atuin/module.nix
@@ -1,0 +1,37 @@
+{
+  config,
+  wlib,
+  lib,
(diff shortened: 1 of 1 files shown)


wrapperModules/a/atuin/module.nix
```diff
@@ -0,0 +1,34 @@
...
+      default = { };
+      description = ''
+        Atuin configuration options.
+      '';
+    };
+  };
+
+  config = {
+    package = lib.mkDefault pkgs.atuin;
+    env.ATUIN_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config";
```
BirdeeHub:
```suggestion
    env.ATUIN_CONFIG_DIR = dirOf config.constructFiles.generatedConfig.path;
```

It seems like there is only 1 config file? But the docs say 2? What is that about? Did they misspeak and mean a config file and a data file?

appleptree: The other config file is server.toml and it contains configuration for the atuin-server binary for selfhosting.

BirdeeHub: Ah. So currently, the `atuin-server` binary is unwrapped then? So it will look for that config in the normal location supposedly rather than our wrapped one?

Should we `config.wrapperVariants.atuin-server = {};` to mirror the setting of ATUIN_CONFIG_DIR for that too, and then make an option for that config file as well and put it in our dir?

I haven't looked at the aerc one yet btw but I will get to it.

appleptree: Yes, let's do it that way.

BirdeeHub:
`passthru.ATUIN_CONFIG_DIR = "${config.wrapper.out}/${config.binName}-config";`

^ lets set that too so you can `wrappedatuin.ATUIN_CONFIG_DIR` and get the path to the constructed dir from outside of the module.

Otherwise, seems good I think? I am not super decided on how much we want to use mkOverride and stuff to force stuff to stay in the same directory or not, so, this will be fine after that I think.

Some of the wrapper modules do stuff like

https://github.com/BirdeeHub/nix-wrapper-modules/blob/039e496dc8ad8883e2af5dce1fa2cf408cb0f05e/wrapperModules/e/eww/module.nix#L66-L67

to force them to stay in the same directory, but I suppose it is up to you how much you want it to do that or not.

wrapperModules/a/atuin/module.nix
```diff
@@ -0,0 +1,51 @@
...
+      description = ''
+        Atuin server configuration options.
+      '';
+    };
+  };
+
+  config = {
+    package = lib.mkDefault pkgs.atuin;
+    env.ATUIN_CONFIG_DIR = dirOf config.constructFiles.generatedConfig.path;
+    passthru.ATUIN_CONFIG_DIR = "${config.wrapper.out}/${config.binName}-config";
```
BirdeeHub:
```suggestion
    passthru.ATUIN_CONFIG_DIR = "${config.wrapper.${config.constructFiles.generatedConfig.output}}/${dirOf config.constructFiles.generatedConfig.relPath}";
```

^ since we are locking generatedServerConfig to this output and relative path, lets set this to the same place too.

BirdeeHub: otherwise, I think it is ready to merge! Thanks for handling this edge case!

PR #492, opened 2026-05-02
feat(wrapperModules.aerc): init

appleptree:
Wrapper module for the aerc CLI mail client.

### DIFF
--- a/wrapperModules/a/aerc/module.nix
+++ b/wrapperModules/a/aerc/module.nix
@@ -1,0 +1,95 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+let
+  iniFmt = pkgs.formats.ini { };
+  iniWithGlobalFmt = pkgs.formats.iniWithGlobalSection { };
+in
+{
+  imports = [ wlib.modules.default ];
+  options = {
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Sorry I have been busy. I will get to these, but there have been some bugs that have been higher priority, and I have had things to do IRL

appleptree:
No problem

BirdeeHub:
Actually, this looks good to me.

I have 1 question

https://www.aerc-docs.com/providers/gmail/#mbsync-configuration

What is mbsync? I don't see an argument to wrap that anywhere, but it also seems like something that we would not be breaking either, so it should work as normal... whatever it is?

Trying to see if there is an arg or a var for that too, but also, does that live update? So we would not want that to be in the store? Or is it static like their config files?

It also looks like they let us run aerc commands directly from the command line, so, there would probably be a way to handle that, but I am not sure if we want to be handling that at all, in which case this would be fine as-is?

Edit: ok, so it appears, we can add an option to trigger it automatically from our config, but that is its own thing to wrap or not, so, we don't need to do that here. So this is probably fine as-is

I was also going to say, maybe our themes option should also allow paths, but you already let people add more themes dirs themselves, so that is probably OK not to do so.

PR #494, opened 2026-05-02
feat(wrapperModules.bat): init

appleptree:
Wrapper Module for the bat CLI.

### DIFF
--- a/wrapperModules/b/bat/module.nix
+++ b/wrapperModules/b/bat/module.nix
@@ -20,7 +20,7 @@
       type = lib.types.attrsOf (wlib.types.file pkgs);
       default = { };
       description = ''
-        Bat themes to link to theme directory
+        Bat themes to copy to `themes/` directory
       '';
     };
   };
--- a/wrapperModules/b/bat/module.nix
+++ b/wrapperModules/b/bat/module.nix
@@ -38,7 +38,7 @@
     in
     {
       package = lib.mkDefault pkgs.bat;
-      env.BAT_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config/themes";
+      env.BAT_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config";
       constructFiles = {
         generatedConfig = {
           content = config.configFile.content;
--- a/wrapperModules/b/bat/module.nix
+++ b/wrapperModules/b/bat/module.nix
@@ -10,7 +10,9 @@
 
   options = {
     configFile = lib.mkOption {
-      type = wlib.types.file pkgs;
+      type = wlib.types.file {
+        path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
+      };
       default.content = "";
       description = ''
         Bat flags to include via config file
--- a/wrapperModules/b/bat/module.nix
+++ b/wrapperModules/b/bat/module.nix
@@ -39,7 +39,7 @@
       env.BAT_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config/themes";
       constructFiles = {
         generatedConfig = {
-          content = builtins.readFile config.configFile.path;
+          content = config.configFile.content;
           relPath = "${config.binName}-config/config";
         };
       }
--- a/wrapperModules/b/bat/module.nix
+++ b/wrapperModules/b/bat/module.nix
@@ -1,0 +1,49 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    configFile = lib.mkOption {
+      type = wlib.types.file pkgs;
+      default.content = "";
+      description = ''
+        Bat flags to include via config file
+      '';
+    };
+    themes = lib.mkOption {
+      type = lib.types.attrsOf (wlib.types.file pkgs);
+      default = { };
+      description = ''
+        Bat themes to link to theme directory
+      '';
+    };
+  };
+
+  config =
+    let
+      themes-constructFiles = lib.concatMapAttrs (name: value: {
+        "themes-${name}" = {
+          content = builtins.readFile value.path;
+          relPath = "themes/${name}";
+        };
+      }) config.themes;
+    in
+    {
+      package = lib.mkDefault pkgs.bat;
+      env.BAT_CONFIG_DIR = "${placeholder "out"}/themes";
+      constructFiles = {
+        generatedConfig = {
+          content = builtins.readFile config.configFile.path;
+          relPath = "${config.binName}-config/config";
+        };
+      }
+      // themes-constructFiles;
+      meta.maintainers = [ wlib.maintainers.appleptree ];
+    };
+}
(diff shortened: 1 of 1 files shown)


wrapperModules/b/bat/module.nix
```diff
@@ -0,0 +1,49 @@
...
+          content = builtins.readFile value.path;
+          relPath = "themes/${name}";
+        };
+      }) config.themes;
+    in
+    {
+      package = lib.mkDefault pkgs.bat;
+      env.BAT_CONFIG_DIR = "${placeholder "out"}/themes";
+      constructFiles = {
+        generatedConfig = {
```
BirdeeHub, line 50: I don't see where you are providing this path to bat?

wrapperModules/b/bat/module.nix
```diff
@@ -0,0 +1,49 @@
...
+          content = builtins.readFile config.configFile.path;
```
BirdeeHub:
```suggestion
          content = config.configFile.content;
```

wrapperModules/b/bat/module.nix
```diff
@@ -0,0 +1,49 @@
...
+  lib,
+  pkgs,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    configFile = lib.mkOption {
+      type = wlib.types.file pkgs;
```
BirdeeHub:
```suggestion
      type = wlib.types.file {
        path = lib.mkOptionDefault config.constructFiles.generatedConfig.path;
      };
```

wrapperModules/b/bat/module.nix
```diff
@@ -36,10 +38,10 @@
     in
     {
       package = lib.mkDefault pkgs.bat;
-      env.BAT_CONFIG_DIR = "${placeholder "out"}/themes";
+      env.BAT_CONFIG_DIR = "${placeholder "out"}/${config.binName}-config/themes";
```
BirdeeHub: Should this be `${config.binName}-config/themes` or should it be just `${config.binName}-config`? Edit: yes it appears so?

BirdeeHub:
https://github.com/sharkdp/bat#adding-new-syntaxes--language-definitions

^ this is in our dir too? What to do about this?

wrapperModules/b/bat/module.nix
appleptree: Yes of course, sorry I overlooked it.

wrapperModules/b/bat/module.nix
```diff
@@ -0,0 +1,65 @@
...
+    let
+      themes-constructFiles = lib.concatMapAttrs (name: value: {
+        "themes-${name}" = {
+          content = builtins.readFile value.path;
+          relPath = "themes/${name}";
+        };
+      }) config.themes;
+      syntaxes-constructFiles = lib.concatMapAttrs (name: value: {
+        "syntaxes-${name}" = {
+          content = builtins.readFile value.path;
```
BirdeeHub: Can these be directories too or just files?

---

If they are only files, this is technically ok but I am still not super sure how I feel about readFile here to be honest. It means you can't use an impure path, as you would be able to if you just did ln -s for example, and runs the risk of IFD.

Could you use `constructFiles.<name>.builder` instead to build the value of content if nonempty or copy it if it is empty? Or something to that effect?

---

If they can be directories too, I am not super sure exactly how you want to do that, maybe just accept just `wlib.types.stringable` for them and let them make or get the file themselves if they want it and pass the path there, and then you just `ln -s` them via `config.buildCommand`?

appleptree: They can be directories, too. So let's just accept paths.

wrapperModules/b/bat/module.nix
```diff
@@ -44,7 +44,7 @@
       }) config.themes;
       syntaxes-constructFiles = lib.concatMapAttrs (name: value: {
         "syntaxes-${name}" = {
-          content = builtins.readFile value.path;
+          content = builtins.readFile value;
```
BirdeeHub: Can we either use `.builder` here to do `ln -s`, or `config.buildCommand` to do `ln -s`?

Then they can be impure paths also

Thats why the readFile was bothering me too.

wrapperModules/b/bat/module.nix
```diff
@@ -26,7 +26,7 @@
       '';
     };
     syntaxes = lib.mkOption {
-      type = lib.types.attrsOf (wlib.types.file pkgs);
+      type = lib.types.attrsOf lib.types.string;
```
BirdeeHub: `wlib.types.stringable` is the type for anything that can be `"${interpolated}"`

wrapperModules/b/bat/module.nix
appleptree: If `.builder` does `ln -s`, what should we set `.content` to? And how does `.builder` know about the file it should symlink then?

BirdeeHub: Since we are only accepting paths, because they can be directories, lets not use constructFiles for that

```nix
      buildCommand.makeBatSyntaxes =
        "mkdir -p ${placeholder "out"}/syntaxes\n"
        + lib.concatMapAttrsStringSep "\n" (
          n: v: "ln -s ${v} ${placeholder "out"}/syntaxes/${n}"
        ) config.syntaxes;
```

Something like that is probably better?

Regardless, if we are going to be using readFile, there isn't a point to accepting a path.

appleptree: Yes, you are right, that's a good solution. Should we do the same for `themes`?

AlexLov:
Hey, could you consider adding `settings` option to be able to configure `bat` in Nix instead of pure text? So it'll be easier to override/extend...

Something like this:
```nix
options = {
  # ...
  # Taken from https://github.com/nix-community/home-manager/blob/master/modules/programs/bat.nix
  settings = lib.mkOption {
    type =
      with lib.types;
      attrsOf (oneOf [
        str
        (listOf str)
        bool
      ]);
    default = { };
    example = {
      theme = "TwoDark";
      pager = "less -FR";
      map-syntax = [
        "*.jenkinsfile:Groovy"
        "*.props:Java Properties"
      ];
    };
    description = ''
      Bat configuration.
    '';
  };
};
config = let
  # ...
  toConfigFile =
    attrs:
    let
      inherit (builtins) isBool attrNames;
      nonBoolFlags = lib.filterAttrs (_: v: !(isBool v)) attrs;
      enabledBoolFlags = lib.filterAttrs (_: v: isBool v && v) attrs;

      keyValuePairs = lib.generators.toKeyValue {
        mkKeyValue = k: v: "--${k}=${lib.escapeShellArg v}";
        listsAsDuplicateKeys = true;
      } nonBoolFlags;
      switches = lib.concatMapStrings (k: ''
        --${k}
      '') (attrNames enabledBoolFlags);
    in
    keyValuePairs + switches;
in {
  # ...
  constructFiles = {
    generatedConfig = {
      relPath = "${config.binName}-config/config";
      content =
        if config.configFile.content != ""
        then config.configFile.content
        else lib.optionalString (config.settings != { }) (toConfigFile config.settings);
    };
  }
  // themes-constructFiles;
  # ...
};
```

PR #496, opened 2026-05-03
feat(wrapperModules.niri): hot reload with new config after build

iynaix:
Hot reloads niri with the new config after building, based on the conversation from: https://github.com/Lassulus/wrappers/pull/135

It currently hot reloads by default, since I think it's probably the behavior most people want / expect.

### DIFF
--- a/wrapperModules/n/niri/module.nix
+++ b/wrapperModules/n/niri/module.nix
@@ -363,6 +363,15 @@ in
         as nix no longer needs to know about this path at build time.
       '';
     };
+    disableConfigHotReload = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        When `true`, the wrapper will not hot reload niri with the new config on rebuild.
+
+        Note: Niri 26.04 is required for this functionality.
+      '';
+    };
   };
   config.filesToPatch = [
     "share/applications/*.desktop"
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Yes looks good.

Thanks for this! I need to learn more about these reload if changed triggers

When we get services options it might be really helpful with dealing with copying impure files for things that need writable config directories

wrapperModules/n/niri/module.nix
```diff
@@ -410,6 +419,21 @@ in
...
+    lib.mkIf (!config.disableConfigHotReload && lib.versionAtLeast config.package.version "26.04")
+      {
+        after = [ "symlinkScript" ];
+        data = ''
+          chmod +w ${placeholder config.outputName}/share/systemd/user/niri.service
+          cat >> ${placeholder config.outputName}/share/systemd/user/niri.service<<EOF
+          [Unit]
+          X-Reload-Triggers=${config.constructFiles.generatedConfig.path}
+          [Service]
+          ExecReload=${lib.getExe config.package} msg action load-config-file --path ${config.constructFiles.generatedConfig.path}
```
BirdeeHub, line 432: Actually hang on.

Question.

How does this work.

Wouldn't the service that is registered be the old one, which would have the old path?

It seems like they were doing it the same way there too but I would still like to understand what I am missing here.

Is it that since you have it specifically installed via nixos or home manager, the service gets updated, but it doesn't know that it needs to reload without this trigger, but when it reloads it reloads the new service instead of the current one?

iynaix: What I think happens (I'm no expert) is that the service file changes, systemd sees that the value of X-Reload-Triggers has changed, and runs ExecReload.

I did try doing it directly in buildPhase but it doesn't work in the nix build sandbox, complains about NIRI_SOCKET not being available.

BirdeeHub: > systemd sees that the value of X-Reload-Triggers has changed, and runs E

ah ok that makes some sense actually

Given that several people have mentioned this, lgtm! Thanks for the explanation!

PR #498, opened 2026-05-03
perf?(core.wrapper): preferLocalBuild by default?

BirdeeHub:
https://nix.dev/manual/nix/2.33/protocols/json/derivation/options.html#preferLocalBuild

This prevents it from checking a remote cache for your wrapper derivation with your settings in it.

Many wrapper derivation generators for various programs in nixpkgs and external flakes set this in their derivation, as the derivation contains user provided settings which will not be in a remote cache.

For the vast majority of wrapper derivations, the cost of checking the network should far exceed the amount of time it takes to link/copy/convert the files in the wrapper derivation.

If you have already built your wrapper derivation locally, it will still check your local cache.

### DIFF
--- a/lib/core.nix
+++ b/lib/core.nix
@@ -916,6 +916,7 @@ in
             dontConfigure = true;
             dontPatch = true;
             dontFixup = true;
+            preferLocalBuild = true;
             name = package.name or "${package.pname or binName}-${version}";
             ${
               if builtins.isString (package.pname or binName) && package.pname or binName != "" then


zenoli:
Under what circumstances would not preferring local builds be faster?

BirdeeHub:
It depends entirely on how heavy the build command is and how many substitutors you have

Linking a directory, and converting a couple json files to toml is almost certainly faster than hitting the network is.

Basically, if it is not marked as preferLocalBuild, it tries all your substitutors to try to find your wrapper derivation. It probably won't find it, because you have your own settings in there. Then it can start building.

As opposed to just checking your local drive and if it isnt there, starting the build.

This is especially noticeable when you have a slow network.

At least, in theory. I haven't adequately tested to see if this is actually significantly faster yet. It likely still checks the local disk first regardless, so it should only affect the first build of the module with a particular set of settings either way on a particular machine.

Building the actual program is almost certainly slower than checking the network is, barring the occasional simple script. But we are just making some links, and editing a few files

The thing I need to check is, would this make it harder to implement caching for tests later. If so, maybe not a good idea? Unsure though.

PR #500, opened 2026-05-03
feat(wrapperModules.oh-my-posh): init

zenoli:
So this is basically a copy of the [starship wrapper ](https://github.com/BirdeeHub/nix-wrapper-modules/pull/488) except:

- "presets" (starship) are named "themes" in omp
- Starship uses TOML, omp uses json by default, which made the build script a bit simpler.

One thing that is missing though:

Omp also allow passing yaml files to --config. I've seen no `importYaml` equivalent to `importJSON/importTOML`.
So in order to support YAML we might also allow another option `configFile` which can be either JSON/TOML/YAML (kinda like I had in my first proposal for the starship module, where @BirdeeHub argued - rightly so - that it was not needed since users can simply use `importTOML` if they want to write a native toml file.

I was thinking reading from this new option, and normalize the yaml/toml/json to json using `yq` in the builder and also generalize the `order` field to define the order between `[theme nix-settings configFile]`.

Thoughts, @BirdeeHub ?

### DIFF
--- a/ci/test-lib.nix
+++ b/ci/test-lib.nix
@@ -10,7 +10,7 @@
 let
   wlib = self.lib;
 
-  errMsg = msg: "(echo ${renderMsg msg} >&2 && return 1)";
+  errMsg = msg: "(echo \"${renderMsg msg}\" >&2 && return 1)";
 
   indentBlock =
     str: num:
@@ -281,9 +281,26 @@ in
   };
 
   /**
+    Returns an `Assertion` that checks whether `path` does **not** exist as a directory.
+
+    # Type
+    ```
+    notIsFile :: String -> Assertion
+    ```
+
+    # Arguments
+    path
+    : The filesystem path that should be absent.
+  */
+  notIsDirectory = path: {
+    cond = ''[ ! -d "${path}" ]'';
+    msg = "Directory ${path} should not exist";
+  };
+
+  /**
     Returns an `Assertion` that checks whether `file` contains a line matching `pattern`.
 
-    The check is performed with `grep -q`, so `pattern` is treated as a basic regular expression.
+    The check is performed with `grep -Eq`, so `pattern` is treated as an extended regular expression.
 
     # Type
     ```
@@ -295,10 +312,10 @@ in
     : Path to the file to search.
 
     pattern
-    : Basic regular expression to search for.
+    : Extended regular expression to search for.
   */
   fileContains = file: pattern: {
-    cond = ''grep -q '${pattern}' "${file}"'';
+    cond = ''grep -Eq -- '${pattern}' "${file}"'';
     msg = "Pattern '${pattern}' not found in ${file}";
   };
 
--- a/wrapperModules/o/oh-my-posh/module.nix
+++ b/wrapperModules/o/oh-my-posh/module.nix
@@ -1,0 +1,296 @@
+{
+  wlib,
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  jsonFmt = pkgs.formats.json { };
+
+  themeKey = "theme";
+  fileKey = "file";
+  settingsKey = "settings";
+
+  defaultOrder = [
+    themeKey
+    fileKey
+    settingsKey
+  ];
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
+    settings = lib.mkOption {
+      inherit (jsonFmt) type;
+      default = { };
+      description = ''
+        Pure nix configuration oh-my-posh.
+        See <https://ohmyposh.dev/docs/configuration/general>
+      '';
+      example = {
+        console_title_template = "{{ .Folder }}";
+      };
+    };
+    configFile = lib.mkOption {
+      type = with lib.types; nullOr (either path package);
+      default = null;
+      description = ''
+        Path to an oh-my-posh configuration file.
+        Supported formats are JSON (`.json`), TOML (`.toml`), and YAML (`.yaml`, `.yml`).
+        See <https://ohmyposh.dev/docs/configuration/general>
+      '';
+      example = lib.literalExpression "./config.yaml";
+    };
+    theme = lib.mkOption {
+      type = with lib.types; either str (listOf str);
+      default = [ ];
+      apply = lib.toList;
+      description = ''
+        One or more built-in oh-my-posh themes to use as configuration.
+        When a list is provided, they will be merged by chaining them with 
+        [`.extends`](https://ohmyposh.dev/docs/configuration/general#extends).
+        Themes later in the list take precedence.
+
+        See <https://ohmyposh.dev/docs/themes/>.
+      '';
+      example = [
+        "1_shell"
+        "agnoster"
+      ];
+    };
+    order = lib.mkOption {
+      type = with lib.types; wlib.types.fixedList 3 (enum defaultOrder);
+      default = defaultOrder;
+      description = ''
+        The order in which the specified settings are merged.
+        Values later in the list will take precedence.
+
+        The allowed keys are:
+
+        - "${themeKey}": Settings from the the specified theme (`config.theme`)
+        - "${fileKey}": Settings from the specified config file (`config.configFile`)
+        - "${settingsKey}": Settings specified as a nix attrs (`config.settings`)
+      '';
+    };
+  };
+
+  config =
+    let
+      nixSettingsFile = pkgs.writeText "settings.json" (builtins.toJSON config.settings);
+
+      stripStoreHash =
+        name:
+        let
+          m = builtins.match "[a-z0-9]{32}-(.*)" name;
+        in
+        if m != null then builtins.head m else name;
+
+      normalizedConfigFile =
+        if config.configFile == null then
+          null
+        else
+          let
+            path = toString config.configFile;
+            baseName = stripStoreHash (baseNameOf path);
+            isJson = lib.hasSuffix ".json" baseName;
+            isToml = lib.hasSuffix ".toml" baseName;
+            isYaml = lib.hasSuffix ".yaml" baseName || lib.hasSuffix ".yml" baseName;
+            configFileName =
+              lib.pipe baseName [
+                (lib.removeSuffix ".toml")
+                (lib.removeSuffix ".yaml")
+                (lib.removeSuffix ".yml")
+              ]
+              + ".json";
+          in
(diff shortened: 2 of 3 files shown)


wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,107 @@
...
+    };
+    flags."--config" = config.constructFiles."config.json".path;
+    meta = {
+      maintainers = with wlib.maintainers; [
+        zenoli
+      ];
+      description = ''
+        Wrapper Module for the [Oh-My-Posh Prompt](https://ohmyposh.dev/).
+
+        Oh-My-Posh is configured via a [JSON/YAML/TOML file](https://ohmyposh.dev/docs/configuration/general).
```
zenoli: YAML is not supported as of now.

BirdeeHub:
given the lack of `importYAML`, having a `lib.types.either lib.types.package lib.types.path` is fine. It doesn't support an impure path though, because we have to merge it at build time, hence not using `wlib.types.stringable`

What would the criteria be to determine which type the provided file is of those 3? File extension? Option? Either?

Somewhat busy tonight, will be reviewing stuff again tomorrow night probably

zenoli:
> What would the criteria be to determine which type the provided file is of those 3? File extension? Option? Either?

I was thinking file extension.

`config.configFile = ./my-config.yaml`;

I try to implement an extra option on top, but only if it is opt-in.

I don't want them to write

```nix
config.configFile = {
    path = ./my-config.yaml;
    fileType = "yaml";
}
```

They should use the single-line syntax from above where `fileType` will be `null` and determined by extension.
And they can option in to the longer second syntax if they choose to. I probably can do that by using a `spec` and `path` as the main field?

BirdeeHub:
> I try to implement an extra option on top

Dont worry about it, file extension is fine.

zenoli:
> Dont worry about it, file extension is fine.

Done!

ci/test-lib.nix
```diff
@@ -298,7 +298,7 @@ in
     : Basic regular expression to search for.
   */
   fileContains = file: pattern: {
-    cond = ''grep -q '${pattern}' "${file}"'';
+    cond = ''grep -q -- '${pattern}' "${file}"'';
```
zenoli: I need this to grep for "--config".

wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,140 @@
...
+              isToml = lib.hasSuffix ".toml" path;
+              isYaml = lib.hasSuffix ".yaml" path || lib.hasSuffix ".yml" path;
+              isJson = lib.hasSuffix ".json" path;
+            in
+            if isJson then
+              "_omp_config_json=${config.configFile}"
+            else if isToml || isYaml then
+              ''
+                _omp_config_json=$(mktemp --suffix=.json)
+                ${pkgs.yq-go}/bin/yq -o=json '.' ${lib.escapeShellArg path} > "$_omp_config_json"
```
zenoli: I wasn't aware of yq-go before, but it seems to be a [totally legit package](https://github.com/mikefarah/yq) (15k stars compared to 3k stars in yq). It transforms both TOML and YAML to json in using one command, whereas I would have to use tomlq when using yq.

BirdeeHub: https://github.com/BirdeeHub/tomlua

^ Totally not a legit package (JK I mean it is a perfectly cromulent TOML parser but it is super basic), but Im going to try to merge this because this is going to make stuff like the starship module super clean

But yeah, nice discovery, it looks good at a glance, will have to check out the module soon!

BirdeeHub: I got a bit distracted though and didn't go over this when I had time last night...

zenoli: > https://github.com/BirdeeHub/tomlua
> 
> ^ Totally not a legit package (JK I mean it is a perfectly cromulent TOML parser but it is super basic), 


Oh cool :-)


> but Im going to try to merge this because this is going to make stuff like the starship module super clean

Sorry, don't understand: You mean you will also use yq-go in the starship module?

BirdeeHub: Well maybe

But I was referencing using tomlua for starship.

it got added to nixpkgs yesterday

would make it a good bit cleaner actually.

Will need to wait a while though, because it got added yesterday and stable won't have it

But we could generate to lua instead of json

and then `tomlua --output "$1" --cmd 'return load('return ' .. ...)()' -- "$1"` to turn it into toml

and then

`tomlua --output "$2" -- "$1" "preset1" "preset2"` to merge them

Then we use only 1 dependency for the whole thing, and said dependency is smaller than either of the dependencies it had before.

---

It doesn't do yaml tho. So, like, yq-go is probably what you want for this one.

zenoli: Ok cool.

Btw, don't merge this pr yet. I played around with oh my posh this week a bit (haven't used it before). Turns out the merging of nix/file/theme configs is quite useless atm because omp defines lists of segments within lists of blocks and merging of lists simply replace old lists. Omp provides an "extend" setting to override configs based on another config. Made a rework that uses this and will push it soon. Still need to add some tests though.

I am also wondering whether the starship module does deep or shallow merges of configs. I think we def want deep merges. Something we could add together with the tomlua switch?

BirdeeHub: tomlua does deep merge and it extends lists by default

It has a nifty feature where you can provide it a lua table, and it will read from the file directly into THAT table.

You can customize what the command does to make it do whatever, but by default it uses the mentioned feature. [it does reduce feeding each file as defaults for the next](https://github.com/BirdeeHub/tomlua/blob/0966ec8120b5a9659a3da8ed6246b990291b9750/bin/tomlua#L281-L302) (also, note, yes, it says `if dont_read then read_file` that is correct, if it didnt read, then the default behavior needs to read, yes, confusing, it just confused me which is why I am saying this) 

So

`tomlua --output "$2" -- "$1" "/path/to/preset1" "/path/to/preset2"`

^ would deep merge them, extending the lists

Currently the starship module does shallow merge though...

So, yes, that will be added alongside the tomlua switch, maybe we should do it before? But at the very least we will do it then.

zenoli: Cool, more reasons two use tomlua then :-)

zenoli:
Ok, config merging is fully reworked.

In a nutshell, if you specify multiple settings they will be chained together using the [.extends](https://ohmyposh.dev/docs/configuration/general#extends) key in the order specified in `config.order`.

In case you specified an .extends key explictly, we don't modify it (the chain is "broken").

See the [description](https://github.com/BirdeeHub/nix-wrapper-modules/pull/500/changes#diff-6384f0452ed03aa1af818b0d0d64b88b26a999f344bfcb3b5ee80e02e5f86a99R240) and tests (there are plenty :-)) for more details.

The main use case it, that people might want to use a theme but override some stuff (which was not really possible in my initial version). Since most of the omp settings are in nested lists (segments within blocks) it is a bit clumsy to override them.
I first tried to recreate the mergin logic described in the link above but that was too much work. So making use of the native "extends" feature felt more natural.

zenoli:
Did some cleanup on the tests. I think this is now ready for review.

wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,296 @@
...
+        in
+        if m != null then builtins.head m else name;
+
+      normalizedConfigFile =
+        if config.configFile == null then
+          null
+        else
+          let
+            path = toString config.configFile;
+            baseName = stripStoreHash (builtins.baseNameOf path);
```
BirdeeHub:
```suggestion
            baseName = stripStoreHash (baseNameOf path);
```

wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,296 @@
...
+      orderedConfigs = lib.concatMap (
+        key:
+        {
+          ${themeKey} = map (p: {
+            srcPath = "${config.package}/share/oh-my-posh/themes/${p}.omp.json";
+            name = "${p}.omp.json";
+          }) config.theme;
+          ${fileKey} = lib.optional (config.configFile != null) {
+            srcPath = "${normalizedConfigFile}";
+            name = stripStoreHash (builtins.baseNameOf (toString normalizedConfigFile));
```
BirdeeHub:
```suggestion
            name = stripStoreHash (baseNameOf (toString normalizedConfigFile));
```

wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,296 @@
...
+              inherit relPath;
+              builder = generateBuilderScript curr.srcPath prev.srcPath;
+            };
+          }
+        else
+          # Recursive: prev will itself be a generated file in config-chain
+          {
+            ${relPath} = {
+              inherit relPath;
+              builder = generateBuilderScript curr.srcPath "${builtins.placeholder "out"}/config-chain/${prev.name}";
```
BirdeeHub:
```suggestion
              builder = generateBuilderScript curr.srcPath "${placeholder "out"}/config-chain/${prev.name}";
```

wrapperModules/o/oh-my-posh/module.nix
```diff
@@ -0,0 +1,296 @@
...
+            reversed = lib.reverseList orderedConfigs;
+            lastConfig = lib.last orderedConfigs;
+          in
+          generateConstructFileEntries (builtins.head reversed) (builtins.tail reversed)
+          // {
+            "config.json" = {
+              relPath = "config.json";
+              builder =
+                let
+                  chainDir = "${builtins.placeholder "out"}/config-chain";
```
BirdeeHub:
```suggestion
                  chainDir = "${placeholder "out"}/config-chain";
```

BirdeeHub (commented):
Just some nits left.

This looks good!

Nice find using their extends feature!

wrapperModules/o/oh-my-posh/module.nix
zenoli: Oh, is there a comprehensive list somewhere on what is in global scope and what not?

zenoli:
Removed the 4 occurences of `builtin`

PR #504, opened 2026-05-04
feat(ci): only render wrapper specific options

zenoli:
**Disclaimer**: This was completely one-shotted by Claude. 
The only thing I did manually was adding [collapsible sections](https://rust-lang.github.io/mdBook/format/configuration/renderers.html?highlight=fold#outputhtmlfold).

I did not understand the whole nix-docs to mdbook logic, so I tried giving claude a try (and was surprised it coming up with a rather concise solution). 

Putting aside if the code is actually good - I think having only the wrapperModule specific options rendered in the docs makes them a lot more usable. Of course ideally we would have a link to the core/default options at the bottom of each wrapperModule.

## Before:

image

## After:

image

### DIFF
--- a/ci/docs/default.nix
+++ b/ci/docs/default.nix
@@ -77,16 +77,23 @@ in
         includeCore = false;
         inherit (config) warningsAreErrors;
         moduleStartsOpen = _: _: true;
-        descriptionStartsOpen = _: _: _: true;
-        descriptionIncluded = _: _: _: true;
+        descriptionStartsOpen =
+          _: _: _:
+          true;
+        descriptionIncluded =
+          _: _: _:
+          true;
       };
     in
     builtins.mapAttrs (buildModuleDocs commonArgs) wlib.modules
     // {
-      default = buildModuleDocs (commonArgs // {
-        excludeFiles = builtins.attrValues defaultSubModules;
-        warningsAreErrors = false;
-      }) "default" wlib.modules.default;
+      default = buildModuleDocs (
+        commonArgs
+        // {
+          excludeFiles = builtins.attrValues defaultSubModules;
+          warningsAreErrors = false;
+        }
+      ) "default" wlib.modules.default;
     };
   config.drv.wrapper_docs = builtins.mapAttrs (buildModuleDocs {
     prefix = "wlib.wrapperModules.";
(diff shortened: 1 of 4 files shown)


BirdeeHub:
I don't want them removed. I want them hidden in the sidebar and present but collapsed in the main view, and only if they are expanded should the values show up in the sidebar.

This is because the options may be different. For example, you can exclude options from the makeWrapper module, modify types via type-merging, etc.

They might not even import all those modules. The neovim module does not import `wlib.modules.symlinkScript` for example.

---

The first issue with the docs currently is that I am filtering based on the file attribute of the module containing a meta.description or meta.maintainer fields, and NO other top level modules are being included at the moment.

---

The second is that the sidebar shortcuts are ALWAYS shown rather than only if their item is expanded.

zenoli:
> This is because the options may be different. For example, you can exclude options from the makeWrapper module, modify types via type-merging, etc.
> 
> They might not even import all those modules. The neovim module does not import `wlib.modules.symlinkScript` for example.

I see. Make sense

zenoli:
This is just pure user feedback on how I think the docs would be more accessible. If you ever get to work on this and see anything usefull in this PR, feel free to use it.

> This is because the options may be different. For example, you can exclude options from the makeWrapper module, modify types via type-merging, etc.

> They might not even import all those modules. The neovim module does not import wlib.modules.symlinkScript for example.

I pushed some commits trying to tackle this. If you just render a single md file, I think mdbook won't let you fold individual sections within this file. Only stuff in separate md-subchapters are foldable. So I told claude to add the imported helper modules as sub-chapters to make them foldable. 

image

(Clicking on neovim itself would display the neovim specific options)

It also seems to respect the `excluded_options` (I verified that `wrapperFunction` and `wrapperImplementation` are not present in the "fish" wrapper.

Don't know if this is just an ad-hoc solution and if it handles the "modified type merging" you mentioned (I don't know this feature).


Also, I added a "default"  supchapter to the Chapter "Helper Modules" and removed the top level Chapter "wlib.modules.default":

image

BirdeeHub:
so, basically, what it comes down to is

There is no built in way to have stuff in the same file, AND expand based on that in the sidebar

---

There might be a way to have multiple files and have them just LOOK like 1 file though? https://rust-lang.github.io/mdBook/format/mdbook.html#including-files

IDK I will have to figure that one out. That may work. Im not sure if the sidebar will pick up on items included in that way? Maybe?

---

BUT there is a way to have 1 file and still change the sidebar however you want

They intentionally allow inclusion of custom JS and CSS for this in the rendering config

And in our html for the page we have <details open> like this </details> and <details> closed like this </details> which are either open or not and marked as such in the html. All we need to do is show or hide sidebar items based on this open or not status

There is at least A way to do this.

How hard it is, I do not yet know.

---

So, the sidebar stuff is all just mdbook and likely JS and CSS skills( of which I have few but query selection is not that hard, I will figure it out eventually when I start doing that)

mdbook is the one detecting and placing those sidebar items for the individual functions on each page

and I do want the docs for each wrapper on 1 page and not 4+

So we will need to fix that in mdbook, or make our own website generator (I would rather not)

---

As for it not including options from all files, that would all be done on the nix side of thing in that ci/docs/per-mod dir

PR #512, opened 2026-05-08
docs(wrapperModules.wezterm): small doc improvements

FluxZA:
I made some edits to the wezterm option docs to make them clearer (in my opinion), particularly if you start reading the docs at one of the later options.

I couldn't quite figure out the use cases for `${placeholder config.outputName}` and `require('nix-info')(defaultval, "path", "to", "item")` in the luaInfo option. If you could provide a use-case for them, I will add an example for the option which includes the use-cases with comments.

### DIFF
--- a/wrapperModules/w/wezterm/module.nix
+++ b/wrapperModules/w/wezterm/module.nix
@@ -57,7 +57,7 @@
       The `defaultval` in this call will be returned if the item you specified does not exist in the luaInfo table.
       This allows you to use a config that works with and without the wrapper's setup.
 
-      To access config values form the final wrapper derivation you can use `''${placeholder config.outputName}` to point to it.
+      To access config values from the final wrapper derivation you can use `''${placeholder config.outputName}` to point to it.
 
       By default, the result of `require('nix-info')` is used as your wezterm config file, see the "wezterm.lua" option for details.
     '';
--- a/wrapperModules/w/wezterm/module.nix
+++ b/wrapperModules/w/wezterm/module.nix
@@ -24,9 +24,10 @@
 
       Note that for this wrapper, the packages will be taken from `config.lua.pkgs`, so make sure that they are available from the derivation you set for the `lua` option.
     '';
-    example = ''
-      # String used to prevent function displaying as `<function>`
+    example = lib.literalMD ''
+      ```nix
       luaEnv = (lp: [ lp.luafilesystem ]);
+      ```
     '';
   };
   options."wezterm.lua" = lib.mkOption {
@@ -48,15 +49,31 @@
     inherit (pkgs.formats.lua { }) type;
     default = { };
     description = ''
-      Defines attributes which are converted to Lua. The converted values are made available to the wezterm config as the result of calling `require('nix-info')`.
-
+      Defines attributes which are converted to Lua.
+      The converted values are made available to the wezterm config as the result of calling `require('nix-info')`.
       The conversion to Lua uses `lib.generators.toLua` which accepts anything other than uncalled nix functions.
 
-      `''${placeholder config.outputName}` is usable here and will point to the final wrapper derivation
+      You can also access specific values from this option by calling `require('nix-info')(defaultval, "path", "to", "item")` in your config.
+      The `defaultval` in this call will be returned if the item you specified does not exist in the luaInfo table.
+      This allows you to use a config that works with and without the wrapper's setup.
 
-      You may also call `require('nix-info')(defaultval, "path", "to", "item")`. This will help prevent indexing errors when querying nested values which may not exist.
+      To access config values form the final wrapper derivation you can use `''${placeholder config.outputName}` to point to it.
 
       By default, the result of `require('nix-info')` is used as your wezterm config file, see the "wezterm.lua" option for details.
+    '';
+    example = lib.literalMD ''
+      ```nix
+      {
+        keys = [
+          {
+            key = "F12";
+            mods = "SUPER|CTRL|ALT|SHIFT";
+            # To get lua expressions, pass a string to lib.generators.mkLuaInline
+            action = lib.generators.mkLuaInline "wezterm.action.Nop";
+          }
+        ];
+      };
+      ```
     '';
   };
   config.constructFiles."wezterm.lua" = {
--- a/wrapperModules/w/wezterm/module.nix
+++ b/wrapperModules/w/wezterm/module.nix
@@ -16,13 +16,17 @@
     type = wlib.types.withPackagesType;
     default = (lp: [ ]);
     description = ''
-      extra lua packages to add to the lua environment for wezterm
+      Extra lua packages to add to the lua environment for wezterm.
+      These packages will be added to package.path and package.cpath in the lua environment.
 
-      value is to be a function from `config.lua.pkgs` to list
+      The value must be a function that can be passed to the [`lua.withPackages`](https://nixos.org/manual/nixpkgs/stable/#lua.withpackages-function) function from `nixpkgs`.
+      This function should take a single argument, and return a list of packages defined as attributes of the argument (see example).
 
-      `config.lua.withPackages config.luaEnv`
-
-      The result will be added to package.path and package.cpath
+      Note that for this wrapper, the packages will be taken from `config.lua.pkgs`, so make sure that they are available from the derivation you set for the `lua` option.
+    '';
+    example = ''
+      # String used to prevent function displaying as `<function>`
+      luaEnv = (lp: [ lp.luafilesystem ]);
     '';
   };
   options."wezterm.lua" = lib.mkOption {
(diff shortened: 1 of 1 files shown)


FluxZA:
On a separate note, is there a way to change how your custom types show up under the type section of an option's doc? If so, I thought it would be nice to include a link to the types documentation in the description.

wrapperModules/w/wezterm/module.nix
```diff
@@ -16,13 +16,17 @@
...
+      This function should take a single argument, and return a list of packages defined as attributes of the argument (see example).
 
-      `config.lua.withPackages config.luaEnv`
-
-      The result will be added to package.path and package.cpath
+      Note that for this wrapper, the packages will be taken from `config.lua.pkgs`, so make sure that they are available from the derivation you set for the `lua` option.
+    '';
+    example = ''
+      # String used to prevent function displaying as `<function>`
+      luaEnv = (lp: [ lp.luafilesystem ]);
```
BirdeeHub:
```suggestion
    example = lib.literalMD ''
      ```nix
      luaEnv = (lp: [ lp.luafilesystem ]);
      ```
```

BirdeeHub (commented):
> `require('nix-info')(defaultval, "path", "to", "item")`

When redefining the content of the config file, you can use that when fetching nested values to avoid indexing errors if the parent was not there.

This way you can have your lua adapt to if the value was provided from nix or not easily.

It is simply convenience.

You can't use it within `luaInfo` very easily because in lua that would in a number of cases be infinite recursion to refer to your own table like that (although I think if you put it in a function you could?) but when redefining the config file it is very useful.

> `${placeholder config.outputName}`

There are a few I can think of, maybe you want your launch command to run something you created via constructFiles, or you wanted to construct some kind of theme file that way or something. IDK.

Honestly, it is more important for `"wezterm.lua".content`, because you can use the placeholders to add more dirs to the `package.path`, but if you wanted to then use `"wezterm.lua".path`, then you could pass that path in via `luaInfo` instead and still have access to it.

Maybe you want to have your config be fennel and build it via config.buildCommand, and then refer to it in the generated config to add it to the lua path or whatever

Regardless, we don't really need to include a direct example of that in the docs, because all the examples are slightly too complicated and specific.

> On a separate note, is there a way to change how your custom types show up under the type section of an option's doc? If so, I thought it would be nice to include a link to the types documentation in the description.

I wish... ALL we get is the `.description` field for the type.

Maybe we could just make the entire description a link? Not sure if our docs would render that correctly? But then that might look really weird in some rendering backends we add down the road. And you are kinda meant to modify the field when you alias types? Kinda? The type system is a bit weird but the description field was definitely not meant for that.

wrapperModules/w/wezterm/module.nix
BirdeeHub: ^ lets use lib.literalMD for this, otherwise, looks good.

FluxZA: Thanks, this is a much nicer way of displaying it. Should I still include the `luaEnv =` part? I had it there because it felt like the string val might be confusing othrewise, but maybe not so with this approach?

FluxZA:
Okay, thanks for the explanations. I agree they wouldn't really fit in an example field, but I will try add some context to the explanations using your examples.

wrapperModules/w/wezterm/module.nix
BirdeeHub: It doesn't matter, up to you.

I haven't been particularly consistent about that myself to be honest.

I guess because normally it just gives the value we should do just the value here too?

But I am pretty sure I did it this way in at least a few places too, and honestly, I would be surprised if that confused someone

FluxZA:
> I wish... ALL we get is the .description field for the type.
>
> Maybe we could just make the entire description a link? Not sure if our docs would render that correctly? But then that might look really weird in some rendering backends we add down the road. And you are kinda meant to modify the field when you alias types? Kinda? The type system is a bit weird but the description field was definitely not meant for that.

I see. I know you are doing some custom stuff with the docgen specifically for wrapper modules, would it be possible to add an additional meta attribute to the types e.g. `typeURL` that wouldn't effect other systems, but would be picked up by the wrapper docs? Otherwise, maybe we can just update the descriptions to include the fact that they are wlib types?

I realise that this isn't necessarily that important, because you can find the info for the types on your own. However, for people like me who are still learning about Nix, it isn't immediately obvious when the type is specific to the wrapper lib. This is because I am not that familiar with all the possible nix types. Having something which immediately signifies a type is from wlib would make the docs a little easier to parse on first look in these cases.

BirdeeHub:
IDK.

I am not sure the module system accepts extra fields on types?

If they do it would be possible, if they do not, then I would rather not as then we would have to add it to the actual value payload somehow and it might get a bit messy.

And we do have our own doc backend which could read that value. It isn't really exactly finished yet, there are some big problems with it I have to work out, but it has been doing the job OK so far.

---

So, actually though, maybe we don't need to attach any value at all

We have 1 attrset with all the types.

I think we could have the docgen detect if this is an unmodified type from that set somehow?

So, like, maybe a thing that could happen.

---

There are some other things that are much higher priority for the docgen though at the moment.

I kinda would rather solve those first

Namely, first and foremost

It was meant to find all the options IMPORTED by modules with `meta.description` or `meta.maintainer` and group those under those files.

Instead it finds ONLY options IN modules with `meta.description` or `meta.maintainer`

So, like, thats a big problem, and one which will need to be solved in some of the same files as I would be adding this detection thing to.

The other problem with the docs I want to solve is add a postprocessing thing that makes the sidebar better, but that likely will be done via mdbook and some JS/CSS and not the actual generator so that wouldn't be a blocker for me addressing this.

FluxZA:
Okay I see. I definitely don't think this needs to be a priority, it was just an idea I had and I wasn't sure if it was viable.

BirdeeHub:
It kinda is, just not straightforwardly. I guess it is one to add to the list of things to try for improving the docs.

FluxZA:
Are the docgen issues something you are looking for help with, or is it easier for you to address them yourself?

BirdeeHub:
No idea

https://github.com/BirdeeHub/nix-wrapper-modules/tree/main/ci/docs/per-mod

^ does any of that code make sense to you?

If so, feel free to give it a try. But I feel it is likely I will be the one who has to address that... Simply because it is complicated AF...

But that directory is what actually grabs the options and renders them into markdown for the pages.

But basically, doing said collection, grouping/sorting, rendering is so complicated, it is likely I will be the one who ends up addressing it

I wish I didn't have to do it, but the docs were so hard to read when the options were not separated by the module that provides them that something needed to be done.

---

I feel like the sidebar thing shouldn't be that hard?

They have an additional-js and additional-css and stuff

https://rust-lang.github.io/mdBook/format/configuration/renderers.html#html-renderer-options

Which can be provided here:

https://github.com/BirdeeHub/nix-wrapper-modules/blob/dc5184095ad488e937ec308b52c9c0b218959d8b/ci/docs/default.nix#L101

Which can fix the sidebar thing theoretically?

I feel like that would have the biggest impact on readability?

https://github.com/BirdeeHub/nix-wrapper-modules/pull/504

^ there was some discussion about some of this stuff here too.

---

Basically, there are 3 things that need improving

1, the thing where options are excluded if not marked with a _file with a meta.description or meta.maintainers

2, the sidebar spam (without removing the options from the page entirely) by making the options only appear in the sidebar when the section they are in is expanded

3, I guess this type description link thing.

---

When the first of those 3 things are fixed, I can turn the docgen for individual modules into either a helper module or a helper function that 3rd party wrapper modules can use too actually.

the other 2 of those things can be fixed either before or after that, doesn't matter, but it would likely be easier to add 3 after 1, because they will need to be addressed in the same files. 2 should be fixed in the mdbook module and not the rendering logic, so it doesn't matter if that is fixed before or after.

FluxZA:
I will take a look and see what feels doable. If I feel like I can work on either issue I will create a draft MR to discuss and so that you know someone is looking into it.

On that note, thanks again for finishing the zsh module, I got blindsided by the amount of work/life that suddenly hit me and appreciate that it was finalised and merged.

wrapperModules/w/wezterm/module.nix
```diff
@@ -31,25 +36,44 @@
...
+      The converted values are made available to the wezterm config as the result of calling `require('nix-info')`.
+      The conversion to Lua uses `lib.generators.toLua` which accepts anything other than uncalled nix functions.
 
-      This means, by default, this will act like your wezterm config file, unless you want to add some lua in between there.
+      You can also access specific values from this option by calling `require('nix-info')(defaultval, "path", "to", "item")` in your config.
+      The `defaultval` in this call will be returned if the item you specified does not exist in the luaInfo table.
+      This allows you to use a config that works with and without the wrapper's setup.
 
-      `''${placeholder config.outputName}` is useable here and will point to the final wrapper derivation
+      To access config values form the final wrapper derivation you can use `''${placeholder config.outputName}` to point to it.
```
BirdeeHub:
```suggestion
      To access config values from the final wrapper derivation you can use `''${placeholder config.outputName}` to point to it.
```

wrapperModules/w/wezterm/module.nix
```diff
@@ -31,25 +36,44 @@
...
 
-      You may also call `require('nix-info')(defaultval, "path", "to", "item")`
-
-      This will help prevent indexing errors when querying nested values which may not exist.
+      By default, the result of `require('nix-info')` is used as your wezterm config file, see the "wezterm.lua" option for details.
```
BirdeeHub, line 62: I want to say the full content, which is `return require('nix-info')` somewhere but wording is hard.

wrapperModules/w/wezterm/module.nix
```diff
@@ -31,25 +36,44 @@
       content = lib.mkOptionDefault "return require('nix-info')";
     };
     default = { };
-    description = "The wezterm config file. provide `.content`, or `.path`";
+    description = ''
+      The wezterm config file.
+
+      By default the content of the `luaInfo` option is used and nothing has to be provided here.
```
FluxZA, line 42:
```suggestion
      By default the full content of the `luaInfo` option is used and nothing has to be provided here.
      This is accomplished by the default  content  being set to `return require('nix-info')`.
      See the luaInfo option for details on `require('nix-info')`.
```

wrapperModules/w/wezterm/module.nix
FluxZA, line 62:
```suggestion
      By default, the full value of this option is used as your wezterm config file.
      See the "wezterm.lua" option for details.
```

FluxZA (commented):
Maybe this can provide the context you want?

wrapperModules/w/wezterm/module.nix
BirdeeHub: ehhh I think what you had before was better than doing this change.

Lets Just fix the form -> from typo and merge it.

wrapperModules/w/wezterm/module.nix
BirdeeHub: ^ lets just fix this, its good enough I think otherwise.

PR #528, opened 2026-05-16
feat(wrapperModules.television): init

allen-liaoo:
Wrapper module for [Television](https://alexpasmantier.github.io/television/). 
Not sure what is the best/recommended way to handle TOML, so I just imitated what other wrapper modules did.

### DIFF
--- a/wrapperModules/t/television/module.nix
+++ b/wrapperModules/t/television/module.nix
@@ -12,7 +12,11 @@ let
     typeName = "TOML";
   };
   isPathLike =
-    x: builtins.isPath x || (builtins.isString x && lib.hasPrefix "/" x) || lib.isStorePath x;
+    x:
+    builtins.isPath x
+    || (lib.isStringLike x && !builtins.isString x)
+    || (builtins.isString x && lib.hasPrefix "/" x)
+    || lib.isStorePath x;
 in
 {
   imports = [ wlib.modules.default ];
@@ -23,7 +27,7 @@ in
       description = "Television configuration options.";
     };
     channels = lib.mkOption {
-      type = types.lazyAttrsOf (types.either types.path tomlFmtType);
+      type = types.lazyAttrsOf (types.either wlib.types.stringable tomlFmtType);
       default = { };
       description = "Television channels to install.";
     };
@@ -36,7 +40,7 @@ in
       '';
     };
     themes = lib.mkOption {
-      type = types.lazyAttrsOf (types.either types.path tomlFmtType);
+      type = types.lazyAttrsOf (types.either wlib.types.stringable tomlFmtType);
       default = { };
       description = "Themes of television to install.";
     };
@@ -58,6 +62,12 @@ in
   };
   config = {
     package = lib.mkDefault pkgs.television;
+    passthru.generatedThemesDir = "${
+      config.wrapper.${config.configDrvOutput}
+    }/${baseNameOf config.themesDir}";
+    passthru.generatedChannelsDir = "${
+      config.wrapper.${config.configDrvOutput}
+    }/${baseNameOf config.channelsDir}";
     flags = {
       "--config-file" = lib.mkIf (config.settings != { }) config.constructFiles.generatedConfig.path;
       "--cable-dir" = lib.mkIf (config.channels != { }) config.channelsDir;
@@ -76,7 +86,7 @@ in
       output = lib.mkOverride 0 config.configDrvOutput;
       ${if isPathLike v then null else "content"} = builtins.toJSON v;
       "builder" =
-        if isPathLike v then ''cp ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
+        if isPathLike v then ''ln -s ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
     }) config.channels
     // builtins.mapAttrs (n: v: {
       key = "theme_${n}";
@@ -84,7 +94,7 @@ in
       output = lib.mkOverride 0 config.configDrvOutput;
       ${if isPathLike v then null else "content"} = builtins.toJSON v;
       "builder" =
-        if isPathLike v then ''cp ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
+        if isPathLike v then ''ln -s ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
     }) config.themes;
     meta.maintainers = [ wlib.maintainers.allen-liaoo ];
   };
--- a/wrapperModules/t/television/module.nix
+++ b/wrapperModules/t/television/module.nix
@@ -1,0 +1,91 @@
+{
+  config,
+  pkgs,
+  wlib,
+  lib,
+  ...
+}:
+let
+  types = lib.types;
+  tomlFmtType = wlib.types.structuredValueWith {
+    nullable = false;
+    typeName = "TOML";
+  };
+  isPathLike =
+    x: builtins.isPath x || (builtins.isString x && lib.hasPrefix "/" x) || lib.isStorePath x;
+in
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    settings = lib.mkOption {
+      type = tomlFmtType;
+      default = { };
+      description = "Television configuration options.";
+    };
+    channels = lib.mkOption {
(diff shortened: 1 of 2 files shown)


wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    settings = lib.mkOption {
+      type = tomlFmtType;
+      default = { };
+      description = "Television configuration options.";
+    };
+    channels = lib.mkOption {
+      type = types.lazyAttrsOf (types.either types.path tomlFmtType);
```
BirdeeHub:
```suggestion
      type = types.lazyAttrsOf (types.either wlib.types.stringable tomlFmtType);
```

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+    channelsDir = lib.mkOption {
+      type = types.str;
+      readOnly = true;
+      default = "${placeholder config.configDrvOutput}/${config.binName}-channels";
+      description = ''
+        The placeholder for the location of the channels directory.
+      '';
+    };
+    themes = lib.mkOption {
+      type = types.lazyAttrsOf (types.either types.path tomlFmtType);
```
BirdeeHub:
```suggestion
      type = types.lazyAttrsOf (types.either wlib.types.stringable tomlFmtType);
```

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+        output = config.configDrvOutput;
+      };
+    }
+    // builtins.mapAttrs (n: v: {
+      key = "channel_${n}";
+      relPath = lib.mkOverride 0 "${config.binName}-channels/${n}.toml";
+      output = lib.mkOverride 0 config.configDrvOutput;
+      ${if isPathLike v then null else "content"} = builtins.toJSON v;
+      "builder" =
+        if isPathLike v then ''cp ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
```
BirdeeHub:
```suggestion
        if isPathLike v then ''ln -s ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
```

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+    }) config.channels
+    // builtins.mapAttrs (n: v: {
+      key = "theme_${n}";
+      relPath = lib.mkOverride 0 "${config.binName}-themes/${n}.toml";
+      output = lib.mkOverride 0 config.configDrvOutput;
+      ${if isPathLike v then null else "content"} = builtins.toJSON v;
+      "builder" =
+        if isPathLike v then ''cp ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
```
BirdeeHub:
```suggestion
        if isPathLike v then ''ln-s ${v} "$2"'' else ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
```

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+    };
+  };
+  config = {
+    package = lib.mkDefault pkgs.television;
+    flags = {
+      "--config-file" = lib.mkIf (config.settings != { }) config.constructFiles.generatedConfig.path;
+      "--cable-dir" = lib.mkIf (config.channels != { }) config.channelsDir;
+    };
+    constructFiles = {
+      generatedConfig = {
```
BirdeeHub, line 76: This is the way we have been handling toml yes.

The mapAttrs you did for the other 2 sets is somewhat more complex because you also wanted to accept paths, but also construct files if not a path.

But it looks good I think.

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+    };
+    configDrvOutput = lib.mkOption {
+      type = types.str;
+      default = config.outputName;
+      description = ''
+        The derivation output name the generated configuration will be output to.
+      '';
+    };
+  };
+  config = {
```
BirdeeHub, line 63:
```suggestion
  config = {
    passthru.generatedThemesDir = "${config.wrapper.${config.configDrvOutput}}/${baseNameOf config.themesDir}";
    passthru.generatedChannelsDir = "${config.wrapper.${config.configDrvOutput}}/${baseNameOf config.channelsDir}";
```

BirdeeHub: Gives an easier way to access the non-placeholder versions of the dirs of these paths.

Of course they could also grab `configFiles.<somename>.outPath` to get it but like, this is nicer.

wrapperModules/t/television/module.nix
```diff
@@ -0,0 +1,91 @@
...
+  ...
+}:
+let
+  types = lib.types;
+  tomlFmtType = wlib.types.structuredValueWith {
+    nullable = false;
+    typeName = "TOML";
+  };
+  isPathLike =
+    x: builtins.isPath x || (builtins.isString x && lib.hasPrefix "/" x) || lib.isStorePath x;
```
BirdeeHub:
```suggestion
    x: builtins.isPath x || (lib.isStringLike x && ! builtins.isString x) || (builtins.isString x && lib.hasPrefix "/" x) || lib.isStorePath x;
```

BirdeeHub: This honestly may be worth adding as `wlib.isPathLike` at some point but that would be a separate PR

wrapperModules/t/television/module.nix
BirdeeHub: ln -s allows them to use an impure path without needing to wlib.mkOutOfStoreSymlink (and we also adjusted the type to wlib.types.stringable to allow them to pass drvs and just a regular string in addition to a store path)

BirdeeHub (commented):
Some improvements about the types of paths things can accept and some convenience things, but it looks good!

allen-liaoo:
Thanks for the feedback! I made the adjustments based on your comments.
An additional thing: Many channels need binaries to run correctly. Many community maintained channels requires `bat`, some requires `rq`, `fd`, etc. (See https://github.com/alexpasmantier/television/tree/main/cable/unix). Is the recommended way of getting these binaries for television be for the wrapper user themselve to wrap the wrapper package with these dependencies? Or should this module expose options to collect dependencies to wrap the package?

BirdeeHub:
It appears there is a potentially infinite number of channels that are possible.

It would be nice if we could parse the channel's TOML file and get the requirements automatically, but in the majority of cases that would be an IFD and the name in requirements may not match the name in nixpkgs. We don't have a great way of managing this without maintaining a list of possible channels and their dependencies as they are named in nixpkgs.

I think we should probably have people be just adding the required dependencies via config.extraPackages?

Let me run this and stuff, but it looks good, if it all works as expected, should be ready to merge!

allen-liaoo:
oh I wasn't aware there is the extraPackages option. That was what I was looking for haha
For testing, feel free to check out my working tv wrapper config: https://github.com/allen-liaoo/nix-packages/blob/master/wrappers/television.nix

BirdeeHub:
Also, fair warning, I am going to change it to runtimePkgs and accept more options (currently it only suffixes and I want people to be able to decide which happens per item)

It will mostly be a drop in replacement. It will be changed to a list of wlib.types.spec types, so that it accepts a list of either (just the packages, or a set with some options and the package). Will link the PR when I finish working on it, and then I will try this out.

There will be a deprecation warning and all that stuff too. I am very nearly done with it or I would review this before opening the PR haha

https://github.com/BirdeeHub/nix-wrapper-modules/pull/539

^ this one

BirdeeHub:
```nix
:bl outputs.wrappers.television.wrap [
  { inherit pkgs; }
  ({ pkgs, lib, ... }: {
    channels.tldr = {
      metadata = { name = "tldr"; };
      source.command = "${lib.getExe pkgs.tldr} --list";
      preview.command = "${lib.getExe pkgs.tldr} '{}'";
      keybindings.ctrl-e = "actions:open";
      actions.open = {
        command = "${lib.getExe pkgs.tldr} '{}'";
        mode = "execute";
      };
    };
  })
]
```

^ seems to work just fine. LGTM

You can either put the paths directly, or include them via extraPackages (soon to be runtimePkgs)

allen-liaoo:
Thanks for merging it!

PR #532, opened 2026-05-17
fix(modules): use `gnused` instead of unfree `replace`

MakeShiftArtist:
It is a simple, drop in replacement that fixes builds failing if `config.allowUnfree = false`

`nixos-unstable` had a recent commit [748036a](https://github.com/NixOS/nixpkgs/commit/748036a3df1a94f5f97dd15b6af0dcae24f126d6) that actually set the license, which is `unfree`. This means all wrappers that don't allow unfree packages will simply fail to build.

This fixes it by using `gnused`. I ran tests using the built in ci
```sh
check
```

And using
```sh
nix flake check -Lv ./ci
```

Both resulted in all checks passing.

I also tested by running the default template before and after. Before, it fails saying replace-2.24 requires unfree, and after is successfully builds and runs.

### DIFF
--- a/modules/symlinkScript/module.nix
+++ b/modules/symlinkScript/module.nix
@@ -93,7 +93,7 @@
                         # Remove symlink and create a real file with patched content
                         rm "$file"
                         # Use replace-literal which works for both text and binary files
-                        ${pkgs.replace}/bin/replace-literal "$oldPath" "$newPath" < "$target" > "$file"
+                        ${pkgs.gnused}/bin/sed "s|$oldPath|$newPath|g" "$target" > "$file"
                         # Preserve permissions
                         chmod --reference="$target" "$file"
                       fi


BirdeeHub:
https://github.com/BirdeeHub/nix-wrapper-modules/pull/525

BirdeeHub:
we are looking into what to replace it with.

gnused is a reasonable option

We were working on a test for it

MakeShiftArtist:
Not sure how I missed that other PR. Closed this as a duplicate, to avoid splitting discourse

BirdeeHub:
Its all good. Thanks for pointing out that gnused (gnused specifcially, rather than the stdenv sed which is different on mac) works for binaries, I forgot.

This looks like a reasonable option for me, but we should still update the comments, and probably also add a test that whatever we are using does in fact support binary files, in case we change it again down the road and forget to check ourselves.

BirdeeHub:
btw if the other person doesn't respond for a few days feel free to reopen, or I guess Ill steal it and add it. Personally I do allowUnfree because why not but the CI breakage can be annoying, for people who set up CI for their configs anyway. They seemed eager to help tho, so it should be merged soon!

Your input is appreciated and stuff :)

PR #539, opened 2026-05-19
feat(modules.makeWrapper): extraPackages -> runtimePkgs

BirdeeHub:
renamed extraPackages -> runtimePkgs and changed its type to accept a wlib.types.spec type instead.

The same changes were performed for runtimeLibraries also, they have been renamed to runtimeLibs

The options accept a list of either packages or sets with a `.data` field and optional prefix, name, before, after fields.

This will not be breaking for most users (outside of the renaming), but it may be breaking for those consuming the resulting config value for other purposes, who will now need to grab the .data field from the result. If all you were doing was setting it, then this will not affect you, beyond renaming the option.

The motivation was to allow fine-grained selection of if the value should be suffixed or prefixed to the path.

In addition, the makeWrapper module saw some updates to its excluded_options feature, used at the import site in some modules, particularly zsh and fish. You can now disable options in either the top level OR the variants, whereas before, you could only exclude both.

https://github.com/BirdeeHub/nix-wrapper-modules/discussions/2#discussioncomment-16747150

^ this has been on the list for a few weeks (not that there really is 1 particular list, it is just something that seemed like a good idea)


BirdeeHub:
Also, @zenoli now that I have gotten to try it a few times, I can say I do actually like the new test helpers. Eventually more will be converted haha

zenoli:
@BirdeeHub btw: I presented your project at the [Nix Zurich ZHF event](https://zurich.nix.ug/) in form of a lightning talk last weekend :-)

Response was quite positive. The discussion afterwards evolved around whether we could automatically derive wrapper modules from hm/nixos modules (to which my 2 cents were that this would be extremely hard to do automatically for various reasons).

BirdeeHub:
I mean someone did it, it is in the readme haha

But yes it is hard. And their implementation only works for linux, and it would require taking home manager as an input, and doing a lot more processing per module.

I need to update the readme that I want the bubblewrap helper when it is added to also be able to use sandbox-exec or whatever it is on mac...

PR #546, opened 2026-05-19
feat(wrapperModules.ghostty): init

MichaelJacobRoss:
Settings are written to a generated Nix store config file and loaded via `--config-default-files=false --config-file=<path>`, so the host config is never loaded -- including on machines that already have their own Ghostty config. The host config path varies by platform: `~/.config/ghostty/config` on Linux, and `~/Library/Application Support/com.mitchellh.ghostty/config` on macOS.

On macOS, `ghostty-bin` is used by default, which is the pre-built binary distribution recommended for nix-darwin by the [Ghostty installation docs](https://ghostty.org/docs/install/binary).

**Known limitations**

- `ghostty +show-config` and `ghostty +validate-config` will show the host config rather than the generated one. Ghostty's action-mode option parsers reject `--config-file`, so there is no way to point them at the generated config.

- If you pass an additional `--config-file` flag at runtime, Ghostty merges it on top of the generated config (later files take precedence for conflicting keys).

### DIFF
--- a/wrapperModules/g/ghostty/module.nix
+++ b/wrapperModules/g/ghostty/module.nix
@@ -1,0 +1,109 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  toGhosttyConf = lib.generators.toKeyValue {
+    listsAsDuplicateKeys = true;
+    mkKeyValue = lib.generators.mkKeyValueDefault {
+      mkValueString = v: if builtins.isBool v then lib.boolToString v else toString v;
+    } " = ";
+  };
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options.settings = lib.mkOption {
+    type =
+      let
+        atom = lib.types.oneOf [
+          lib.types.bool
+          lib.types.float
+          lib.types.int
+          lib.types.str
+        ];
+      in
+      lib.types.attrsOf (lib.types.either (lib.types.listOf atom) atom);
+    default = { };
+    example = lib.literalExpression ''
+      {
+        font-size = 14;
+        theme = "Catppuccin Mocha";
+        window-decoration = false;
+        keybind = [
+          "ctrl+a>-=new_split:down"
+          "ctrl+a>==new_split:right"
+        ];
+      }
+    '';
+    description = ''
+      Ghostty configuration written to a generated config file. The wrapper
+      passes {option}`--config-default-files=false` and
+      {option}`--config-file=<generated>`, so the host config is never loaded
+      -- including when the wrapper is used ephemerally on a machine that has
+      its own Ghostty config. The host config path varies by platform:
+      {file}`~/.config/ghostty/config` on Linux, and
+      {file}`~/Library/Application Support/com.mitchellh.ghostty/config` on
+      macOS.
+      See <https://ghostty.org/docs/config/reference> for all available options.
+
+      Note: if you pass an additional {option}`--config-file` flag at runtime,
+      Ghostty will merge it on top of the generated config (later files take
+      precedence for conflicting keys).
+
+      Note: {command}`ghostty +show-config` and {command}`ghostty
+      +validate-config` bypass the generated config entirely (their option
+      parsers reject {option}`--config-file`), so they will show the host
+      config if one exists -- this affects both installed and ephemeral use.
+
+      On macOS, `ghostty-bin` is used by default, which is the pre-built
+      binary distribution [recommended for nix-darwin](https://ghostty.org/docs/install/binary)
+      by the Ghostty installation docs.
+    '';
+  };
+
+  config = {
+    package = lib.mkDefault (if pkgs.stdenv.isDarwin then pkgs.ghostty-bin else pkgs.ghostty);
+    constructFiles.ghosttyConfig = {
+      content = toGhosttyConf config.settings;
+      relPath = "${config.binName}-config";
+    };
+    addFlag = [
+      "--config-default-files=false"
+      "--config-file=${config.constructFiles.ghosttyConfig.path}"
+    ];
+    # Ghostty has no environment variable for specifying a config file, so CLI
+    # flags are the only option: --config-default-files=false prevents loading
+    # the host's ~/.config/ghostty/config (including on machines where one
+    # exists), and --config-file points at the generated one.
+    #
+    # However, Ghostty +actions and --help have their own minimal Options
+    # struct with no _diagnostics field.  Any flag unrecognised by the action
+    # parser causes error.InvalidField and a silent exit with no output.
+    # argv0type detects these and execs Ghostty directly without injecting any
+    # flags.
+    #
+    # Consequence: +show-config and +validate-config will show the host's
+    # ~/.config/ghostty/config if present, not the generated config, because
+    # their option parsers also reject --config-file.  This applies equally to
+    # installed and ephemeral use.
+    argv0type =
+      let
+        binPath = lib.escapeShellArg config.wrapperPaths.input;
+      in
+      cmd: ''
+        for _ghostty_arg in "$@"; do
+          case "$_ghostty_arg" in
+            +*|--help) exec -a "$0" ${binPath} "$@";;
+          esac
+        done
+        exec -a "$0" ${cmd}
+      '';
+    meta.description = "Ghostty terminal emulator";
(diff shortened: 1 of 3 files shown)


MichaelJacobRoss:
Converted to draft — this is blocked on #548 which fixes the `wlib-formatting` CI crash on macOS. Once that merges I'll rebase and mark ready.

MichaelJacobRoss:
Please excuse the number of edits and CI runs on this PR. I previously thought that this PR was blocked by a CI-related change so every time I needed to test that change, I'd re-run the CI pipeline on the other PR and then rebase here.

wrapperModules/g/ghostty/module.nix
```diff
@@ -0,0 +1,102 @@
...
+    # Consequence: +show-config and +validate-config will show the host's
+    # ~/.config/ghostty/config if present, not the generated config, because
+    # their option parsers also reject --config-file.  This applies equally to
+    # installed and ephemeral use.
+    argv0type =
+      let
+        binPath = lib.escapeShellArg config.wrapperPaths.input;
+      in
+      cmd: ''
+        for _ghostty_arg in "$@"; do
```
BirdeeHub, line 98: What is this for?

Are we intercepting args to fix their bug?

BirdeeHub (commented):
https://github.com/BirdeeHub/nix-wrapper-modules/pull/393

^ This PR might have some useful things in it for you (The last PR attempt, which ran into similar bugs in ghostty but was otherwise quite good)

wrapperModules/g/ghostty/module.nix
MichaelJacobRoss: Yes, exactly. This is the client-side workaround for the bug discussed in #393.

The argv0type block scans $@ for any arg that looks like a +action or --help / -h, and if it finds one, execs the real Ghostty binary directly (skipping flag injection). For normal invocations, it falls through and the flags are injected as usual.

This means the wrapper is fully usable now without waiting for Ghostty upstream to fix it. ghostty --font-size=14 works, and ghostty +show-config also works (though it'll show the host config rather than the generated one, as documented in the option description).

It's up to you if you think the known limitations in the PR description mean we ought to wait for the upstream fix, but personally, I'm already using this setup and the only behavioral difference compared to the non-wrapped version of ghostty is pretty niche.

MichaelJacobRoss: I noticed the XDG concern from #393 and I think argv0type is probably the cleanest solution we're going to get for a while without that env bleed. Rather than touching XDG_CONFIG_HOME at all, it just detects the problematic invocation forms at runtime and bypasses flag injection entirely for those cases. The tradeoff is that +show-config and friends show the host config instead of the generated one, but that's documented in the option description and seems preferable to polluting the terminal's environment.

wrapperModules/g/ghostty/module.nix
```diff
@@ -0,0 +1,102 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  toGhosttyConf = lib.generators.toKeyValue {
+    mkKeyValue = lib.generators.mkKeyValueDefault {
```
BirdeeHub, line 11:
```suggestion
    mkKeyValue = lib.generators.mkKeyValueDefault {
      listsAsDuplicateKeys = true;
```

wrapperModules/g/ghostty/module.nix
```diff
@@ -0,0 +1,102 @@
...
+
+  options.settings = lib.mkOption {
+    type = lib.types.attrsOf (
+      lib.types.oneOf [
+        lib.types.bool
+        lib.types.float
+        lib.types.int
+        lib.types.str
+      ]
+    );
```
BirdeeHub:
```suggestion
    type = let
      atom = lib.types.oneOf [
        lib.types.bool
        lib.types.float
        lib.types.int
        lib.types.str
      ];
    in lib.types.attrsOf (lib.types.either (lib.types.listOf atom) atom);
```

BirdeeHub (commented):
Some changes requested to the other PR to allow

```nix
  settings.keybind = [
    "ctrl+a>-=new_split:down"
    "ctrl+a>==new_split:right"
  ];
```

for

```
keybind=ctrl+a>-=new_split:down
keybind=ctrl+a>==new_split:right
```

and

```nix
  settings.config-file = [
    "${./more}"
    "${./files}"
  ];
```

wrapperModules/g/ghostty/module.nix
BirdeeHub: Alternatively, we could use `inherit (pkgs.formats.keyValue { listsAsDuplicateKeys = true; }) type;`

Or even add https://github.com/NixOS/nixpkgs/blob/2bc4da5851acf09c933e582394e122fb5402ea4e/pkgs/pkgs-lib/formats.nix#L417-L446 <-- the type to wlib.types (because they have neglected to give us a way to use it without grabbing it from pkgs) or open a PR for them to move it in nixpkgs (If you do that, give [this PR](https://github.com/NixOS/nixpkgs/pull/516798) a shoutout too! That is our `wlib.types.structuredValueWith` type)

MichaelJacobRoss:
@BirdeeHub ah sorry I'll look in more detail tomorrow and get back to you

wrapperModules/g/ghostty/module.nix
MichaelJacobRoss: Applied both suggestions: listsAsDuplicateKeys = true in the generator, and the type updated to attrsOf (either (listOf atom) atom). I also updated the example and tests to demonstrate the duplicate-key behaviour with a keybind list.

I went with the explicit either union rather than inheriting from pkgs.formats.keyValue to keep the type out of pkgs. If your nixpkgs PR (#516798) lands and that type becomes available in lib without needing a pkgs instance, happy to consolidate here at that point.

BirdeeHub:
I think I am OK with the intercepting of those args as a workaround for now. They have at least acknowledged the issue, and people could set argv0type themselves if they don't want it.

One question, what happened to our filesToPatch we had there before? Are they not necessary due to using ghostty-bin on mac? Or is there some reason why they were removed?

I think other than that this is good though, but I am curious about where that went, because you had them in your original version of the PR.

Once we add those back, or you explain why we don't need them anymore, I guess we should probably just go ahead and merge it rather than waiting around for ghostty to fix the bug, its been long enough...

wrapperModules/g/ghostty/module.nix
BirdeeHub: #516798

This is actually a different type entirely, I was just mentioning it.

jonas-elhs:
Is there anything holding this back?

BirdeeHub:
> One question, what happened to our filesToPatch we had there before? Are they not necessary due to using ghostty-bin on mac? Or is there some reason why they were removed?
> 
> I think other than that this is good though, but I am curious about where that went, because you had them in your original version of the PR.
> 
> Once we add those back, or you explain why we don't need them anymore, I guess we should probably just go ahead and merge it rather than waiting around for ghostty to fix the bug, its been long enough...

^ This was.

```nix
  config.filesToPatch = [
    "share/applications/*.desktop"
    "share/dbus-1/services/com.mitchellh.ghostty.service"
    "share/systemd/user/app-com.mitchellh.ghostty.service"
  ];
```

^ Where did these go? You had them here before and now they are not.

Sorry for taking so long with this, I am back now I had to take a break.

wrapperModules/g/ghostty/module.nix
```diff
@@ -0,0 +1,109 @@
...
+      parsers reject {option}`--config-file`), so they will show the host
+      config if one exists -- this affects both installed and ephemeral use.
+
+      On macOS, `ghostty-bin` is used by default, which is the pre-built
+      binary distribution [recommended for nix-darwin](https://ghostty.org/docs/install/binary)
+      by the Ghostty installation docs.
+    '';
+  };
+
+  config = {
```
BirdeeHub, line 68:
```suggestion
  config = {
    filesToPatch = [
      "share/applications/*.desktop"
      "share/dbus-1/services/com.mitchellh.ghostty.service"
      "share/systemd/user/app-com.mitchellh.ghostty.service"
    ];
```

BirdeeHub:
Since it is mostly my fault this has taken so long, I am just going to add it and then merge it.

LGTM. Hopefully we get to remove the workaround soon!

PR #551, opened 2026-05-23
feat(wrapperModules.iamb): init

aliaslion:
Adds wrapper module for iamb, a TUI Matrix client with vi-like keybinds.

Note that I'm not completely sure what `generatedConfig.output` and `.placeholder` do, I cobbled this together from the Alacritty and Yazi modules. Would appreciate some info about that :)

### DIFF
--- a/maintainers/default.nix
+++ b/maintainers/default.nix
@@ -111,4 +111,9 @@
     github = "allen-liaoo";
     githubId = 16383622;
   };
+  aliaslion = {
+    name = "aliaslion";
+    github = "aliaslion";
+    githubId = "122117018";
+  };
 }
--- a/wrapperModules/i/iamb/module.nix
+++ b/wrapperModules/i/iamb/module.nix
@@ -1,0 +1,35 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    settings = lib.mkOption {
+      type = wlib.types.structuredValueWith {
+        nullable = false;
+        typeName = "TOML";
+      };
+      default = { };
+      description = ''
+        Configuration of iamb.
+        See {manpage}`iamb(5)` or <https://iamb.chat/configure.html>
+
+        Note: at least one profile is required for startup.
+      '';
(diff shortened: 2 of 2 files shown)


BirdeeHub:
> Note that I'm not completely sure what generatedConfig.output and .placeholder do, I cobbled this together from the Alacritty and Yazi modules. Would appreciate some info about that :)

So, basically, when we have many files we want to make into a directory, the `placeholder` value is used as the location you group them together in in the derivation.

Placeholders are weird values. They point at the associated output in whatever derivation they END UP in.

So, outside of that derivation, you can't really use them.

You don't need most of that for this module.

using `constructFiles` gives you both placeholder paths, and final ones usable from outside the module.

Since we are only generating 1 file, that is probably enough, but passing the thing out via passthru is still nice.

```nix
{
  config,
  lib,
  wlib,
  pkgs,
  ...
}:
{
  options = {
    settings = lib.mkOption {
      type = wlib.types.structuredValueWith {
        nullable = false;
        typeName = "TOML";
      };
      default = { };
      description = ''
        Configuration of iamb.
        See {manpage}`iamb(5)` or <https://iamb.chat/configure.html>

        Note: profiles is required for first startup.
      '';
      example.profiles.myuser.user_id = "@user2:example.com";
    };
  };
  config.package = lib.mkDefault pkgs.iamb;
  config.flags."--config-directory" = dirOf (dirOf config.constructFiles.config.path);
  config.passthru.generatedConfig = dirOf (dirOf config.constructFiles.config.outPath);
  config.constructFiles.config = {
    relPath = "${config.binName}-config/iamb/config.toml";
    builder = ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
  };
  config.meta.maintainers = [ wlib.maintainers.aliaslion ];
}
```

Maybe something like that?

aliaslion:
OK, thank you, I think I understand it now. Also didn't know about `dirOf`.

BirdeeHub:
> Also didn't know about dirOf.

It makes sense here because we have a dir, but it just has 1 file in it.

The docs do not seem to say anything about any other files. So, I think this is good!

PR #553, opened 2026-05-23
feat(wrapperModules.quickshell): add option to set component's module

ormoyo:
Fixes the issue fixed on #552 but in a declarative manner. 

Basically, to allow nested components inside sub-directories to be imported with `import qs.xxx.xxx`.

My extra solution is to add the option `components.<name>.module` and make components an attrset of specs. (To make no breaking changes)

### DIFF
--- a/wrapperModules/q/quickshell/module.nix
+++ b/wrapperModules/q/quickshell/module.nix
@@ -7,7 +7,8 @@
 }:
 let
   inherit (lib)
-    mapAttrs'
+    isStringLike
+    mapAttrs
     mkDefault
     mkIf
     mkOption
@@ -17,34 +18,15 @@ let
   isLinkable = wlib.types.linkable.check;
   makeForce = lib.mkOverride 0;
 
-  componentModule =
-    { name, config, ... }:
-    {
-      options = {
-        name = mkOption {
-          type = types.str;
-          default =
-            let
-              firstChar = builtins.substring 0 1 name;
-              rest = builtins.substring 1 (-1) name;
-            in
-            if (isLinkable config.data) then
-              (builtins.baseNameOf config.data)
-            else
-              (lib.toUpper firstChar) + rest + ".qml";
-          description = "The name of this component (either filename or directory name)";
-        };
-        data = mkOption {
-          type = types.either wlib.types.linkable types.lines;
-          description = "The component's inlined text or path";
-        };
-        module = mkOption {
-          type = types.nullOr types.str;
-          default = null;
-          description = "The component's module, to be imported by `import qs.<module>`";
-        };
-      };
-    };
+  componentType = types.submodule {
+    freeformType = types.lazyAttrsOf (
+      types.oneOf [
+        wlib.types.linkable
+        types.lines
+        (componentType // { description = "nested components"; })
+      ]
+    );
+  };
 in
 {
   imports = [ wlib.modules.default ];
@@ -60,9 +42,30 @@ in
       '';
     };
     components = mkOption {
-      type = types.attrsOf (wlib.types.spec componentModule);
+      type = componentType;
       default = { };
       description = "Quickshell components to include in the configuration";
+      example = lib.literalExpression ''
+        {
+          some.path."Bar.qml" = ./some-widget.qml;
+          light.clock = '''
+            Text {
+              text: "hello world"
+            }
+          ''';
+          dark.clock = "/etc/quickshell/Clock.qml";
+          lockscreensDir =
+            let
+              repo = pkgs.fetchFromGitHub {
+                owner = "Darkkal44";
+                repo = "qylock";
+                rev = "cde4d11e9e3d385620becdc877a0521e40a55e47";
+                hash = "sha256-17kRwrkdfe+hJdChMxove73zNCKcSi0nmSrO8Fh8hz0=";
+              };
+            in
+            "''${repo}/quickshell-lockscreen";
+        }
+      '';
     };
     generated.output = mkOption {
       type = types.str;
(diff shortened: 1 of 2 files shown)


wrapperModules/q/quickshell/module.nix
```diff
@@ -16,6 +16,20 @@ let
...
   makeForce = lib.mkOverride 0;
+
+  componentModule = {
+    options = {
+      data = mkOption {
+        type = types.either wlib.types.linkable types.lines;
+        description = "The component's inlined text or path";
+      };
+      module = mkOption {
+        type = types.nullOr types.str;
```
BirdeeHub:
```suggestion
        type = types.bool;
```

is module (just means dont impose auto capitalization or .qml)

BirdeeHub:
There are a lot of implications and possible ways to handle this, so give me a moment. I did not know about this either.

I wonder if we can do something recursive with just a bool for "capitalize me" or not? I can look into it too I have been interested in quickshell recently

I do prefer this over the other PR's approach that disables all the stuff

but I wonder what I can cook up if I think about it a bit. If I am drawing a blank, this design seems to address the issue and I am OK with going with it?

Except it is still capitalizing and suffixing `.qml` in the module case I think so that will need to be fixed

wrapperModules/q/quickshell/module.nix
```diff
@@ -62,15 +76,17 @@ in
...
         value = {
-          content = mkIf (!linkable) val;
-          builder = mkIf linkable ''ln -s ${val} "$2"'';
+          content = mkIf (!linkable) val.data;
+          builder = mkIf linkable ''ln -s ${val.data} "$2"'';
           output = makeForce config.generated.output;
-          relPath = makeForce "${config.binName}-config/${capitalizedName}.qml";
+          relPath = makeForce "${config.binName}-config/${
+            if val.module != null then "${lib.replaceString "." "/" val.module}/" else ""
+          }${capitalizedName}.qml";
```
BirdeeHub: It is still doing this capitalize .qml thing in the module case.

BirdeeHub:
So, I'm somewhat drawing a blank, so, once we fix the autocapitalization thing, I think this is a fine approach, but I am thinking people who already have a quickshell config might not want to worry about all this.

To be fair, they don't have to use this module, they are free to provide a directory to quickshell themselves

```nix
{ config, lib, wlib, pkgs, ... }: {
  package = pkgs.quickshell;
  flags."--path" = ./mypath;
}
```

But I anticipate that a good number of people won't really realize how simple that is and will just complain about the lack of an option to just use a dir.

Rather than passing config.generated.placeholder to --path directly, what we could do, is make a configDir option, and make config.generated.placeholder its default value, and pass config.configDir to --path instead. This way we dont have a bunch of conditional logic, the dir will still be generated so people can still include it if they wish from the dir the pass, and such, but people can easily just pass their own dir.

ormoyo:
I think your recursive idea fits better. With it, external quickshell components can be imported like: `components.external.foo = pkgs.fetchFromGithub { <some repo> }`. 

It's also more similar to the `settings.foo.bar = "zoo"` syntax used in most modules here.

My `components.<name>.module` solution doesn't solve components with the same name (For example, a night clock from `import qs.night` vs light mode clock from `import qs.light`). 

Even if you deleted that comment, I think it has a lot of merit for the recursive type.

ormoyo:
It the meantime, I've disabled auto capitalization for linked components (it's using `builtins.baseNameOf` now) with the added option `components.<name>.name` to allow the user to change when needed.

From what I saw, quickshell only imports components with a capital letter, so I kept that for inlined components.

BirdeeHub:
I went ahead and merged the other PR in the meantime. We should have an option like that, and they fixed up their implementation to be much less impactful on our logic for the option.

The problem I am having, is the recursive type I put there, I deleted because it was awkward to use.

I am thinking of a few different designs instead, which are less ambiguous as to what is a new layer of nesting and what is a value.

ormoyo:
For me, what comes to mind is using `<module>.freeformType`.

I know it's still quite similar to your recursive type but I just don't see how is it worse then the generic `settings` option for most modules.

I made a commit for it as showcase

BirdeeHub:
Hey, sorry, Ive been working on systemd options in my spare time this last week, havent gotten to this yet.

I am about 80% of the way done with them, if you were curious, should be up in a PR in the next few days

BirdeeHub:
Ok. So, I started using quickshell this last week and am now finally actually equipped to handle this, sorry about that, I will _actually_ get to this soon now.

In the meantime, [heres my very recently made config using how this module currently works](https://github.com/BirdeeHub/birdeeSystems/tree/523e463e23a9567ae316c0bb1e20a99021314fe5/common/features/quickshell) plus my own modifications to it.

I tried to add a hot reload service thing. It didn't work though I will need to try harder than that :laughing:

wrapperModules/q/quickshell/module.nix
```diff
@@ -16,6 +17,16 @@ let
...
+
+  componentType = types.submodule {
+    freeformType = types.lazyAttrsOf (
+      types.oneOf [
+        wlib.types.linkable
+        types.lines
+        (componentType // { description = "nested components"; })
+      ]
+    );
+  };
```
BirdeeHub:
```suggestion
  componentType = types.lazyAttrsOf (
    types.oneOf [
      wlib.types.linkable
      types.lines
      (componentType // { description = "nested components"; })
    ]
  );
```

^ is this the same thing?

wrapperModules/q/quickshell/module.nix
```diff
@@ -64,24 +96,34 @@ in
...
+      getComponents' =
+        prefix: val:
+        if builtins.isAttrs val then
+          lib.foldlAttrs (
+            acc: name: v:
+            let
+              value = if isStringLike v then builtins.toString v else v;
+
+              # Auto capitalization for qml files or inlined text
+              attrIsFile = builtins.isString v || (isStringLike v && lib.hasSuffix ".qml" value);
```
BirdeeHub, line 107: I like this design a lot better but something feels weird in these conditions here somehow, unsure if that feeling is correct or not, looking into it now.

What happens if I pass it a path to a dir? But we also can't check if it is one directly either, because that would be IFD and also it might be an out of store link. So, kinda tricky.

wrapperModules/q/quickshell/module.nix
```diff
@@ -85,17 +88,34 @@ in
...
-      };
-    }) config.components
+    let
+      getComponents' =
+        prefix: val:
+        if builtins.isAttrs val then
+          lib.foldlAttrs (
+            acc: name: v:
+            let
+              value = if isStringLike v then builtins.toString v else v;
```
ormoyo, line 104: In this line, it uses ``toString`` in case the value is a package derivation. in which case we don't want to recurse anymore

wrapperModules/q/quickshell/module.nix
ormoyo: I do agree, recursion can be mind boggling. It took me quiet a lot of tries to get this right

ormoyo: Well, in case of a dir path, the script just checks if it's a "string like" and symlinks to its string representation. Even if it's IFD it will be counted as a "string like"

wrapperModules/q/quickshell/module.nix
```diff
@@ -64,24 +96,34 @@ in
...
+
+              # Auto capitalization for qml files or inlined text
+              attrIsFile = builtins.isString v || (isStringLike v && lib.hasSuffix ".qml" value);
+
+              firstChar = builtins.substring 0 1 name;
+              restChars = builtins.substring 1 (-1) name;
+              finalName = if !attrIsFile then name else (lib.toUpper firstChar) + restChars + ".qml";
```
BirdeeHub, line 111: also attrIsFile can be true when it ends in `.qml` which makes this double `.qml`

ormoyo: Actually, the symlink created does not inherit the name of the target, only the attribute name set in the module option. So it won't have double ``.qml`` (``name != value``)

wrapperModules/q/quickshell/module.nix
ormoyo: Yep, that's right

ormoyo:
I forgot to submit my comments until now ;(

PR #558, opened 2026-06-05
fix(wrapperModules.git): allow duplicate settings keys

Noah765:
For example, multiple instances of the credential.helper option are allowed to define multiple credential helpers.


BirdeeHub:
Wow you would really expect that the gitIni format would accurately generate the gitIni format but I guess not?

PR #559, opened 2026-06-07
docs(wrapperModules.nushell): add notes about the differences to normal config file placement

Noah765:
See https://www.nushell.sh/book/configuration.html#flag-behavior.

<details>
<summary>Outdated summary</summary>
The problem with setting the `--config` flag is that the configuration is loaded even in non-interactive environments. See https://www.nushell.sh/book/configuration.html#scenarios.

Note that this is a breaking change since it changes `$nu.default-config-dir` and affects the location of the history file, which has to be set to a non-Nix store location using `$env.config.history.path`. If you feel like this trade-off isn't worthwhile or have any suggestions, please suggest them or close this PR.
</details>

### DIFF
--- a/wrapperModules/n/nushell/module.nix
+++ b/wrapperModules/n/nushell/module.nix
@@ -19,6 +19,11 @@
         Provide either `.content` to inline the file contents or `.path` to reference an existing file.
         This file is passed to Nushell using `--env-config`, and is typically used to define
         environment variables or startup commands that apply to all shells.
+
+        Note that Nushell also reads this file in non-interactive environments unless
+        `--no-config-file` is specified. This is a subtle yet significant difference from
+        placing the config file at $XDG_CONFIG_HOME/nushell/env.nu. For more details, see
+        https://www.nushell.sh/book/configuration.html#flag-behavior.
       '';
     };
     "config.nu" = lib.mkOption {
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Hey! Sorry, took me a bit to get to this.

I recognize that the current nushell module is not great and will need to be improved.

However I am hesitant to export XDG_CONFIG_HOME for a whole shell. For many programs this is not so important, but this is a shell, and it will thus affect all the programs you launch using the shell. Is there a way we can use a different variable, or reset it before any of the user provided config and runtime code gets called?

Noah765:
> Hey! Sorry, took me a bit to get to this.

Don't worry — you're doing a great job here, and it's just a hobby project after all!

> However I am hesitant to export XDG_CONFIG_HOME for a whole shell. For many programs this is not so important, but this is a shell, and it will thus affect all the programs you launch using the shell. Is there a way we can use a different variable, or reset it before any of the user provided config and runtime code gets called?

I don't think there is another variable that can be used to configure the config path, and I cannot think of a way to reset XDG_CONFIG_HOME without using more flags which would change the behavior of Nushell in more unexpected ways. Furthermore, `$nu.default-config-dir` is set to $XDG_CONFIG_HOME/nushell at startup and cannot be changed.
See https://github.com/nushell/nushell/blob/main/crates/nu-path/src/helpers.rs#L22 and https://github.com/nushell/nushell/blob/main/src/config_files.rs#L58.

I believe that using the `--config` and `--env-config` is the least destructive option, so I updated this PR to add some clarifying documentation.

PR #561, opened 2026-06-08
feat(wrapperModules.zed): init

SibaldH:
## Summary

Adds a new `zed` wrapper module.

This provides a Home Manager-like declarative interface for Zed without depending on Home Manager. The wrapper can generate Zed config files and symlink them into the user’s `$XDG_CONFIG_HOME/zed` directory before launching Zed.

Supported options include:

* `userSettings` for `settings.json`
* `userKeymaps` for `keymap.json`
* `userTasks` for `tasks.json`
* `userDebug` for `debug.json`
* `extensions`, translated into `auto_install_extensions`
* `themes`, supporting generated JSON themes and existing theme file paths
* `linkConfig`, to control whether generated config is symlinked on launch
* `forceSymlinks`, to opt into replacing existing user config files or symlinks

## Design notes

Zed expects its configuration files in the user config directory, so this wrapper does not try to pass a custom config path to Zed. Instead, it generates the config files into the wrapper derivation using `constructFiles`, then links individual files into `$XDG_CONFIG_HOME/zed` when the wrapped binary starts.

The default behavior is intentionally conservative: existing real files or unrelated symlinks are not overwritten unless `forceSymlinks = true` is set.

## Testing

* [ ] `nix fmt`
* [ ] `nix build ./ci#checks.x86_64-linux.wrapperModule-zed -Lv`
* [ ] `nix flake check -Lv ./ci`

## Reviewer notes

I opened this as a draft because I would like feedback on the config-linking approach.

In particular, I would appreciate feedback on:

1. whether `linkConfig = true` is a good default;
2. whether the `forceSymlinks` safety behavior matches the expectations of this repository;
3. whether the option names should stay close to Home Manager’s Zed module names or use more generic wrapper-module naming.

### DIFF
--- a/wrapperModules/z/zed/module.nix
+++ b/wrapperModules/z/zed/module.nix
@@ -1,0 +1,357 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib)
+    mkDefault
+    mkIf
+    mkOption
+    optionalAttrs
+    optionalString
+    types
+    ;
+
+  jsonFormat = pkgs.formats.json { };
+  json5 = pkgs.python3Packages.toPythonApplication pkgs.python3Packages.json5;
+
+  isPathLike = value: builtins.isPath value || lib.isStorePath value;
+
+  generatedSettings =
+    config.userSettings
+    // optionalAttrs (config.extensions != [ ]) {
+      auto_install_extensions = lib.genAttrs config.extensions (_: true);
+    };
+
+  pathThemes = lib.filterAttrs (_name: value: isPathLike value) config.themes;
+
+  generatedThemes = lib.filterAttrs (_name: value: !(isPathLike value)) config.themes;
+
+  hasGeneratedConfig =
+    generatedSettings != { }
+    || config.userKeymaps != [ ]
+    || config.userTasks != [ ]
+    || config.userDebug != [ ]
+    || config.themes != { };
+
+  zedConfigDir = "${config.generatedConfig.placeholder}/zed";
+
+  mkJsonFile = relPath: value: {
+    inherit relPath;
+    content = builtins.toJSON value;
+  };
+
+  mkThemeFile = name: value: {
+    relPath = "${config.generatedConfig.relPath}/zed/themes/${name}.json";
+    content = if builtins.isString value then value else builtins.toJSON value;
+  };
+
+in
+{
+  imports = [ wlib.modules.default ];
+
+  options = {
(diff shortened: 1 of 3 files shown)


BirdeeHub:
Im a bit surprised zed modifies its settings file directly rather than using a data dir

This is looking OK in light of that, but you have isPathLike, I am wondering if it might be solved more elegantly with the [wlib.types.linkable](https://birdeehub.github.io/nix-wrapper-modules/lib/types.html#function-library-wlib.types.linkable) type?

I also might have another thing or 2 to comment on, will look further.

Sorry, I have been having quite the time the last few weeks I haven't had much time to review PRs

wrapperModules/z/zed/module.nix
```diff
@@ -0,0 +1,338 @@
...
+        ln -sfn ${lib.escapeShellArg value} ${lib.escapeShellArg "${generatedZedConfigDir}/themes/${name}.json"}
+      '') pathThemes;
+    };
+
+    runShell = mkIf (hasGeneratedConfig && config.linkConfig) [
+      {
+        name = "NIX_ZED_LINK_CONFIG";
+        "esc-fn" = wlib.escapeShellArgWithEnv;
+
+        data = ''
```
BirdeeHub, line 287: Im not super sure about this implementation here.

If we are just symlinking it to the correct location at runtime, shouldn't we be setting XDG_CONFIG_HOME ourselves rather than this? If we are symlinking, we do not gain the ability to edit the files, so we aren't getting anything extra out of this rather than setting XDG_CONFIG_HOME

Basically, what I think we should do here is a combo of what we do for [helix](https://github.com/BirdeeHub/nix-wrapper-modules/blob/6e7f66fa2cdf4d63162580b438f7fcf87c28a46f/wrapperModules/h/helix/module.nix#L84) and for [noctalia shell](https://github.com/BirdeeHub/nix-wrapper-modules/blob/6e7f66fa2cdf4d63162580b438f7fcf87c28a46f/wrapperModules/n/noctalia-shell/module.nix#L292-L299)

https://github.com/BirdeeHub/nix-wrapper-modules/blob/6e7f66fa2cdf4d63162580b438f7fcf87c28a46f/wrapperModules/n/noctalia-shell/module.nix#L347-L352

So, we should use XDG_CONFIG_HOME here and point directly to the store, and then have an option to use an out of store config, which copies it to the location instead like we have for noctalia, that way they can accept runtime edits from the gui if they want out of store, otherwise it is all properly bundled in the store and doesn't need runtime instructions.

PR #564, opened 2026-06-16
feat(modules.systemd): systemd helper module

BirdeeHub:
# Things To Be Done / Questions To Answer:

- [x] Should this be added to `wlib.modules.default` or not?
- [ ] You need to log out and log back in for the FIRST time the service is added in order for the service to be automatically enabled. After that, changes are reflected on rebuild. It would be nice to have it not need that.
- [x] More docs
- [x] If you generate the documentation, or look through the module.nix, you will notice there is a LOT of options. This is because, in the freeform modules, most of the options have been documented. Not all of them, but most of them. We might want to remove some, and leave the user to infer more from the systemd manual pages? It is a 1 to 1 translation to make this mapping easy. This is also 1 reason why we might not want to have it in `wlib.modules.default`, not everything needs a service, and there are other service runners besides systemd.

## about the second one:
 
You can still start it yourself right away. It is installed, just not enabled.

The action to enable it has been done. Systemd does not know about that yet.

I am not entirely sure why you cannot `systemctl daemon-reload` or `systemctl daemon-reexec`, but it seems like you have to log out or restart. It seems to check for enable status at a very particular time and then cache the results.

I tried mirroring most of the options to nixos options, and that kinda solved this issue, but not entirely

When you mirror the options, nixos uses the overrides mechanism, so you can mirror just the wantedBy requiredBy and upheldBy options. It mostly solves the issue, but sometimes the service is enabled but doesn't seem to run until you restart anyway? But systemd is reporting everything should be running

 home manager doesn't merge with the ones from the package, so we would need to mirror ALL of them for home manager and ignore the file from the package and this would not work with placeholders...

I think there is something that can be done with the generated install module that can work around this, maybe with an activation script that can solve this, but that is what has been taking so long. For now, the options work, they are a lot like the home manager ones, but you need to log out or restart the first time you add the service for it to be enabled. After that, changes are reflected on rebuild, as the old file no longer exists on the systemd path and has been replaced with a new one with the same name, so it inherits the enabled status and reflects the new values.

Also, sorry I have been absent the last week, will be getting back to it!

### DIFF
--- a/modules/systemd/module.nix
+++ b/modules/systemd/module.nix
@@ -504,32 +504,32 @@ let
             freeformType = sectionType;
             options = {
               OnActiveSec = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
-                default = null;
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
                 description = "Timer relative to when this timer unit was activated.";
               };
               OnBootSec = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
-                default = null;
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
                 description = "Timer relative to boot time.";
               };
               OnStartupSec = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
-                default = null;
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
                 description = "Timer relative to when the service manager started.";
               };
               OnUnitActiveSec = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
-                default = null;
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
                 description = "Timer relative to when the triggered unit was last activated.";
               };
               OnUnitInactiveSec = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
-                default = null;
+                type = lib.types.listOf lib.types.str;
+                default = [ ];
                 description = "Timer relative to when the triggered unit was last deactivated.";
               };
               OnCalendar = lib.mkOption {
-                type = lib.types.nullOr lib.types.str;
+                type = lib.types.listOf lib.types.str;
                 default = [ ];
                 description = "Realtime (wallclock) calendar event expression.";
               };
@@ -656,25 +656,25 @@ let
                 default = true;
                 description = "Enable ${id} unit.";
               };
-              install = lib.mkOption {
-                type = lib.types.listOf (
-                  lib.types.enum [
-                    "nixos"
-                    "homeManager"
-                    "hjem"
-                  ]
-                );
-                default = [
-                  "nixos"
-                  "homeManager"
-                  "hjem"
-                ];
-                description = "Create installation logic for other nix module systems for ${id} unit. If a list, only for the named module system classes.";
+              doInstall = lib.mkOption {
+                type = lib.types.bool;
+                default = true;
+                description = "Create .wants .requires and .upholds links for ${id} unit.";
               };
               overwrite = lib.mkOption {
                 type = lib.types.bool;
                 default = false;
                 description = "Overwrite existing unit file instead of appending generated content to it if present.";
+              };
+              prefixedContent = lib.mkOption {
+                type = lib.types.lines;
+                default = "";
+                description = "Content to prepend to the beginning of the generated ${id} unit file.";
+              };
+              suffixedContent = lib.mkOption {
+                type = lib.types.lines;
+                default = "";
+                description = "Content to append to the end of the generated ${id} unit file.";
               };
             };
         };
@@ -684,21 +684,6 @@ let
           type = lib.types.bool;
           default = true;
           description = "Enable generation of systemd ${name} units.";
-        };
-        install = lib.mkOption {
-          type = lib.types.listOf (
-            lib.types.enum [
-              "nixos"
-              "homeManager"
-              "hjem"
-            ]
-          );
-          default = [
-            "nixos"
-            "homeManager"
-            "hjem"
-          ];
-          description = "Create installation logic for other nix module systems for ${name} units. If a list, only for the named module system classes.";
         };
         service = lib.mkOption {
           type = lib.types.attrsOf (
@@ -905,8 +890,22 @@ in
     ./config.nix
   ];
   config.meta.maintainers = [ wlib.maintainers.birdee ];
+  options.install.systemd = lib.mkOption {
+    type = lib.types.listOf (
+      lib.types.enum [
+        "nixos"
+        "homeManager"
+        "hjem"
+      ]
+    );
+    default = [
+      "nixos"
+      "homeManager"
+      "hjem"
+    ];
+    description = "Add the service files in the derivation to the specified module systems via the install module";
+  };
   # systemd.<name>.{user, system}.{target, path, timer, service, socket, scope, device, mount, automount, swap, path, slice}.{ relevant filemod + enable, install fields }
-  # enable is a bool option, and install is a list of strings for which classnames to reflect the values to that module system or not, or a bool true for all or false for none
   options.systemd = lib.mkOption {
     type = lib.types.submodule {
       options = {
--- a/modules/systemd/module.nix
+++ b/modules/systemd/module.nix
@@ -1,0 +1,924 @@
+{
+  config,
+  lib,
+  wlib,
+  ...
+}:
+let
+  atom = lib.types.nullOr (
+    lib.types.oneOf [
+      lib.types.bool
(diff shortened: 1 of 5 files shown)


modules/systemd/module.nix
```diff
@@ -0,0 +1,923 @@
...
+                default = null;
+                description = "Maximum time the scope may be active (e.g. 1h, 30m).";
+              };
+            };
+          };
+        };
+      };
+    };
+  };
+  extensionsSubmodule = { name, ... }: {
```
BirdeeHub, line 638: The `systemd.{user,service}` submodule, containing options for each unit file extension type

The interesting thing to note is `extraMod` which adds some extra per-service options for enabling or disabling various actions we perform in translating the options to unit files.

The code above this comment is just the systemd docs written out as nix options basically. Below is more interesting to figure out how it works. These options are processed in ./config.nix

BirdeeHub:
@zenoli @iynaix @rachitvrma @FluxZA is anyone available and interested in reviewing or giving advice? This is still a work in progress but it should be most of the way there already.

zenoli:
Hi @BirdeeHub 
Cool stuff!
I have a lot going on right now. I'm glad to skim through it on the weekend if you can wait that long, I am by no means a systemd expert though, so only expect a superficial review from me :-)

BirdeeHub:
> Hi @BirdeeHub Cool stuff! I have a lot going on right now. I'm glad to skim through it on the weekend if you can wait that long, I am by no means a systemd expert though, so only expect a superficial review from me :-)

It will definitely be waiting that long.

Try to ignore the noise from the transposed systemd options when you do review it.

FluxZA:
Hi All

I also will only have a chance to look over this on the weekend. I have some experience with systemd and how nixos handles it so hopefully I can help. Also happy to help out with documentation.

rachitvrma:
Hey everyone, I have been quite busy with some college work. I'll be happy to look into this, though.

zenoli:
My apologies, I did not get around to look at it, unfortunately I don't think I will have time in the next days :-(

FluxZA:
Hi, sorry this is a little late, my weekend was busier than I expected and I also ended up getting more absorbed in the manga I am reading than I planned 😂 

I have gone through everything once, and I have some initial comments and questions. I haven't looked into the restart/logout issue yet, but I am still happy to if you want a second opinion.

***

Regarding inclusion in the default module, I agree that this is not necessary and would probably lead to the docs being cluttered. I was even thinking you could potentially split the current module to separate the systemd options and the extra mods options, so that if you are reading the docs you know if you are looking at something systems related or something related to the nix-wrap implementation. This would also create more separation for if you wanted to work with other service handlers in future.

This leads to the question of the docs for the systemd options. In my opinion we should have as many of them documented as possible, but keep the descriptions very basic with links to the systemd documentation. That way the docs aren't too bloated and if anything changes from systemd's side we don't have to maintain our docs to match the new behaviour.

***

Regarding your implementation, I am unsure of how the overwrite or append would work. Is this catering for a case where the host system already has a service and the wrapper needs it to behave differently?

***

Regarding docs, do you want help adding more docs or just reviewing the current docs?

BirdeeHub:
> Regarding your implementation, I am unsure of how the overwrite or append would work. Is this catering for a case where the host system already has a service and the wrapper needs it to behave differently?

Close, it is catering for the case where the derivation already provides a service file with that name, for if you wish to modify it rather than overwriting it.

If overwrite is true, it will just replace it.

---

That is 1 vote for "not included in `wlib.modules.default`" so far. (There is not actually any voting system but that's 1 vote regardless.)

---

> Regarding docs, do you want help adding more docs or just reviewing the current docs?

I was going to add a `meta.description` and review the freeform options that I have there already.

I am more interested in people reviewing the top level options that are not directly translated, and the implementation, as opposed to the list of systemd options for each section and their docs, as those are just transcribed from `man systemd.*` and I can review that just fine with a bit of time (already taken 2 passes at it but I probably still have 1 or 2 with the wrong type I will find when I look again)

FluxZA:
Ahh okay, that makes sense. Cool I will go back and look at those descriptions in particular and leave comments if I have any.

modules/systemd/module.nix
```diff
@@ -0,0 +1,923 @@
...
+    options =
+      let
+        extraMod = { _prefix, ... }: {
+          options =
+            let
```
FluxZA, line 643: The id generation here might benefit from some internal documentation. It took me a long time to figure out what it was doing (could just be a shortcoming in my nix knowledge though🤦🏼)

My understanding is that it is used to dynamically update the descriptions based of of where the extraMods function ends up being imported, using the last 3 levels of the option hierarchy if they exist. One thing I noticed when actually generating the docs, the final element (used in `${get 2})` seems to be a space for some of the options, I am not sure if that is intended.

FluxZA: Let me know if this is not the type of doc suggestions you want. I had an hour to start looking this eve and this is the first thing I came across which I didn't understand. Will continue looking at more the rest of the week and I will try adjust feedback based on the level you are looking for.

FluxZA:
Reading through things further I realised my grasp on systemd units was not as solid as I thought. It took me a while to find time to get better acquainted, sorry. I have now been able to go through things again with a bit of a better understanding and I have a few comments:

- The `systemd` option could have a description pointing to the  `install.systemd` option to make sure people don't miss the fact that they can control how the units are installed.
- The `enable` option under the extra mod section could include the fact that it is not related to `systemctl enable` to help avoid confusion. On that note, the `doInstall` option could potentially reference that it is related to `systemctl enable`.
- I am not sure if it would completely replace the need for the `overwrite`, `prefixedContent` and `suffixedContent` options, but an option that populates the `override.conf` file for a unit would help in cases where you only want to change some sections of a pkg supplied unit.

Other than that, things feel pretty good. We can definitely add to the docs in places as you mentioned, but I like the general flow of things. For adding to docs, would you want to do it as part of this PR or merge the module and then work on docs after?

BirdeeHub:
> I am not sure if it would completely replace the need for the overwrite, prefixedContent and suffixedContent options, but an option that populates the override.conf file for a unit would help in cases where you only want to change some sections of a pkg supplied unit.

I should indeed look into using overrides maybe. I wonder if that can be used to improve enablement.

BirdeeHub:
> but an option that populates the override.conf file for a unit would help in cases where you only want to change some sections of a pkg supplied unit.

So, actually, it already handles that. systemd unit override files have the same effect as appending the text to the end of the service file. If there is already a service file in that location, it should extend the file that was there, not overwrite it. That is what overwrite is for, is to make it NOT act like override. If you want it to overwrite, you have to mark that it should do so.

populating override.conf might be helpful for other reasons though so I will still be looking into it.

FluxZA:
That makes sense. What I was suggesting was that if you did have access to the override mechanism, then you could just use it to override everything in the original file to act like it had been overwritten. I do agree that it probably isn't the best way to do things, that is just what I was thinking at the time.

Looking more now into the difference between appending existing keys to a file vs putting them in an override file, there doesn't seem to be any difference. The main use case for override files that I can gather is for when you don't want to, or can't, edit the main file. In the case of a wrapper this would probably never occur.

BirdeeHub:
> The main use case for override files that I can gather is for when you don't want to, or can't, edit the main file. In the case of a wrapper this would probably never occur.

We already require that the derivation we are wrapping be accessible, and we make the wrapper, so we can put the combined content into the final wrapper at build time, which is how it currently works.

FluxZA:
Cool, sorry for the red herring of a suggestion 😂

BirdeeHub:
Yeah I would be looking into using overrides in order to get it to auto enable easier without restarting. It might be that I can piggyback on an existing service file easier than I can provide the base one. Unsure.

BirdeeHub:
I investigated using overrides, but I didn't find a way to do the automatic enablement better without just ignoring the generated file and translating it straight for the target module system's options, which isn't what these should be for.

So, I am leaning towards merging this now as it is.

I will go through and look at the types again, to do 1 last pass in case I missed any that do not match the docs.

If any are discovered to be wrong in the future we can of course update them.

But, as is, I have uses for this module, 2 modules in this repo also have uses for this module, and while initial automatic enablement requires a restart, that only matters if the service had wantedBy requiredBy or upheldBy definitions to begin with, and on future rebuilds it does not require a restart.

BirdeeHub:
Alright, I'm merging it Im tired of this sitting in the PR's. I'm using it in my config, I like it, it does something useful, its going in IDK.

FluxZA:
Sorry I missed your last comment, but great work merging this. I see from the Hypridle PR that it's already paying dividends.

PR #567, opened 2026-06-28
feat(wrapperModules.hyprland): init

jonas-elhs:
This PR adds a wrapper module for hyprland, though only for the lua configuration. In addition it automatically loads hyprland plugins.

Additional features that could be added are old hyprlang configuration based on an attrset and support for split configuration files.

### DIFF
--- a/maintainers/default.nix
+++ b/maintainers/default.nix
@@ -1,5 +1,4 @@
-{ lib }:
-{
+{ lib }: {
   inherit (lib.maintainers) birdee ameer;
   patwid = {
     name = "Patrick Widmer";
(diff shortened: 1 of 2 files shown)


iynaix:
Perhaps you could add something like:

```nix
config.drv.installPhase = ''
    runHook preInstall
    export XDG_RUNTIME_DIR=$(mktemp -d)
    ${lib.getExe config.package} --verify-config -c "${config.constructFiles.generatedConfig.path}"
    runHook postInstall
'';
```

to validate the hyprland config, similar to the validation in the niri and mango modules?

jonas-elhs:
Thanks for the tip!

PR #570, opened 2026-06-29
feat: added a wrapper for `neovide`

Nuclear-Squid:
As the title says, this PR adds a wrapper module for `neovide`, a fancy GUI for Neovim. It can be configured in two places (for two different usages) :
- there’s a toml config file for things that are mostly static (like the font you use, or where to log backtraces): https://neovide.dev/config-file.html
- there’s a bit of lua config, which is specified (with a guard) in your neovim config: https://neovide.dev/configuration.html

This PR only handles the first one (the config file) as I go over the assumption that you’ll have your lua config with the rest of your neovim config. (I’m also not really sure if that would be possible to implement, since wrapped packages seem to be independent)

First time opening a PR, hopefully I didn’t mess it up too much :sweat_smile:

### DIFF
--- a/wrapperModules/n/neovide/module.nix
+++ b/wrapperModules/n/neovide/module.nix
@@ -1,0 +1,33 @@
+{
+  config,
+  lib,
+  wlib,
+  pkgs,
+  ...
+}:
+{
+  imports = [ wlib.modules.default ];
+  options = {
+    settings = lib.mkOption {
+      type = wlib.types.structuredValueWith {
+        nullable = false;
+        typeName = "TOML";
+      };
+      default = {};
+      description = ''
+        Configuration passed to neovide using the `NEOVIDE_CONFIG` environment variable.
+        See <https://neovide.dev/config-file.html> for the config options.
+      '';
+    };
+    config = {
+      package = pkgs.neovide;
+      env.NEOVIDE_CONFIG = "${config.binDir}/config.toml";
+      constructFiles.generatedConfig = {
+        content = builtins.toJSON config.settings;
+        relPath = "${config.binDir}/config.toml";
+        builder = ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
+      };
+      meta.maintainers = [ wlib.maintainers.nuclear-squid ];
+    };
+  };
+}
--- a/maintainers/default.nix
+++ b/maintainers/default.nix
@@ -116,4 +116,9 @@
     github = "aliaslion";
     githubId = "122117018";
   };
+  nuclear-squid = {
+    name = "Nuclear-Squid";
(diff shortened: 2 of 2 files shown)


wrapperModules/n/neovide/module.nix
```diff
@@ -0,0 +1,33 @@
...
+      package = pkgs.neovide;
+      env.NEOVIDE_CONFIG = "${config.binDir}/config.toml";
+      constructFiles.generatedConfig = {
+        content = builtins.toJSON config.settings;
+        relPath = "${config.binDir}/config.toml";
+        builder = ''${pkgs.remarshal}/bin/json2toml "$1" "$2"'';
+      };
+      meta.maintainers = [ wlib.maintainers.nuclear-squid ];
+    };
+  };
```
MacaylaMarvelous81, line 32: Looks like you've mistakenly placed the module's `config` inside of its `options`, so your wrapper module isn't actually setting any options or configuring anything. Therefore, this wrapper module will not function.

MacaylaMarvelous81 (changes requested):
This wrapper module fails to build a wrapper.

```sh
[jomarm@dell-inspiron7773:~/nix-wrapper-modules]$ nix-build --expr '(import ./. {}).wrappers.neovide.wrap { pkgs = import <nixpkgs> {}; }'
error:
       … while calling a functor (an attribute set with a '__functor' attribute)
         at «string»:1:1:
            1| (import ./. {}).wrappers.neovide.wrap { pkgs = import <nixpkgs> {}; }
             | ^

       … while evaluating the attribute 'mergedValue'
         at /nix/store/kz804kyjidvk0bqkpax50n3gl0d42gkd-z40ippcpb9xbmjyads2br13snirhdm0i-nixpkgs-src/lib/modules.nix:1244:5:
         1243|     # Type-check the remaining definitions, and merge them. Or throw if no definitions.
         1244|     mergedValue =
             |     ^
         1245|       if isDefined then

       … while evaluating definitions from `/nix/store/y41wzpvpvinf4gi81ksmp484sm08p67z-nix-wrapper-modules/lib/core.nix':

       … while evaluating the option `wrapper':

       … while evaluating definitions from `/nix/store/y41wzpvpvinf4gi81ksmp484sm08p67z-nix-wrapper-modules/lib/core.nix':

       … while evaluating the option `buildCommand.makeWrapper.data':

       … while evaluating definitions from `/nix/store/y41wzpvpvinf4gi81ksmp484sm08p67z-nix-wrapper-modules/modules/makeWrapper/module.nix':

       … while evaluating the option `outputName':

       … while evaluating the option `package':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: The option `package' was accessed but has no value defined. Try setting the option.
```

PR #572, opened 2026-07-03
fix(wrapperModules.emacs): set `package-user-dir` variable

HeitorAugustoLN:
Currently it gets set to a path in the nix store, due to --init-directory, making commands like list-packages, or any imperative package management command work


BirdeeHub:
So, this seems fine, if anyone were to already be setting that themselves in their config, then theirs would win anyway.

Question. If you do this, `config.emacsPackages` still works, correct? If so, then yes lets do this.

HeitorAugustoLN:
Yes, it worked fine when I tested the changes

BirdeeHub:
Thank you! Sorry for taking so long to get to this, thank you for the quick response!

PR #573, opened 2026-07-03
feat(test-lib): add notFileContains helper

AlexLov:
This adds a complementary helper (to already existed `fileContains`) -- `fileNotContains` for simple comparison.

Generally it just negates result of `grep` expression from `fileContains` helper.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,6 +184,7 @@ of the wrapper (if any) and ensure that the tests are only run on the required p
 let
   inherit (tlib)
     fileContains
+    fileNotContains
     isDirectory
     isFile
     notIsFile
@@ -202,7 +203,7 @@ test { wrapper = "direnv"; } { # <-- Specify the name of the wrapper here (*)
     in
     [
       "[[ -d ${wrapper} ]]" # <-- a simple condition to be asserted
-      {                     
+      {
         cond = "[[ -d ${wrapper} ]]";
         msg = "No directory found for wrapper."; # <-- you can also specify a custom error message
       }
@@ -214,7 +215,7 @@ test { wrapper = "direnv"; } { # <-- Specify the name of the wrapper here (*)
       wrapper = self.wrappers.direnv.wrap {
         inherit pkgs;
       };
-    in 
+    in
     '' # <-- no need to provide a list if there is only one assertion
       "${wrapper}/bin/direnv" --version |
       grep -q "${wrapper.version}"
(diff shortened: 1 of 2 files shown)


MacaylaMarvelous81 (approved):
<details>
<summary>Imperative test</summary>

```
[jomarm@dell-inspiron7773:~/nix-wrapper-modules]$ nix-instantiate --eval --expr '(((import <nixpkgs> {}).callPackage ./ci/test-lib.nix { self = import ./. {}; }).fileNotContains "maintainers/default.nix" "alexlov").cond'
"! grep -Eq -- 'alexlov' \"maintainers/default.nix\""

[jomarm@dell-inspiron7773:~/nix-wrapper-modules]$ ! grep -Eq -- 'alexlov' "maintainers/default.nix"; echo $?
1
```
</details>

CONTRIBUTING.md
```diff
@@ -184,6 +184,7 @@ of the wrapper (if any) and ensure that the tests are only run on the required p
 let
   inherit (tlib)
     fileContains
+    fileNotContains
```
zenoli: What about `notFileContains`?
Would be more consistent with `notIsFile`

```suggestion
    notFileContains
```

AlexLov: `notFileContains` sounds and looks very strange and weird to me.
I’m not native in English though, and don’t have strong opinion on this.

MacaylaMarvelous81: In my opinion `notFileContains` makes sense because it reads like "the negation of `fileContains`" and makes its name consistent with the other negative helpers. Also, the names of `notIsFile` and `notIsDirectory` do not really follow English grammar; as a phrase you might say that something "is not a file" or "is not a directory".

zenoli: > notFileContains sounds and looks very strange and weird to me.

I don't disagree but I think having a consistent naming is more important.

If we want a very "clean" solution we could introduce a `not` operator in the form of a nix function `not :: Assertion -> Assertion` which takes an assertion and adds the `"!"` in front of the condition.

Then it could be used as `not (fileContains my-file)` and remove all the existing `not*` assertions functions.
It would be a breaking change though.

I would also be fine with renaming the other functions to match your convention (`isNotDirectory, isNotFile` etc) but this would be breaking too and hence conforming to the existing convention seems more reasonable to me.

AlexLov: Ok, renamed it to be more consistent.
Thanks for the input!

zenoli (approved):
Thanks for the update. LGTM!

@BirdeeHub I think this is ready to be merged.

PR #575, opened 2026-07-03
fix(wrapperModules.jujutsu): update itself in compltetions files

AlexLov:
Jujutsu's can autocomplete aliases declared in its config, but jujutsu from those autocompletion files should know about the configured aliases, so it's better to override it by default with configured/wrapped binary.

This resolves frustration I had when I couldn't get autocompletion works as expected. 
Also, I tested it only on zsh, but added all completion files assuming all shells works the same way regarding this issue.

P.S.: this same issue might affect git as well. But I haven't gotten around configuring wrapped version of git yet to confirm it.


BirdeeHub:
Yeah this seems like a good idea to me. Thank you!

PR #577, opened 2026-07-09
feat(wrapperModules.mangowc): reload with new config after build

WeraPea:
Adds mango-reload.service which reloads mango with new config after building, based on #496.

Differs from niri's implementation by the fact that it uses a standalone service and requires users to start that service themself, as mango doesn't provide a systemd service.

Currently requires git version of mango as 0.15 with the required `load_config_file` dispatch has not released yet.

### DIFF
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -193,6 +193,25 @@
         default = "";
         internal = true;
       };
+
+      hotReload.enable = lib.mkOption {
+        type = lib.types.bool;
+        default = true;
+        description = ''
+          When `true`, the wrapper will generate mango-reload service which will reload mango with the new config on rebuild.
+
+          Note: You have to start the mango-reload.service yourself.
+          For Example:
+          ```
+            systemd.packages = [ config.wrappers.mango.wrapper ];
+            wrappers.mango.settings.exec-once = [
+              "systemctl --user start mango-reload.service"
+            ];
+          ```
+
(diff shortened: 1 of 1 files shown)


wyndon:
I'm not sure how to use this. I've patched my input to include this PR, but I don't have a mango-reload.service that I can start.

WeraPea:
> I'm not sure how to use this. I've patched my input to include this PR, but I don't have a mango-reload.service that I can start.

For the service to be picked up by systemd you need to add mango to systemd packages with:
```
systemd.packages = [ config.wrappers.mango.wrapper ];
```

BirdeeHub:
This looks good! One thing to note, we have a PR with systemd unit file options that will be landing in the near future, and this can be done that way, and it will do basically the same thing. It can even automatically enable it if you provide wanted by directives! (although on first install, it requires a restart to pick up that it has been enabled)

We can rewrite this to use the systemd module when it is released.

BirdeeHub:
Is there a target or something you could use to run this service that occurs after mango starts? Then you wouldn't have to start it yourself.

WeraPea:
Only through mango flake's home-manager module, and even in this case it is started with the generated config file and not through mango itself.

BirdeeHub:
Should we add to the generated config something that calls a systemd target if they have this set to true, and set this to be wantedBy that target? Then we can have the config call it ourselves in our config too?

This is fine for now, but when we have systemd options that can enable themselves it is something to think about.

PR #578, opened 2026-07-10
chore: update to new pkgs.mango

ssoberun:
title, pkgs.mangowc no longer works on unstable(?)

### DIFF
--- a/wrapperModules/m/mangowc/module.nix
+++ b/wrapperModules/m/mangowc/module.nix
@@ -250,7 +250,7 @@
     };
 
     flags."-c" = config.configFile.path;
-    package = lib.mkDefault pkgs.mangowc;
+    package = lib.mkDefault pkgs.mango;
     passthru.providedSessions = config.package.passthru.providedSessions;
 
     meta.platforms = lib.platforms.linux;


ssoberun:
it seems it has changed in unstable but not in 26.05, hence the check on ubuntu-latest failing

wrapperModules/m/mangowc/module.nix
```diff
@@ -250,7 +250,7 @@
     };
 
     flags."-c" = config.configFile.path;
-    package = lib.mkDefault pkgs.mangowc;
+    package = lib.mkDefault pkgs.mango;
```
AlexLov, line 253: I would suggest adding version check, something like this:
```
    package = lib.mkDefault (if lib.versionAtLeast lib.version "26.11.20260629" then pkgs.mango else pkgs.mangowc);
```
(please apply formatting and/or move the condition to `let..in` block)

Note, I took the date from [PR](https://github.com/NixOS/nixpkgs/pull/534515)'s merge date, so it might not be accurate

BirdeeHub:
I missed this and fixed it already. Thank you for the help though!

PR #579, opened 2026-07-11
feat(wrapperModules.ripgrep): init

Nadim147c:
Added ripgrep module. I've added two options:

- `settings`: attrset of options' names and values.
- `ignores`: lines of glob for ripgrep ignore.


wrapperModules/r/ripgrep/check.nix
zenoli, line 1: In case you're interested, we have a dedicated test-lib to make writing tests more convenient:

https://birdeehub.github.io/nix-wrapper-modules/md/CONTRIBUTING.html#writing-tests

Nadim147c: I was copying/following the `yt-dlp` tests. I will rewrite the checks.

zenoli: Yeah, I think those were written before the tlib was introduced. Unlucky :-)

PR #581, opened 2026-07-13
feat(wlib.toKdl): add custom value to support literal regex props

iynaix:
Adds an `_raw` attribute to be able to use raw literals in kdl, e.g.

```nix
excludes = [  { app-id._raw = "^org\.telegram\.desktop$"; } ];
```

to produce

```kdl
exclude app-id=r#"^org\.telegram\.desktop$"#
```

as show in the [niri docs](https://niri-wm.github.io/niri/Configuration%3A-Window-Rules.html#title-and-app-id)

Fixes https://github.com/BirdeeHub/nix-wrapper-modules/discussions/566

### DIFF
--- a/lib/toKdl.nix
+++ b/lib/toKdl.nix
@@ -19,7 +19,8 @@ let
                 res = lib.fix v;
               in
               lib.optionalString (res ? type) "(${toString res.type})"
-              + lib.optionalString (res ? content) "${toJSON res.content}"
+              + lib.optionalString (res ? content && !res ? custom) "${toJSON res.content}"
+              + lib.optionalString (res ? custom) res.custom
             else if isAttrs v || isList v then
               toJSON (toJSON v)
             else
--- a/ci/checks/toKdl.nix
+++ b/ci/checks/toKdl.nix
@@ -61,6 +61,17 @@ let
       };
     }
     {
+      description = "function with props custom content";
+      expected = ''"h" "k"=r#"v"#'';
+      actual = toKdl {
+        h = _: {
+          props = {
+            k = _: { custom = ''r#"v"#''; };
+          };
+        };
+      };
+    }
+    {
       description = "function with content only";
       expected = "\"i\"  {\n  \"j\" 1\n}";
       actual = toKdl {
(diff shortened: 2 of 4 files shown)


BirdeeHub:
I feel like there is a way to do it already with the custom content thing, but having a dedicated way to do specifically these should be in the spec. Will need to pull this branch and look at the code locally for this one, toKdl ended up being a fairly complex thing

iynaix:
I didn't know custom content existed :sweat_smile: . I gave it a try with:

```nix
{
  excludes = [
    {
      app-id = _: { content = "r#\"^org\.telegram\.desktop$\"#""; };
    }
  ];
}
```

which looks like it should work; but unfortunately because of the `"${toJson res.content}"`, it escapes the quotes again and produces

```kdl
exclude app-id=r#\"^org\.telegram\.desktop$\"#
```

in the output file.

BirdeeHub:
yeah the custom one is only at the node level not the value level. The value level doesn't have that, I didn't realize it would be required, but it does accept a function form too, either to omit it, or to add a type annotation for stuff in props.

But, if we want to add one of these raw things and do it at the block level, we should probably add it to that there, like

```nix
{
  excludes = [
    {
      app-id = _: {
        raw = true;
        content = ''^org\.telegram\.desktop$'';
      };
    }
  ];
}
```

Because otherwise it is ambigous as to if this is a block which contains `_raw "somesstring"` as an item in it or a raw string unless it is in the props section explicitly

Unfortunately this is also more verbose. So maybe instead of `raw = true;` we could provide a `custom` field there instead or in addition, and then it could be the following instead, which I feel is an acceptable level of verbosity to avoid ambiguity?

```nix
{
  excludes = [
    {
      app-id = _: { custom = ''r#"^org\.telegram\.desktop$"#''; };
    }
  ];
}
```

iynaix:
Yes, I agree that `_: { custom = ''r#"^org\.telegram\.desktop$"#'';`  is cleaner, and more generic as well. I've updated the implementation to use that instead. The implementation also turns out to be simpler and feels less "hacky".

lib/toKdl.nix
```diff
@@ -20,6 +20,7 @@ let
               in
               lib.optionalString (res ? type) "(${toString res.type})"
               + lib.optionalString (res ? content) "${toJSON res.content}"
```
BirdeeHub:
```suggestion
              + lib.optionalString (res ? content && ! res ? custom) "${toJSON res.content}"
```

Should we make custom override content instead of producing invalid output when someone supplies both by mistake?

Otherwise yes this is how I would do it and it looks good to me.

iynaix: Agreed, made the change.

Issue #582, opened 2026-07-15
fish: abbreviations should be position "command" instead of "anywhere" by default


iynaix:
It seems rather odd that the default position for abbreviations would be `anywhere`, since I imagine the majority of abbreviations the people use are not meant to be expanded anywhere, or at least that's the case for me.

Home-manager has it defaulted to `command` https://github.com/nix-community/home-manager/blob/a45a7c451455a51ae740ec3bce4024b312809c29/modules/programs/fish.nix#L187

@ormoyo

ormoyo:
When I made the fish module, it was the first month I started using fish. 
I didn't know many of the fish conventions.

Thanks for the suggestion ;)

PR #583, opened 2026-07-16
fix(wrapperModules.fish): fix abbreviation argument order

iynaix:
Currently, doing

```nix
abbreviations = {
  "--help" = {
    expansion = "--help | bat --plain --language=help";
    position = "anywhere";
  };
};
```

generates

```fish
abbr --add --help  --position anywhere    "--help | bat --plain --language=help"
```
which results in `abbr --help` being called instead of the abbreviation being set.

This fix changes it to generate the following instead.

```fish
abbr --add  --position anywhere    -- --help "--help | bat --plain --language=help"
```

I've also wrapped the shellAliases and abbreviations to only be evaluated in interactive shells. Any output produced by fish breaks completion generation, since it gets prepended to the python completion generator: https://github.com/NixOS/nixpkgs/blame/6368bc923cec55a5f78960ade0cb4dd99580e087/nixos/modules/programs/fish.nix#L273

I've also set the abbreviation position to "command" as outlined in #582. If this is not desired, I can remove this change.

@ormoyo

### DIFF
--- a/wrapperModules/f/fish/module.nix
+++ b/wrapperModules/f/fish/module.nix
@@ -468,10 +468,9 @@ in
 
         mkAbbrStr =
           abbr:
-          (foldl' (
-            acc: elem: acc + " " + (mkAbbrArg elem abbr)
-          ) "abbr --add ${abbr.word} ${mkCursorArg abbr}" abbrArgs)
-          + optionalString (abbr.function == null) " \"${abbr.expansion}\"";
+          (foldl' (acc: elem: acc + " " + (mkAbbrArg elem abbr)) "abbr --add ${mkCursorArg abbr}" abbrArgs)
+          + " -- ${abbr.word}"
+          + optionalString (abbr.function == null) " ${lib.escapeShellArg abbr.expansion}";
 
         abbrs = concatStringsSep "\n" (map mkAbbrStr (attrValues cfg.abbreviations));
         aliases = concatStringsSep "\n" (
@@ -485,9 +484,11 @@ in
         pluginCompletions
         customPluginSources
         customPluginCompletions
+        "if status is-interactive"
         aliases
         abbrs
         completions
+        "end"
         cfg.configFile.content
       ]);
   };
--- a/wrapperModules/f/fish/module.nix
+++ b/wrapperModules/f/fish/module.nix
@@ -45,7 +45,7 @@ let
             "anywhere"
             "command"
           ];
-          default = "anywhere";
+          default = "command";
           description = ''
             The scope of the abbreviation.
 


BirdeeHub:
Thank you for the fixes and the continuing support.

I am not sure about changing the default. if you have an abbreviation in command position only, is that meaningfully different from an alias? If not, I would like to keep it as is, otherwise sure I guess we can change it but I would like to not do changes like that too often because we can't easily make a deprecation warning for that one.

Other than that, looks good to me, so either we change the default or not, but the rest is ready for merge.

iynaix:
I understand not wanting to change the default, but it's really unfortunate since that would mean having to add `position = command` to pretty much all the abbrs. Yes, it's pretty much an alias that expands when used, so the history shows the actual command instead of the alias.

The default is for --position is command:
https://fishshell.com/docs/current/cmds/abbr.html#add-subcommand

A quick search also shows that the main usage is to use the default --position:
https://github.com/search?q=path%3Aconfig.fish+abbr+--add&type=code
For --position anywhere
https://github.com/search?q=path%3Aconfig.fish+abbr+--add+--position+anywhere&type=code

BirdeeHub:
That definitely seems like there is precedent for using command as a default in more than just home manager.

> since that would mean having to add position = command to pretty much all the abbrs

They still work in command position when using anywhere, so theoretically you could leave them like that and still use them the same way? Although you might then have unintended expansions, so that is definitely a reason why one would have to be more specific.

Ultimately, I don't use fish, it doesn't matter to me, and I understand why one would want to match convention about this to some degree. I am OK with changing the default I just don't want to do it for no reason so I figured I would ask for more info about it and the reasoning for it.

---

Also, you can use type-merging to do something like this to change the default for one's own module (this particular snippet is untested, but I know type-merging works with the spec type, so minus syntax mistakes this would work, you declare the option again with the same type but a different submodule passed to it, which gets merged in with the existing submodule for that type which was already declared elsewhere)

```nix
{ pkgs, lib, wlib, config, ... }: {
  imports = [ wlib.wrapperModules.fish ];
  options.abbreviations = lib.mkOption {
    type = lib.types.attrsOf (wlib.types.spec {
      config.position = lib.mkDefault "command";
    });
  };
}
```

That being said, type-merging is also a little known feature of the nixpkgs module system which is not always easy to understand. So, picking the best default we can for the base module is always good as it isn't something we can rely on the user to figure out.

iynaix:
I guess it could be left as it is, since it's only triggered if the abbreviation is a standalone word followed by a space / enter. It just irks me since it doesn't follow the "principle of least astonishment" by deferring from the fish default.

I did not know about type merging, maybe I'll do that instead of patching it out instead.

PR #586, opened 2026-08-01
feat(wrapperModules.yazi): support init.lua

Treeniks:
I wanted to make use of yazi's [builtin session plugin](https://yazi-rs.github.io/docs/dds#session.lua) but found no way to set code for the `init.lua` file, so I added a new option for it, mimicking home-manager's `programs.yazi.initLua`.

### DIFF
--- a/wrapperModules/y/yazi/module.nix
+++ b/wrapperModules/y/yazi/module.nix
@@ -434,6 +434,20 @@ in
       };
     };
   };
+  options.initLua = lib.mkOption {
+    default = null;
+    description = ''
+      Content of the init.lua file.
+    '';
+    type = lib.types.nullOr (lib.types.either lib.types.path lib.types.lines);
+    example = lib.literalMD ''
+      ```lua
+      require("session"):setup {
+        sync_yanked = true,
+      }
+      ```
+    '';
+  };
   options.generatedConfig.output = lib.mkOption {
     type = lib.types.str;
     default = config.outputName;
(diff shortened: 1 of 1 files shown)


BirdeeHub:
This is good, but maybe we could do slightly better, in that we could use wlib.types.linkable and then check, and link instead of builtins.readFile, and that way we could accept an impure path?

Treeniks:
There aren't many examples on how to use `wlib.types.linkable` but I think I got it.

BirdeeHub:
Yeah that looks right, sorry I probably should have explained better.

wrapperModules/y/yazi/module.nix
```diff
@@ -520,6 +534,14 @@ in
         theme = config.settings.theme;
         vfs = config.settings.vfs;
         package = config.settings.package;
+      }
+    )
+    // {
+      initlua = lib.mkIf (config.initLua != null) {
+        relPath = "${config.binName}-config/init.lua";
```
BirdeeHub, line 541:
```suggestion
        relPath = lib.mkOverride 0 "${config.binName}-config/init.lua";
        output = lib.mkOverride 0 config.generatedConfig.output;
```

Also, I just realized, the above mapping expression does this, so we should do it in this bit too.

Setting output means that even if you wanted to put the config in a separate derivation output, the whole config will still stay together, and the lib.mkOverride 0 prevents it from being overridden individually.

We could debatably swap the lib.mkOverride 0 for lib.mkForce or just remove the lib.mkOverride 0, but we should do it for all of them if we do that, and we should definitely still set output. There is an option to change it for the whole generated config, so it seemed like a good idea to do the override to make sure it follows that option.

Other than that, all good to go!

PR #590, opened 2026-08-15
feat(wrapperModules.hypridle): init

jonas-elhs:
- Added me as maintainer
- Extracted `toHyprconfValue` from hyprlock module to lib
- Added hypridle module

Note: this module does not create a systemd service for example that automatically starts hypridle, should I add a comment for that in the module?

### DIFF
--- a/lib/lib.nix
+++ b/lib/lib.nix
@@ -721,6 +721,56 @@ in
   toKdl = import ./toKdl.nix { inherit lib wlib; };
 
   /**
+    Converts a Nix value to a Hypr conf string.
+
+    Parameters:
+    - `attrs`: Attrset of hypr conf keywords assigned either a value (string, boolean, number), a list of such values or an attrset that has the same structure:
+        value: a keyword statement is generated (https://wiki.hypr.land/0.54.0/Configuring/Keywords/)
+        list: for each value a separate keyword statement is generated
+        attrset: a section to configure variables is generated (https://wiki.hypr.land/0.54.0/Configuring/Variables/)
+    - `indentLevel`: (optional, default: 0) Start at a custom indent level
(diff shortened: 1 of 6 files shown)


FluxZA:
Thanks for submitting this. I think it would be good to specify that this wrapper doesn't handle the systemd integration. You could add it to the `pre` section of `meta.description` to display it as a section at the start of the module's docs.

As an aside, we are currently working on the systemd implementation for wrappers, so that is something that you could make use of soon if you did want to manage the service or give users the option to do so.

jonas-elhs:
I have now adapted the new systemd module and it works!
Is there a way however to access the wrapped executable from inside the wrapper module, because currently I have to use the unwrapped `hypridle` program in the `ExecStart` field of the systemd service?

BirdeeHub:
`config.wrapperPaths.placeholder`

https://birdeehub.github.io/nix-wrapper-modules/lib/core.html#wrapperpathsplaceholder

You could also do `${placeholder "out"}/bin/hypridle` but then if they changed the name it wouldn't follow it, there is a variable to target for a reason :smile:

(this is also why I am such a stickler for making sure placeholders work in all the options and not introducing unnecessary intermediate derivations, for things like this where you want to reference a path inside the final wrapper derivation)

It is also good to note that you cannot use `config.wrapper` to build a service file which goes inside `config.wrapper`, hence the use of the placeholders.

BirdeeHub:
Also, sorry about taking a while to get to this.

I took a 3 week unannounced mental health break from the project and a lot stacked up, and there are several hyprland related modules so I want to see what things I will need to think about deduplicating and stuff. So I will review them somewhat together, or at least in sequence

jonas-elhs:
Thanks for the answer :) and take all the time you need

PR #594, opened 2026-08-23
feat(lib.escapeShellArgsWithEnv): init

AlexLov:
This adds complementary function `escapeShellArgsWithEnv` (it's Args) to run `escapeShellArgWithEnv` function against a list of strings and concatenate the result with " " (space) as separator.

Works exact same way `escapeShellArgs` works with `escapeShellArg`.

### DIFF
--- a/lib/lib.nix
+++ b/lib/lib.nix
@@ -328,6 +328,29 @@ in
   escapeShellArgWithEnv = arg: ''"${lib.escape [ ''\'' ''"'' ] (toString arg)}"'';
 
   /**
+    Escape all shell arguments while preserving environment variable expansion.
+
+    Technically just applies escapeShellArgWithEnv to each element of list and
+    concatenate them with " " (space) as separator.
+    Useful to make shell command with arguments from a list.
+
+    # Type
+
+    ```
+    escapeShellArgs :: [String] -> String
+    ```
+
+    # Examples
+
+    ```nix
+    escapeShellArgsWithEnv [ "install" "$HOME/config.txt" "$XDG_CONFIG_HOME/app/config.txt" ]
+
+    => "install \"$HOME/config.txt\" \"$XDG_CONFIG_HOME/app/config.txt\""
+    ```
+  */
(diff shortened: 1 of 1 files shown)


BirdeeHub:
Im still not super sold on the base escapeShellArgWithEnv function to be honest. It is useful sometimes but there are circumstances where it breaks down I think. So I am less sure about how necessary this one is to add on top of it, when it is simply `builtins.concatStringsSep " " wlib.escapeShellArgWithEnv`.

Most of the places in the library that accept an escaping function accept one which escapes a single value, needing more would mean you are putting more than 1 thing in 1 option, and that often has some custom logic to do anyway.

Is this something you find yourself using constantly? I would be interested to hear where you are needing this often enough to have a builtin helper for this.

AlexLov:
It's not that I need it constantly. But once I need to use the same code at least twice, I decided that it's worth a function.
Of course, I can have this function in my own configuration, just thought it would be good to contribute back to the awesome project.
Currently I use it in configuration of my wrapped git like this (simplified version):
```nix
settings.difftool = let
  difftasticArgs = [
    "$LOCAL"
    "$REMOTE"
  ];
in {
  "difftastic".cmd = escapeShellArgsWithEnv ([
    (lib.getExe' config.wrappers.difftastic.wrapper "difft")
  ] ++ difftasticArgs);

  "difftastic-inline".cmd = escapeShellArgsWithEnv ([
    (lib.getExe' config.wrappers.difftastic.wrapper "difft")
    "--display=inline"
  ] ++ difftasticArgs);
};
```

BirdeeHub:
Ill see how I feel about it after I get to more of these PRs

The main issue, is that it just isn't a very good escaping function to begin with. I'm pretty sure there are several edge cases where it just doesn't work. Honestly for most of the options I just use `toString` or `lib.id` for the esc-fn field and escape stuff how I want usually when I need something unescaped. So I am not really sure I want to add a helper to make it easier to use, I kinda would prefer if I hadn't added it in the first place somewhat.

This function, and the `wlib.types.file` one, they did get rewritten but, they didnt get removed when I rewrote the repo this one was based on. I should probably have just not added them, but here we are.

PR #595, opened 2026-08-23
fix(wrapperModules.ghostty): fix evaluation warning about stdenv.isDarwin

AlexLov:
Fixes annoying evaluation warning:
```
evaluation warning: stdenv.isDarwin is deprecated, use stdenv.hostPlatform.isDarwin instead
```


BirdeeHub:
LGTM! Thanks!

PR #598, opened 2026-08-25
feat(wrapperModules.noctalia): init

jasdeep-dhillon:
Add myself as a maintainer and added a wrapperModule for pkgs.noctalia. Most of it just edited from the noctalia-shell module. Ran `nix fmt` and `nix flake check -Lv ./ci` to check the docs and modules.

I have not overriden $NOCTALIA_STATE_HOME variable as that causes the GUI interface to not work. 
Please let me now if there are any changes or additions that are needed.

### DIFF
--- a/maintainers/default.nix
+++ b/maintainers/default.nix
@@ -121,4 +121,9 @@
     github = "aliaslion";
     githubId = 122117018;
   };
+  jasdeep = {
+    name = "Jasdeep Dhillon";
+    github = "jasdeep-dhillon";
+    githubId = 59929844;
+  };
 }
--- a/wrapperModules/n/noctalia/module.nix
+++ b/wrapperModules/n/noctalia/module.nix
@@ -1,0 +1,73 @@
+{
+  config,
+  wlib,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  tomlFmt = pkgs.formats.toml { };
+in
+{
+  imports = [ wlib.modules.default ];
+  options = {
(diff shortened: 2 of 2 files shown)


KendleMintJed:
One thing I'm noticing is that your description of the setting option is
```
Noctalia shell configuration settings as an attribute set,
to be written to $NOCTALIA_CONFIG_HOME/noctalia/settings.json`.
```
even though `typeName = "TOML"`

and the type of your colors option is `typeName = "TOML"`
even though `content = builtins.toJSON config.colors;`

jasdeep-dhillon:
Sorry about that I forgot to change the description for settings and colors. Added a meta description as well though  it could probably be better.

iynaix:
Perhaps you could add something like 

```nix
config.drv.installPhase = ''
    runHook preInstall
    ${lib.getExe config.package} validate "${config.constructFiles.settings.path}"
    runHook postInstall
'';
```

to validate the noctalia config, similar to the validation done in the niri config?

jasdeep-dhillon:
> Perhaps you could add something like
> 
> ```nix
> config.drv.installPhase = ''
>     runHook preInstall
>     ${lib.getExe config.package} validate "${config.constructFiles.settings.path}"
>     runHook postInstall
> '';
> ```
> 
> to validate the noctalia config, similar to the validation done in the niri config?

I did try to do it before but the error reporting was pretty bad as it got lost in the derivation build logs when I tried it. I will try to look into it again and see if  error reporting could be a bit better.

